PR #18, opened 2019-07-13
transaction: re-export TransactionBuilder

mikrohash:
TransactionBuilder is not accessible by other crates. This means that other crates cannot create new transactions, only parse bytes from existing transactions into transaction objects. TransactionBuilder is required by external crates; for example by tangle app extensions - such as CHAT.ixi - which create new transactions.


wusyong:
LGTM, however commit message should be "re-export" I think.

mikrohash:
change is already pushed to dev branch: https://github.com/iotaledger/bee/commit/2c64e629d7975efd2b441c9920087e0f16c891e1

PR #32, opened 2019-08-27
Added rustfmt.toml

Alex6323:


wusyong:
According to [rustfmt configurations](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#edition), edition will be picked up by `Cargo.toml` if provided. It seems this file is not needed in this case.

SuperFluffy:
Yes, that's indeed the case. But I would like to make it explicit and not depend on an implementation detail. With all other `rustfmt` options we are depending on the default values, which is not the case with `edition`.

SuperFluffy:
Merged this without successful CI because there is nothing to test.

PR #51, opened 2019-09-10
trinary: Implement type Trytes

### LINKED ISSUE #50: trinary: Trytes Module

wusyong:
## Description

Resolve #50 and [RFC#4](https://github.com/iotaledger/bee-rfcs/pull/4)

TODO:
- [x] Suggestions from RFC
- [ ] Documentation
- [ ] Add lints
- [ ] Cleanup layout
- [ ] Test coverage

## Testing

Pass CI

### DIFF
--- a/iota-trinary/src/trytes.rs
+++ b/iota-trinary/src/trytes.rs
@@ -1,0 +1,855 @@
+use alloc::{string::String, vec::Vec};
+use core::ops::{self, Add, AddAssign, Deref, DerefMut, Index, IndexMut};
+use core::str::{self, FromStr};
+use core::{fmt, ptr};
+
+pub use core::convert::TryFrom;
+
+use crate::Result;
+
+pub type Tryte = u8;
+
+/// A char array holding all acceptable characters in the tryte
+/// alphabet. Used because strings can't be cheaply indexed in rust.
+pub const TRYTE_ALPHABET: [u8; 27] = [
+    57, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 80, 81, 82, 83, 84, 85, 86, 87,
+    88, 89, 90,
+];
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Trytes {
+    vec: Vec<Tryte>,
+}
+
+impl Trytes {
+    /// Creates a new empty `Trytes`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use iota_trinary::*;
+    ///
+    /// let t = Trytes::new();
+    /// ```
+    #[inline]
+    pub fn new() -> Trytes {
+        Trytes { vec: Vec::new() }
+    }
+
+    /// Creates a new empty `Trytes` with a particular capacity.
+    ///
+    /// If the given capacity is `0`, no allocation will occur, and this method
+    /// is identical to the [`new`] method.
+    ///
+    /// [`new`]: #method.new
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use iota_trinary::*;
+    ///
+    /// let mut t = Trytes::with_capacity(10);
+    ///
+    /// // The Trytes contains no chars, even though it has capacity for more
+    /// assert_eq!(t.len(), 0);
+    ///
+    /// // These are all done without reallocating...
+    /// let cap = t.capacity();
+    /// for _ in 0..10 {
+    ///     t.push('a');
+    /// }
+    ///
+    /// assert_eq!(t.capacity(), cap);
+    ///
+    /// // ...but this may make the vector reallocate
+    /// t.push('a');
+    /// ```
+    #[inline]
+    pub fn with_capacity(capacity: usize) -> Trytes {
+        Trytes {
+            vec: Vec::with_capacity(capacity),
+        }
+    }
+
+    #[inline]
+    pub fn from_utf8(vec: Vec<u8>) -> Result<Trytes> {
+        Self::all_tryte_alphabete(vec.iter().copied())?;
+        Ok(Trytes { vec })
+    }
+
+    pub unsafe fn from_utf8_unchecked(bytes: Vec<u8>) -> Trytes {
+        Trytes { vec: bytes }
+    }
+
+    #[inline]
+    pub unsafe fn from_raw_parts(buf: *mut u8, length: usize, capacity: usize) -> Trytes {
+        Trytes {
+            vec: Vec::from_raw_parts(buf, length, capacity),
+        }
+    }
+
+    /// Converts a `Trytes` into a byte vector.
+    ///
+    /// This consumes the `Trytes`, so we do not need to copy its contents.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use iota_trinary::*;
+    ///
+    /// let t = Trytes::try_from("ABCD").unwrap();
+    /// let bytes = t.into_bytes();
+    ///
+    /// assert_eq!(&[65, 66, 67, 68][..], &bytes[..]);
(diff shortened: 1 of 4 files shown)


Alex6323:
Some feedback from my side on the implementation so far:

1. Add the following crate-level attributes to 'lib.rs' and fix all warnings
		#![warn(
			missing_debug_implementations,
			missing_docs,
			rust_2018_idioms,
			unreachable_pub
		)]

2. Do we really need to expose `TRYTE_ALPHABET`? Consider making this only `pub(crate)`.
3. Suggestion: Change error message "Invalid trytes alphabet" to something a bit more descriptive like "String contains characters, that are not part of the tryte alphabet A-Z,9."
4. In your implementation `Trytes` is a "newtype" for `BytesMut`, but oftentimes I only need immutable tryte sequences, so shouldn't this type be named `TrytesMut` accordingly? And then we should have an immutable, but dynamically sized `Trytes` type as well wrapping `Bytes`. Additionally for most commonly used fixed sized types I would like to have `Trytes81` wrapping a `smallvec` or a (fixed size) array. What do you think?
5. Suggestion: Use `Self` wherever you can. Example: Instead of writing
	```Rust
	pub fn new() -> Trytes {
        Trytes(BytesMut::new())
    }
	```
	you could do this:
	```Rust
	pub fn new() -> Self {
        Self(BytesMut::new())
    }
	```
	It's just a personal preference, but makes renaming stuff somewhat easier.
6. Make byte encoding of trytes an implementation detail by allowing for 
	* 3trits/1byte (or 1tryte/1byte)
	* 9trits/2bytes (or 3trytes/2bytes)
	* 10trits/2bytes
	It probably makes sense to introduce a `TryteEncoding` enum to represent those, and then follow zesterer's suggestions.

wusyong:
@Alex6323 Thanks for the feedbacks! I will reply each suggestion below. But I think you can also review under "filed changed" section next time.

1. Right that's in TODO list

2. 3. 5. Will do!

4. That was my initial thought too. But since this is  more like a sub set for `String` type I would like to keep thing simple. Beside, there are actually many time we want a mutable object like replace nonce, add checksum etc.

6. Do we really need various of "Trytes" encoded byte? I think we only need encoding technique for "Trits".

iota-trinary/src/trytes.rs
```diff
@@ -0,0 +1,675 @@
...
+    /// use iota_trinary::*;
+    ///
+    /// let t = Trytes::try_from("VBCDFDTCAD").unwrap();
+    ///
+    /// assert_eq!("Lorem", t.decode());
+    /// ```
+    pub fn decode(&self) -> String {
+        let mut s = String::with_capacity(self.len() / 2);
+        for slice in self.chunks(2) {
+            let first: u8 = TRYTE_ALPHABET.iter().position(|&x| x == slice[0]).unwrap() as u8;
```
zesterer: There's no requirement to specify the type here with `: u8` - Rust's type inference is more than capable of figuring out what it should be.

iota-trinary/src/trytes.rs
```diff
@@ -0,0 +1,675 @@
...
+            let second: u8 = TRYTE_ALPHABET.iter().position(|&x| x == slice[1]).unwrap() as u8;
```
zesterer: Instead of walking the entire array here, it would be more efficient to create a function that goes the other way. Something like this:

```rust
fn chat_to_tryte(c: char) -> Option<Tryte> {
    match c {
        'A'..'B' => Some(c as u8 - b'A'),
        '9' => Some(26),
        _ => None,
    }
}
```

iota-trinary/src/trytes.rs
```diff
@@ -0,0 +1,675 @@
...
+
+impl Deref for Trytes {
+    type Target = [Tryte];
+
+    fn deref(&self) -> &Self::Target {
+        &self.0[..]
+    }
+}
+
+impl DerefMut for Trytes {
```
zesterer: I don't recommend implementing traits like this for `Trytes` at all. `Tryte` is just a type alias of `u8`, which means that it's trivial to modify the structure to contain bytes that do not correspond to valid trytes. If this data structure is to be sound, inputs and outputs should be validated (with the option to disable that validation for release builds).

iota-trinary/src/trytes.rs
```diff
@@ -0,0 +1,675 @@
...
+    /// ```
+    /// use iota_trinary::*;
+    ///
+    /// let t = Trytes::try_from("VBCDFDTCAD").unwrap();
+    ///
+    /// assert_eq!("Lorem", t.decode());
+    /// ```
+    pub fn decode(&self) -> String {
+        let mut s = String::with_capacity(self.len() / 2);
+        for slice in self.chunks(2) {
```
zesterer: Are we 100% sure that we'll *always* have an even number of trytes in this data structure? If we don't, `self.chunks(2)` will produce an iterator where the last slice only has a single element. This would make the `second` line below panic.

iota-trinary/src/trytes.rs
```diff
@@ -0,0 +1,675 @@
...
+        ptr::copy(
+            self.0.as_ptr().add(idx),
+            self.0.as_mut_ptr().add(idx + amt),
+            len - idx,
+        );
+        ptr::copy(bytes.as_ptr(), self.0.as_mut_ptr().add(idx), amt);
+        self.0.set_len(len + amt);
+    }
+
+    fn all_tryte_alphabete<I>(vals: I) -> Result<()>
```
zesterer: More idiomatic syntax for this would look like `fn all_tryte_alphabete(vals: impl Iterator<Item=u8>) -> Result<()>`.

iota-trinary/src/trytes.rs
```diff
@@ -0,0 +1,675 @@
...
+    /// ```
+    /// use iota_trinary::*;
+    ///
+    /// let mut t = Trytes::new();
+    ///
+    /// t.encode("Lorem");
+    ///
+    /// assert_eq!("VBCDFDTCAD", t.as_str());
+    /// ```
+    pub fn encode(&mut self, plain: &str) {
```
zesterer: 'Encode' implies a 1:1 mapping between an input and an output, but this appends to the buffer. Consider renaming this to `append_encoding` perhaps?

iota-trinary/src/trytes.rs
```diff
@@ -0,0 +1,675 @@
...
+        ptr::copy(bytes.as_ptr(), self.0.as_mut_ptr().add(idx), amt);
+        self.0.set_len(len + amt);
+    }
+
+    fn all_tryte_alphabete<I>(vals: I) -> Result<()>
+    where
+        I: Iterator<Item = u8>,
+    {
+        let mut v = vals;
+        ensure!(v.all(Self::is_tryte_alphabete), "Invalid trytes alphabete.");
```
zesterer: I strongly recommend not using either `ensure!` or the `failure` crate in general. Whenever a failure occurs, the crate performs an allocation. This is likely to result in significantly slower code for a lot of use-cases. Instead, just roll your own error enum.

iota-trinary/src/trytes.rs
```diff
@@ -0,0 +1,649 @@
...
+    ///
+    /// let mut t = Trytes::with_capacity(10);
+    ///
+    /// // The Trytes contains no chars, even though it has capacity for more
+    /// assert_eq!(t.len(), 0);
+    ///
+    /// // These are all done without reallocating...
+    /// let cap = t.capacity();
+    /// for _ in 0..10 {
+    ///     t.push('a');
```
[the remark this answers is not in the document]
wusyong: yeah, I should give `A` as example instead. But since Tryte RFC is on hold. This draft would probably not continue. I will close this draft, once 3T1B is going to implement.

PR #66, opened 2020-03-26
Update contribution guidelines and README

JakeSCahill:
## Description

Updated the contribution documentation to use the latest IF templates.


JakeSCahill:
Work in progress

Still adding some missing content to the CONTTRIBUTING.md

.github/SECURITY.md
```diff
@@ -0,0 +1,23 @@
+<h2>Responsible disclosure policy</h2>
```
thibault-martinez, line 1: Any specific reason why this is HTML and not markdown ?

JakeSCahill: No specific reason.

The security guys wrote that and I just kept it the same.

PR #79, opened 2020-06-01
Add build and test workflows

thibault-martinez:
# Description of change

Adds `build` and `test` Github workflows.

## Type of change

Tested in `bee-p`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,50 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        rust: [nightly, beta, stable]
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Install ${{ matrix.rust }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,50 @@
...
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
wusyong, line 16: Are we also considering `windows-latest` and `macos-latest`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,50 @@
...
+        rust: [nightly, beta, stable]
```
wusyong, line 17: I think we just need stable since we already choose to develop on it.

.github/workflows/build.yml
thibault-martinez: Yes we are, in the near future. But as a start to be able to start releasing crate let's start with  ubuntu only.

.github/workflows/build.yml
thibault-martinez: Yeah, doesn't really hurt tbh. It's cached so they are all very fast and it gives us an idea if everything is compatible with the future stable versions.

PR #112, opened 2020-07-28
Signing scheme

thibault-martinez:
## Change checklist


bee-signing/src/ternary/mss.rs
```diff
@@ -0,0 +1,407 @@
...
+
+        // Create the merkle tree by hashing from botton to top.
+        for depth in (0..self.depth - 1).rev() {
+            for i in 0..(1 << depth) {
+                let index = (1 << depth) + i - 1;
+                let left_index = index * 2 + 1;
+                let right_index = left_index + 1;
+
+                sponge
+                    .absorb(&tree[left_index * HASH_LENGTH..(left_index + 1) * HASH_LENGTH])
```
zesterer, line 177: And `and_then` chain might make the error handling a little clearer here

bee-signing/src/ternary/mss.rs
```diff
@@ -0,0 +1,407 @@
...
+                break;
+            }
+
+            if signature_index & j != 0 {
+                sponge.absorb(sibling).map_err(|_| Self::Error::FailedSpongeOperation)?;
+                sponge.absorb(&hash).map_err(|_| Self::Error::FailedSpongeOperation)?;
+            } else {
+                sponge.absorb(&hash).map_err(|_| Self::Error::FailedSpongeOperation)?;
+                sponge
+                    .absorb(&sibling)
```
zesterer, line 342: Why is `sibling` taken by reference here and not above? It also seems to be the cause of the bizarre formatting.

bee-signing/src/ternary/mss.rs
```diff
@@ -0,0 +1,407 @@
...
+/// Merkle Signature Scheme signature.
+pub struct MssSignature<S> {
+    state: TritBuf<T1B1Buf>,
+    index: Option<u64>,
+    sponge: PhantomData<S>,
+}
+
+impl<S: Sponge + Default> MssSignature<S> {
+    /// Set the index of the signature.
+    pub fn index(mut self, index: u64) -> Self {
```
zesterer, line 383: I think this method name might be a little misleading. Single-noun methods are usually getters in Rust

bee-signing/src/ternary/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+mod constants;
+mod scheme;
+
+pub mod mss;
+pub mod seed;
+pub mod wots;
+
+pub use constants::SIGNATURE_FRAGMENT_LENGTH;
+pub use scheme::{PrivateKey, PrivateKeyGenerator, PublicKey, RecoverableSignature, Signature};
```
wusyong, line 22: Could we also export `Seed` here?
```suggestion
pub use scheme::{PrivateKey, PrivateKeyGenerator, PublicKey, RecoverableSignature, Signature};
pub use seed::Seed;
```

bee-signing/src/ternary/wots/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+        if message.len() != HASH_LENGTH {
+            return Err(Error::InvalidMessageLength(message.len()));
+        }
+
+        let mut sponge = S::default();
+        let mut signature = self.state.clone();
+
+        for (i, chunk) in signature.chunks_mut(HASH_LENGTH).enumerate() {
+            // Safe to unwrap because 3 trits can't underflow/overflow an i8.
+            let val = i8::try_from(&message[i * 3..i * 3 + 3]).unwrap();
```
zesterer, line 147: You can cast `Btrit` to `i8` using `x as i8`

thibault-martinez: But this is not a `Btrit`, this is a `&Trits[_]`

bee-signing/src/ternary/mod.rs
thibault-martinez: Whoops! Yes of course.

thibault-martinez: Ah but wait, it is exported though `seed` already.

bee-signing/src/ternary/mss.rs
thibault-martinez: Good catch, `sibling` is already a reference. Removing it.

bee-signing/src/ternary/mss.rs
thibault-martinez: So, `set_index` ?

thibault-martinez: Or `with_index` ?

bee-signing/src/ternary/wots/normalize.rs
```diff
@@ -0,0 +1,79 @@
...
+                        break;
+                    }
+                }
+                sum -= 1;
+            }
+        } else {
+            while sum < 0 {
+                for t in &mut normalized[i * MESSAGE_FRAGMENT_LENGTH..(i + 1) * MESSAGE_FRAGMENT_LENGTH] {
+                    if (*t as i8) < Tryte::MAX_VALUE as i8 {
+                        *t += 1;
```
zesterer, line 63: These loops can be combined with the `if sum > 0 {` check being hoisted into the middle of it.

thibault-martinez: Done

bee-signing/src/ternary/mss.rs
```diff
@@ -0,0 +1,407 @@
...
+        self
+    }
+
+    /// Builds the private key generator.
+    pub fn build(self) -> Result<MssPrivateKeyGenerator<S, G>, Error> {
+        let depth = match self.depth {
+            Some(depth) => match depth {
+                0..=MAX_MSS_DEPTH => depth,
+                _ => return Err(Error::InvalidDepth(depth)),
+            },
```
Alex6323, line 120: I'ld check that in the setter already not even allowing an invalid state of the builder. The reason is *cautiousness*, because we cannot not know if some future dev decides to introduce a `build_xyz` method for some reason.

bee-signing/src/ternary/mss.rs
thibault-martinez: Done

bee-signing/src/ternary/scheme.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// Errors occuring while handling public keys.
+    type Error;
+
+    /// Returns the size of the signature.
+    fn size(&self) -> usize;
+
+    /// Creates a signature from trits.
+    fn from_trits(buf: TritBuf<T1B1Buf>) -> Result<Self, Self::Error>
+    where
+        Self: std::marker::Sized;
```
zesterer, line 219: `Sized` is in the prelude so no `std::marker::` is required.

thibault-martinez: Done

bee-signing/src/ternary/seed.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub fn rand() -> Self {
+        // `ThreadRng` implements `CryptoRng` so it is safe to use in cryptographic contexts.
+        // https://rust-random.github.io/rand/rand/trait.CryptoRng.html
+        let mut rng = rand::thread_rng();
+        let trits = [Btrit::NegOne, Btrit::Zero, Btrit::PlusOne];
+        let mut seed = [Btrit::Zero; HASH_LENGTH];
+
+        for trit in seed.iter_mut() {
+            *trit = trits[rng.gen_range(0, trits.len())];
+        }
```
Alex6323, line 63: I propose to use rand's `sample` API instead. It much cleaner IMHO, don't know if it also has performance benefits. I just find `gen_range` in a loop not so elegant here.

bee-signing/src/ternary/wots/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+#[derive(Debug, Error, PartialEq)]
+pub enum Error {
+    /// Missing security level in generator.
+    #[error("Missing security level in generator.")]
+    MissingSecurityLevel,
+    /// Failed sponge operation.
+    #[error("Failed sponge operation.")]
+    FailedSpongeOperation,
+    /// Invalid entropy length.
+    #[error("Invalid entropy length, should be 243 trits.")]
```
Alex6323, line 48: a nice to have would be to print the actual provided entropy length. 
```suggestion
    #[error("Invalid entropy length, should be 243 trits, but was {0}.")]
```
Same with the other errors down below.

thibault-martinez: Done.

bee-signing/src/ternary/seed.rs
thibault-martinez: Can you give me an example with `sample` ? Not sure to see how to use it here.

bee-signing/src/ternary/mss.rs
thibault-martinez: Hmm I don't know. I feel like the whole point of having a builder is to possibly be in invalid state. Having only one method to check at the end seems easier. Also, I do not understand the second sentence.

Alex6323: Never mind. Leave it as it is. My suggestion would make things more complicated, and another hypothetical `build_xyz` method would be a wrong implementation of the builder pattern.

rufsam: I guess an opinion of @zesterer and @christianpoveda on this would be helpful too. We should be consistent with all our builders.

bee-signing/src/ternary/seed.rs
Alex6323: okay, os you basically want to uniformly random select between the 3 possible trit values.
https://docs.rs/rand/0.7.3/rand/distributions/uniform/struct.Uniform.html#method.sample

Alex6323: so from reading the docs it seems to actually be faster and also less biased.

thibault-martinez: Alright, done.

PR #137, opened 2020-11-02
Add Packable trait

thibault-martinez:
## Change checklist


bee-common/bee-common/src/packable.rs
```diff
@@ -0,0 +1,75 @@
...
+    where
+        Self: Sized;
+}
+
+macro_rules! impl_packable_for_num {
+    ($ty:ident) => {
+        impl Packable for $ty {
+            type Error = std::io::Error;
+
+            fn packed_len(&self) -> usize {
```
zesterer, line 46: This doesn't look correct. `size_of` doesn't include the size of associated buffers or things pointed to by this type.

bee-common/bee-common/src/packable.rs
```diff
@@ -0,0 +1,75 @@
...
+                std::mem::size_of::<$ty>()
```
zesterer: Perhaps use `std::mem::size_of_val(&self.to_le_bytes())` to minimise potential problems with discrepancies with le byte conversion? "the size of this type in memory is equivalent to the number of bytes it produces when encoded as a list of little-endian bytes" is, afaik, not something that can be relied upon to be true across all platforms.

bee-common/bee-common/src/packable.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub use std::io::{Read, Write};
+
+/// A trait to pack and unpack types to and from bytes.
+pub trait Packable {
+    /// Associated error type.
+    type Error;
+
+    /// Returns the length of the packed bytes.
+    fn packed_len(&self) -> usize;
```
zesterer, line 22: Perhaps this could be changed to be an approximation for the purpose of capacity allocation? (it seems like estimating capacity is the primary use for it) That way, it might be made considerably less costly to compute in some cases (i.e: hashmaps not needing to iterate all elements) without being 'incorrect'. Something akin to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint

thibault-martinez: I am not sure we can do this. It's sometimes used for verification purpose to compare the length of something read with an announced length.

thibault-martinez: Meaning there are situations we actually rely on this being the exact size.

zesterer: Okay, fair enough

zesterer (approved):

louaykamel (approved):

PR #143, opened 2020-11-18
bee-storage crate

louaykamel:
- [x] Finalize Documentation


bee-storage/bee-storage/src/access/batch.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[async_trait::async_trait]
+/// BatchBuilder trait will extend the Backend with Batch builder functionality,
+/// therefore it should be explicitly implemented for the corresponding Backend.
+pub trait BatchBuilder: Backend + Sized {
+    /// Batch type acts like memory buffer which queue all the write operations.
+    type Batch: Default + Send + Sized;
+
+    /// This method will create and return the constraint Batch object
+    fn batch_begin() -> Self::Batch {
```
zesterer, line 14: Any reason why this isn't `begin_batch`? Seems like a weird ordering. Ditto `batch_commit`.

louaykamel: @thibault-martinez please resolve this.

bee-storage/bee-storage/src/storage.rs
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use async_trait::async_trait;
+use serde::de::DeserializeOwned;
+
+use std::error::Error;
+
+#[async_trait]
+/// Trait to be implemented on storage backend, which determine how to start and shutdown the storage.
+pub trait Backend: Sized + Send + Sync + 'static {
```
Alex6323: I'ld prefer the term `StorageBackend` or just `Storage`. The reason is, that the term `Backend` is not specific enough as there are a lot of backends for all kinds of things out there. One could argue, that one only needs to import the module `storage`, and in the code one would always use `storage::Backend` which then is crystal clear when reading it. But if we tend to import the trait itself (which I think is more likely), then IMHO `StorageBackend` or `Storage` would be a tad nicer to read.

bee-storage/bee-storage/src/access/batch.rs
samuel-rufi: what about having a `begin_batch`, `commit_batch`, `insert_to_batch` and a `delete_from_batch`?

bee-storage/bee-storage/src/access/fetch.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::backend::StorageBackend;
+
+/// `Fetch<K, V>` trait extends the `StorageBackend` with `fetch` operation for the (key: K, value: V pair);
+/// therefore, it should be explicitly implemented for the corresponding `StorageBackend`.
+#[async_trait::async_trait]
+pub trait Fetch<K, V>: StorageBackend {
```
samuel-rufi, line 9: If we ever need two fetch operations which have the same input type and the same output type, but do have a different inner implemenation, is this possible to do?

thibault-martinez: That's not possible per se. The solution is to have wrapper types to better express what you are trying to do.

samuel-rufi: okey true, wrapper types would handle it
so it won't be a problem

samuel-rufi (dismissed):
Besides that open comment, everything is fine in my opinion. Great job.

bee-storage/bee-storage/src/access/batch.rs
Alex6323: Depending on whether `batch` would be used as the *object* or as part of the *verb*, I'ld go with Josh calling the function `begin_batch` (batch is the object), or `batch_begin` (batch is part of the verb). In the latter case something else is begun, that might be omitted because it's obvious from the context. That is something @LouayKamel has to decide.

Alex6323: But from the docs it seems like the object is the `batch` itself, and hence `begin_batch`, `create_batch` or `new_batch` would be cleaner.

bee-storage/bee-storage/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "bee-storage"
+version = "0.1.0-alpha"
+authors = ["IOTA Stiftung"]
+edition = "2018"
+description = ""
```
Alex6323: missing description

louaykamel (commented):
LGTM

bee-storage/bee-storage/README.md
```diff
@@ -0,0 +1 @@
+# bee-storage
```
Alex6323, line 1: I suggest adding a small sentence like "Storage layer for the Bee framework.", or something similar.

bee-storage/bee-storage/src/access/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+/// Holds the contract for exist access operation.
+pub mod exist;
+/// Holds the contract for fetch access operation.
+pub mod fetch;
+/// Holds the contract for insert access operation.
+pub mod insert;
+/// Holds the contract for stream access operations.
+pub mod stream;
+/// Holds the contract for truncate access operations.
+pub mod truncate;
```
Alex6323: I think it would make the API simpler if we'ld make those modules private, and only use the exports below. It is unlikely that those modules will ever contain more than the corresponding trait.

bee-storage/bee-storage/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+//! ## Access implementation:
+//! - Traits contracts which define the general purpose database operations such as (insert, fetch, ...);
+//! ## Backend implementation:
+//! - Trait contract to start and shutdown backends;
+//! - Configuration and associated builder to configure different backends;
+//!
+//! This crate tries to simplify the implementation of various storage backends and provides unified access API for the
+//! application/user space.
+
+#![deny(missing_docs)]
```
Alex6323, line 17: I'ld also include `#![deny(warnings)]`, bc I think for a production library this should be the default. We can still explicitly allow things that would throw a warning, but it would force us to think about why a warning is okay in a particular situation.

Alex6323 (changes requested):

bee-storage/bee-storage/src/lib.rs
thibault-martinez: Sure why not. The CI is actually already doing it repository wise so this is not strictly necessary but it can't hurt to add it.

bee-storage/bee-storage/src/access/mod.rs
louaykamel: I agree, much simpler and cleaner.

samuel-rufi (dismissed):

bee-storage/bee-storage/src/access/stream.rs
```diff
@@ -0,0 +1,17 @@
...
+use crate::backend::StorageBackend;
+
+use futures::stream::Stream;
+
+/// `AsStream<'a, K, V>` trait extends the `StorageBackend` with `stream` operation for the (key: K, value: V) pair;
+/// therefore, it should be explicitly implemented for the corresponding `StorageBackend`.
+#[async_trait::async_trait]
+pub trait AsStream<'a, K, V>: StorageBackend {
+    /// Type to iterate through the <K, V> collection.
+    type Stream: Stream;
```
zesterer: Should `Stream` perhaps be bound by `Stream<Item = (K, V)>`?

bee-storage/bee-storage/src/access/stream.rs
```diff
@@ -0,0 +1,17 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::backend::StorageBackend;
+
+use futures::stream::Stream;
+
+/// `AsStream<'a, K, V>` trait extends the `StorageBackend` with `stream` operation for the (key: K, value: V) pair;
+/// therefore, it should be explicitly implemented for the corresponding `StorageBackend`.
+#[async_trait::async_trait]
+pub trait AsStream<'a, K, V>: StorageBackend {
```
zesterer, line 11: If this was changed to `IntoStream` (in the style of `IntoIterator`) then it could consume `self` which makes it more generally applicable (it can still be implemented for all current cases by being implemented by a reference to them)

bee-storage/bee-storage/src/access/truncate.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::backend::StorageBackend;
+
+/// `Truncate<K, V>` trait extends the `StorageBackend` with `truncate` operation for the (key: K, value: V) pair;
+/// therefore, it should be explicitly implemented for the corresponding `StorageBackend`.
+#[async_trait::async_trait]
+pub trait Truncate<K, V>: StorageBackend {
+    /// Truncates all the entries associated with the (K, V) pair from the storage.
```
zesterer, line 10: What does truncation mean in this context?

thibault-martinez: Removing all entries in a column family

bee-storage/bee-storage/src/access/stream.rs
thibault-martinez: Good idea :)

zesterer (approved):

samuel-rufi (approved):

Alex6323 (approved):

thibault-martinez (approved):

PR #144, opened 2020-11-26
Stabilised event bus

zesterer:
# Description of change

Stabilising `event` (with minor changes) from `bee-p`

## Type of change

The event bus has prolonged use in `bee-p`.

Note that use of `dashmap` 3.11 can imply a deadlock on mutable insertion. This needs looking into (perhaps with a possible replacement with a `Mutex<HashMap<_, _>>` temporarily) but any fix will not be a breaking change.

## Change checklist


bee-common/bee-common/Cargo.toml
```diff
@@ -17,6 +17,7 @@ futures = "0.3"
 log = { version = "0.4", features = ["serde"] }
 serde = { version = "1.0", features = ["derive" ] }
 thiserror = "1.0"
+dashmap = "3.11"
```
thibault-martinez: Alphanumeric order please

bee-common/bee-common/src/event.rs
```diff
@@ -0,0 +1,64 @@
...
+
+//! A module that provides a generic, type-safe event bus for arbitrary event types.
+
+use dashmap::DashMap;
+use std::any::{Any, TypeId};
+
+type Listener<'a> = dyn Fn(&dyn Any) + Send + Sync + 'a;
+
+/// An event bus for arbitrary event types.
+pub struct Bus<'a, ID = TypeId> {
```
thibault-martinez, line 13: Even if it's in the `event` module, I wonder if `Bus` is a sufficient name or not. When imported, I am not sure it's expressive enough. What do you think of `EventBus` ?

bee-common/bee-common/src/event.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A module that provides a generic, type-safe event bus for arbitrary event types.
+
+use dashmap::DashMap;
```
thibault-martinez, line 6:
```suggestion
use dashmap::DashMap;

```

bee-common/bee-common/src/event.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::any::{Any, TypeId};
+
+type Listener<'a> = dyn Fn(&dyn Any) + Send + Sync + 'a;
+
+/// An event bus for arbitrary event types.
+pub struct Bus<'a, ID = TypeId> {
+    listeners: DashMap<TypeId, Vec<(Box<Listener<'a>>, ID)>>,
```
thibault-martinez, line 14: Why does ID default to TypeID ?

bee-common/bee-common/src/event.rs
```diff
@@ -0,0 +1,64 @@
...
+impl<'a, ID> Default for Bus<'a, ID> {
+    fn default() -> Self {
+        Self {
+            listeners: DashMap::default(),
+        }
+    }
+}
+
+impl<'a, ID: Clone + PartialEq> Bus<'a, ID> {
+    /// Dispatch an event via is this event bus.
```
thibault-martinez: Weird sentence ?

bee-common/bee-common/src/event.rs
```diff
@@ -0,0 +1,64 @@
...
+    ///
+    /// All active listeners registered for this event will be invoked.
+    pub fn dispatch<E: Any>(&self, event: E) {
+        if let Some(mut ls) = self.listeners.get_mut(&TypeId::of::<E>()) {
+            ls.iter_mut().for_each(|(l, _)| l(&event))
+        }
+    }
+
+    /// Add an event listener bound to a specific event type, `E`, and registered
```
thibault-martinez: I think the 120 columns policy should apply to comments as well.

bee-common/bee-common/src/event.rs
```diff
@@ -0,0 +1,64 @@
...
+        self.listeners
+            .iter_mut()
+            .for_each(|mut listeners| listeners.retain(|(_, listener_id)| listener_id != &id));
+    }
+}
+
+impl<'a> Bus<'a, TypeId> {
+    /// Add an event listener bound to a specific event type, `E`, and registered using the
+    /// `TypeId` of a worker, `W`.
+    pub fn add_worker_listener<W: Any, E: Any, F: Fn(&E) + Send + Sync + 'a>(&self, handler: F) {
```
thibault-martinez: I know it's generic and it's just wording, but I'm not sure what to think about the usage of `worker` here. I mean, it probably could be used in any context out of bee and our worker design.

bee-common/bee-common/src/event.rs
```diff
@@ -0,0 +1,64 @@
...
+        self.add_listener_raw(TypeId::of::<W>(), handler);
+    }
+
+    /// Add an event listener bound to a specific event type, `E`, registered using a hidden type
+    /// that will prevent its removal until the event bus is dropped.
+    pub fn add_listener<E: Any, F: Fn(&E) + Send + Sync + 'a>(&self, handler: F) {
+        struct Static;
```
thibault-martinez, line 62: What difference with using `()` ?

thibault-martinez (changes requested):
This is missing tests, I think they were some in the `tests` folder, don't forget to import them and check if they are still up to date.

bee-common/bee-common/src/event.rs
zesterer: This type is hidden and so can't be referred to externally (i.e: you can't purge by its `TypeId` earlier)

bee-common/bee-common/src/event.rs
zesterer: I thought the same but I can't think of an alternative. I originally went for `add_id_listener` but that sounds strange.

bee-common/bee-common/src/event.rs
thibault-martinez: And this is a different type for each `E`, right ?

bee-common/bee-common/src/event.rs
zesterer: I imagine that this will be the most common use-case.

thibault-martinez: Yeah, I think I just have a hard time imagining it not being TypeID, did you have an example in mind ?

bee-common/bee-common/src/event.rs
zesterer: I personally think it's expressive enough ("bus" is commonly used as some central communication system) but I can understand the opposing view. Is this something you'd like me to change?

bee-common/bee-common/Cargo.toml
zesterer: Will do

bee-common/bee-common/src/event.rs
zesterer: Will do

bee-common/bee-common/src/event.rs
zesterer: Whoops. Will fix.

bee-common/bee-common/src/event.rs
zesterer: Will change them.

bee-common/bee-common/src/event.rs
thibault-martinez: Not necessarily no, let's see what the others think

bee-common/bee-common/src/event.rs
zesterer: Nothing specific. It just felt weird (as you mentioned in a comment before) to make the assumption that handlers would be tied to workers/specific types.

bee-common/bee-common/src/event.rs
zesterer: No, it's the same type. Although even if it was, it wouldn't matter: the effect would be the same.

thibault-martinez: Interesting, I would have thought since it's generic, that would be a different type every time.

bee-common/bee-common/src/event.rs
Alex6323: I think I'ld either call it `EventBus` as well, or rename the module to `bus.rs`, and prevent to use the term `event` like here:
`pub fn dispatch<E: Any>(&self, event: E) {`.
Instead I'ld use `message` and the generic parameter `M`, or `data`/`D`. We **use** this bus for an event system like we use a `Dashmap` here to store the listeners, but it's implementation is more generic, as Josh said, it could also model some central communication system, and the more general term for the thing that gets sent around in such a system is a `message`, or plain old `data`.

bee-common/bee-common/src/event.rs
Alex6323: Are we sure we want to use `DashMap`?

bee-common/bee-common/src/event.rs
Alex6323: I suggest:
`add_listener` (for the one tied to a custom type) and `add_static_listener` (for the hidden type)

bee-common/bee-common/src/event.rs
samuel-rufi: I really don't have a strong opinion about this, but `EventBus` is more expressive and can be recognized better, so i am fine with "EventBus" if you guys want to use that.

> Instead I'ld use message and the generic parameter

i think the term `event` is generic enough to reflect this kind of communication systems (independent of low-level or high-level)
i would even prefer to avoid using the term message, since we have bee-message and  protocol-messages
and having bus-messages might even make more confusion

bee-common/bee-common/src/event.rs
samuel-rufi: Agree with Alex, maybe we should avoid using DashMap for now till the new release?

bee-common/bee-common/src/event.rs
Alex6323: long story short: there is no need to call it or the enclosing module `EventBus`/`event.rs` at all. We use it as such, but it could be used for something else as well.

bee-common/bee-common/src/event.rs
thibault-martinez: To be fair, most of the locks we thought were due to Dashmap ended up being something really different. To this day, I am not sure we ever had a lock due to Dashmap. The whole protocol uses thousands of operations a second and we had no lock in ages.

thibault-martinez: Not saying we should absolutely keep it, but I am not sure we're having issues with it.

bee-common/bee-common/src/event.rs
Alex6323: tbh, I am not in favor of using `message` either. Look at this `Bus` implementation, where it's just a `val` (https://docs.rs/bus/2.2.3/src/bus/lib.rs.html#335). What we actually have here is some sort of SPMC (broadcast) channel where the values being sent can by of `Any` type.

zesterer: So what is the conclusion of this discussion?

thibault-martinez (approved):

samuel-rufi (approved):

Alex6323 (approved):

PR #148, opened 2020-12-02
Trigger "LatestSolidMilestoneChanged" event only when a milestone is solid and valid

samuel-rufi:
# Description of change

Currently, the `LatestSolidMilestoneChanged` event might get triggered in the `PropagatorWorker` **or** in the `MilestoneValidatorWorker`. Since these two workers are working independently there might be the case that the `LatestSolidMilestoneChanged`  gets triggered even if the milestone was not validated yet. 

A `LatestSolidMilestoneChanged` event only should get triggered when the milestone is solid and valid.

This PR's solves exactly that: the  `LatestSolidMilestoneChanged` will only get fired when all conditions are met.
A `SolidMilestoneAnnouncerWorker` waits for the conditions to be full-filled and fires the event.

## Type of change

only self-review yet

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have updated the CHANGELOG.md, if my changes are significant enough

### DIFF
--- a/bee-protocol/src/protocol/protocol.rs
+++ b/bee-protocol/src/protocol/protocol.rs
@@ -29,6 +29,7 @@ use futures::channel::oneshot;
 use log::{debug, error, info};
 use tokio::task::spawn;
 
+use crate::worker::SolidMilestoneAnnouncerWorker;
 use std::sync::Arc;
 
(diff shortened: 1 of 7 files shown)


samuel-rufi:
It appeared to be fine since the milestone flag will only get set by the milestone validator.

PR #156, opened 2020-12-15
feat(logger): add `filters` option

lucasfernog:
# Description of change

Adds filter support to the bee-common logger. 

## Type of change

With my own logger usage, with and without the filters array.

## Change checklist

### DIFF
--- a/bee-common/bee-common/src/logger/config.rs
+++ b/bee-common/bee-common/src/logger/config.rs
@@ -22,6 +22,8 @@ pub struct LoggerOutputConfigBuilder {
     name: Option<String>,
     /// Log level of an output.
     level: Option<LevelFilter>,
+    /// Log filters of the output.
+    filters: Option<Vec<String>>,
 }
 
 impl LoggerOutputConfigBuilder {
@@ -42,11 +44,19 @@ impl LoggerOutputConfigBuilder {
         self
     }
 
+    /// Sets a collection of filters of a logger output.
+    /// A message is logged only if one of the filters is part of the log's metadata target.
+    pub fn filters(mut self, filters: &[&str]) -> Self {
+        self.filters.replace(filters.iter().map(|f| f.to_string()).collect());
+        self
+    }
+
     /// Builds a logger output configuration.
     pub fn finish(self) -> LoggerOutputConfig {
         LoggerOutputConfig {
             name: self.name.unwrap_or_else(|| DEFAULT_OUTPUT_NAME.to_owned()),
             level: self.level.unwrap_or(DEFAULT_OUTPUT_LEVEL),
+            filters: self.filters,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


bee-common/bee-common/src/logger/config.rs
```diff
@@ -58,6 +70,8 @@ pub struct LoggerOutputConfig {
     pub(crate) name: String,
     /// Log level of an output.
     pub(crate) level: LevelFilter,
+    /// Log filters of the output.
+    pub(crate) filters: Option<Vec<String>>,
```
Alex6323: IMHO with another way of filtering the log output we should refactor the naming a bit. I'ld rename `level` to `level_filter` and `filters` to `target_filters`. I think I'ld also introduce a typedef `type TargetFilters = Vec<String>`, but you can ignore this if you don't like to do this. I'ld also rename all the fields accordingly. They are not that short anymore, but I think it helps with clarity.

Alex6323 (commented):
Good and helpful PR. Once merged I would update Bee's CLI so that we can easily make use of this added filter option.

Alex6323 (dismissed):

bee-common/bee-common/src/logger/config.rs
```diff
@@ -57,7 +70,9 @@ pub struct LoggerOutputConfig {
     /// Name of an output file, or `stdout` for standard output.
     pub(crate) name: String,
     /// Log level of an output.
-    pub(crate) level: LevelFilter,
+    pub(crate) level_filter: LevelFilter,
+    /// Log filters of the output.
+    pub(crate) target_filters: Option<Vec<String>>,
```
thibault-martinez: `Option` is needed for the builder to inform serde that the field is optional, but IMO in the built conf we should just have a vector, possibly empty. Option makes it more complex than it should be.

bee-common/bee-common/src/logger/config.rs
```diff
@@ -21,7 +21,9 @@ pub struct LoggerOutputConfigBuilder {
     /// Name of an output file, or `stdout` for standard output.
     name: Option<String>,
     /// Log level of an output.
```
thibault-martinez:
```suggestion
    /// Log level filter of an output.
```

bee-common/bee-common/src/logger/config.rs
```diff
@@ -21,7 +21,9 @@ pub struct LoggerOutputConfigBuilder {
...
-    level: Option<LevelFilter>,
+    level_filter: Option<LevelFilter>,
+    /// Log filters of the output.
```
thibault-martinez:
```suggestion
    /// Log target filters of an output.
```

thibault-martinez (commented):
Nice feature!

thibault-martinez (approved):

Alex6323 (approved):

samuel-rufi (approved):

PR #171, opened 2021-01-05
Network cleaning

Alex6323:
This PR:
* adds documentation to bee-network such that it satisfies #[warn(missing_docs)]
* fixes all wawrnings such that it satisfies #[deny(warnings)]
* reduces the necessary configuration to a minimum
* changes the reconnect interval to 30s (same as Hornet)
* uses the `unknown_peers_limit` from bee-peering
* refactors some function names
* fixes and refactors the example code, so that it uses the appropriate node related traits


bee-node/src/logger.rs
```diff
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{node::BeeNodeBuilder, storage::Backend};
-
```
thibault-martinez: Can you put it back :p ?

Alex6323: sure :joy:

thibault-martinez (approved):

PR #173, opened 2021-01-08
Add 'bee-network' crate'

Alex6323:
# Description of change

Adds a networking layer to the Bee framework.

Be sure to reference any related issues by adding `fixes issue #`.

## Type of change

## Change checklist


bee-network/examples/common/args.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    #[structopt(short = "p", long = "peers")]
+    peer_addresses: Vec<String>,
+
+    #[structopt(short = "m", long = "msg")]
+    message: String,
+}
+
+impl Args {
+    pub fn into_config(self) -> ExampleConfig {
```
zesterer: Maybe `impl From<Args> for ExampleConfig`?

zesterer (changes requested):

bee-network/examples/common/args.rs
Alex6323: sure, can do that :+1:

bee-network/examples/common/utf8msg.rs
```diff
@@ -0,0 +1,33 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::fmt;
+
+#[derive(Clone)]
+pub struct Utf8Message {
```
zesterer, line 7: What's the purpose of this type? It seems to fairly transparently wrap `String::from_utf8` and `str::as_bytes`.

zesterer: What's the purpose of this type? It seems to fairly transparently wrap String::from_utf8 and str::as_bytes.

bee-network/examples/gossip.rs
```diff
@@ -0,0 +1,462 @@
...
+        }
+    }
+
+    info!("[EXAMPLE] ...finished.");
+
+    node.run().await;
+}
+
+#[allow(clippy::type_complexity)]
+struct ExampleNode {
```
zesterer, line 84: I wonder whether we could use an actual instance of the bee node here by including `bee-node` as a dev dependency :thinking:

bee-network/src/config.rs
```diff
@@ -0,0 +1,67 @@
...
+impl NetworkConfigBuilder {
+    /// Creates a new default builder.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Specifies a bind address.
+    pub fn bind_address(mut self, address: &str) -> Self {
+        self.bind_address
+            .replace(Multiaddr::from_str(address).unwrap_or_else(|e| panic!("Error parsing address: {:?}", e)));
```
zesterer, line 32: This is the same behaviour as `.expect` (it will attach the error also)

bee-network/src/conns/dial.rs
```diff
@@ -0,0 +1,177 @@
...
+    banned_addrs: &BannedAddrList,
+    banned_peers: &BannedPeerList,
+) -> Result<(), Error> {
+    // Prevent dialing oneself.
+    if peer_id == &local_keys.public().into_peer_id() {
+        return Err(Error::DialedSelf(peer_id.short()));
+    }
+
+    // Prevent duplicate connections.
+    if let Ok(connected) = peers.is(peer_id, |_, state| state.is_connected()).await {
```
zesterer: `if let Ok(true)` is a valid pattern and avoids the nested branch

bee-network/src/conns/dial.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    // Prevent connecting to dishonest peers or peers we have no up-to-date information about.
+    if &id != peer_id {
+        return Err(Error::PeerIdMismatch {
+            expected: peer_id.to_string(),
+            received: id.to_string(),
+        });
+    }
+
+    let peer_id = id;
```
zesterer: This realiasing is a little confusing. I recommend using a name like `transport_id` instead of `id` and then using that throughout the remaining code.

bee-network/src/conns/dial.rs
```diff
@@ -0,0 +1,177 @@
...
+        muxer,
+        Origin::Outbound,
+        internal_event_sender.clone(),
+    )
+    .await?;
+
+    Ok(())
+}
+
+#[inline]
```
zesterer: `#[inline]` attributes are generally unnecessary, with the exception of non-polymorphic publicly exported functions that would otherwise not be monomorphized. The compiler is in a better place to make a judgement about what should be inlined or not because it knows what `info!` expands to and whether it's likely to result in code bloat.

bee-network/src/conns/dial.rs
```diff
@@ -0,0 +1,177 @@
...
+    banned_addrs: &BannedAddrList,
+    banned_peers: &BannedPeerList,
+) -> Result<(), Error> {
+    // Prevent dialing listen addresses.
+    if LISTEN_ADDRESSES.read().unwrap().contains(address) {
+        return Err(Error::DialedOwnAddress(address.clone()));
+    }
+
+    // Prevent dialing banned addresses.
+    if banned_addrs.contains(&address.to_string()) {
```
zesterer: I wonder whether the `banned_addrs` list should first be parsed into a list of `Multiaddr` To speed up comparison and avoid any address normalisation issues (I believe there are multiple ways to textually express an IPv6 address, for example).

bee-network/src/conns/dial.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    let (peer_id, muxer) = build_transport(local_keys)
+        .map_err(|_| Error::CreatingTransportFailed)?
+        .dial(address.clone())
+        .map_err(|_| Error::DialingFailed(address.clone()))?
+        .await
+        .map_err(|_| Error::DialingFailed(address.clone()))?;
+
+    // Prevent duplicate connections.
+    if let Ok(connected) = peers.is(&peer_id, |_, state| state.is_connected()).await {
```
zesterer: `if let Ok(true) = ...`?

bee-network/src/conns/dial.rs
```diff
@@ -0,0 +1,177 @@
...
+        muxer,
+        Origin::Outbound,
+        internal_event_sender.clone(),
+    )
+    .await?;
+
+    Ok(())
+}
+
+#[inline]
```
zesterer: -`#[inline]`

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+    },
+};
+
+type ListenerUpgrade = Pin<Box<(dyn Future<Output = Result<(PeerId, StreamMuxerBox), io::Error>> + Send + 'static)>>;
+type PeerListener = Pin<Box<dyn Stream<Item = Result<ListenerEvent<ListenerUpgrade, io::Error>, io::Error>> + Send>>;
+
+pub static NUM_LISTENER_EVENT_PROCESSING_ERRORS: AtomicUsize = AtomicUsize::new(0);
+
+lazy_static! {
+    pub static ref LISTEN_ADDRESSES: Arc<RwLock<Vec<Multiaddr>>> = Arc::new(RwLock::new(Vec::new()));
```
zesterer, line 41: Very nervous about global state like this. Could you add a comment justifying its existence?

As an aside, `RwLock` can have poor performance semantics, particularly if reading and writing happen with comparable frequency. Consider switching to a `Mutex`.

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+            ..
+        } = config;
+
+        node.spawn::<Self, _, _>(|shutdown| async move {
+            info!("Listener started.");
+
+            let mut incoming = ShutdownStream::new(shutdown, peer_listener);
+
+            while let Some(Ok(listener_event)) = incoming.next().await {
+                //
```
zesterer: Empty comment?

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+                if let Some((upgrade, peer_address)) = listener_event.into_upgrade() {
+                    // TODO: try again to move this block into its own function (beware: lifetime issues ahead!!!)
+
+                    // Prevent accepting from banned addresses.
+                    let peer_address_str = peer_address.to_string();
+                    if banned_addrs.contains(&peer_address_str) {
+                        trace!("Ignoring peer. Cause: {} is banned.", peer_address_str);
```
zesterer: Does `MultiAddr` not implement `Display`? The `.to_string()` seems to imply that it does. If so, you could display it directly and remove the additional `peer_address_str` binding.

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+                        NUM_LISTENER_EVENT_PROCESSING_ERRORS.fetch_add(1, Ordering::Relaxed);
```
zesterer, line 125: Is this atomic used for debugging or something?

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+                        Ok(u) => u,
+                        Err(_) => {
+                            trace!("Ignoring peer. Cause: Handshake failed.");
+                            NUM_LISTENER_EVENT_PROCESSING_ERRORS.fetch_add(1, Ordering::Relaxed);
+                            continue;
+                        }
+                    };
+
+                    // Prevent accepting duplicate connections.
+                    if let Ok(connected) = peers.is(&peer_id, |_, state| state.is_connected()).await {
```
zesterer: `if let Ok(true) = ...`

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+                    }
+
+                    // Prevent accepting banned peers.
+                    if banned_peers.contains(&peer_id) {
+                        trace!("Ignoring peer. Cause: {} is banned.", peer_id);
+                        NUM_LISTENER_EVENT_PROCESSING_ERRORS.fetch_add(1, Ordering::Relaxed);
+                        continue;
+                    }
+
+                    let peer_info = if let Ok(peer_info) = peers.get_info(&peer_id).await {
```
zesterer, line 152: This code seems conceptually very similar to that in `dial.rs`. Is abstracting this functionality an option?

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+                    .await
+                    {
+                        error!("Error occurred during upgrading the connection. Cause: {}", e);
+                        NUM_LISTENER_EVENT_PROCESSING_ERRORS.fetch_add(1, Ordering::Relaxed);
+                        continue;
+                    }
+                }
+            }
+
+            info!("Listener stopped.")
```
zesterer: Although the types match, not having a semicolon after this `info!` misleadingly looks like it's returning something important at a glance.

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+        //                     continue;
+        //                 }
+        //             } else {
+        //                 error!("Fatal: Listener event stream stopped.");
+        //                 NUM_LISTENER_EVENT_PROCESSING_ERRORS.fetch_add(1, Ordering::Relaxed);
+        //                 break;
+        //             }
+        //         },
+        //     }
+        // }
```
zesterer: Unused code?

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+        Ok(Self::default())
+    }
+
+    async fn stop(self, _node: &mut N) -> Result<(), Self::Error> {
+        info!("Stopping spawned tasks...");
+        Ok(())
+    }
+}
+
+#[inline]
```
zesterer: -`#[inline]`

bee-network/src/conns/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+use libp2p::core::{
+    muxing::{event_from_ref_and_wrap, outbound_from_ref_and_wrap, StreamMuxerBox},
+    upgrade,
+};
+use log::*;
+use tokio::task::JoinHandle;
+
+use std::{fmt, sync::Arc};
+
+const MSG_BUFFER_SIZE: usize = 10000;
```
zesterer: Somewhat undescriptive constant quite far from its usage. Consider moving it within its usage function with a comment justifying this value in particular.

bee-network/src/conns/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+) -> Result<(), Error> {
+    let muxer = Arc::new(muxer);
+    let (message_sender, message_receiver) = peers::channel();
+
+    let internal_event_sender_clone = internal_event_sender.clone();
+
+    let substream = match origin {
+        Origin::Outbound => {
+            let outbound = outbound_from_ref_and_wrap(muxer)
+                .fuse()
```
zesterer, line 44: What does this `.fuse()` do?

bee-network/src/conns/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+                    }
+
+                }
+                recv_result = recv_message(&mut substream, &mut buffer).fuse() => {
+                    trace!("Incoming substream event.");
+                    match recv_result {
+                        Ok(num_read) => {
+                            if let Err(e) = process_read(peer_id.clone(), num_read, &mut internal_event_sender, &buffer).await
+                            {
+                                error!("{:?}", e);
```
zesterer: Consider adding context to this error.

bee-network/src/conns/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+                        Err(e) => {
+                            debug!("{:?}", e);
+
+                            if let Err(e) = internal_event_sender
+                                .send(InternalEvent::ConnectionDropped {
+                                    peer_id: peer_id.clone(),
+                                })
+                                .map_err(|_| Error::InternalEventSendFailure("ConnectionDropped"))
+                            {
+                                error!("{:?}", e);
```
zesterer: More context?

bee-network/src/conns/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+
+async fn process_read(
+    peer_id: PeerId,
+    num_read: usize,
+    internal_event_sender: &mut InternalEventSender,
+    buffer: &[u8],
+) -> Result<(), Error> {
+    // Allocate a properly sized message buffer
+    let mut message = vec![0u8; num_read];
+    message.copy_from_slice(&buffer[0..num_read]);
```
zesterer: `buffer[..num_read].to_vec()` would work fine here

bee-network/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+use bee_runtime::node::{Node, NodeBuilder};
+
+use libp2p::identity;
+use log::info;
+use tokio::sync::mpsc::UnboundedReceiver;
+
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
+
+pub(crate) static RECONNECT_INTERVAL_SECS: AtomicU64 = AtomicU64::new(DEFAULT_RECONNECT_INTERVAL_SECS);
+pub(crate) static NETWORK_ID: AtomicU64 = AtomicU64::new(0);
```
zesterer, line 51: Consider removing this static so that multiple instances don't conflict within the same program.

bee-network/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    let local_keys = identity::Keypair::Ed25519(local_keys);
+    let local_id = PeerId::from_public_key(local_keys.public());
+    info!("Own peer id: {}", local_id);
+
+    let (command_sender, command_receiver) = commands::channel();
+    let (event_sender, event_receiver) = events::channel::<Event>();
+    let (internal_event_sender, internal_event_receiver) = events::channel::<InternalEvent>();
+
+    let banned_addrs = BannedAddrList::new();
```
zesterer, line 74: It's not immediate clear that this type is actually a handle to a ban list rather than a self-owned type and this could result in refactoring errors. Perhaps the `Arc` should be removed from the type alias and then it can be used like `Arc<BannedAddrList>` elsewhere to make it clear that it's shared state?

bee-network/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+    let conn_manager_config = ConnectionManagerConfig::new(
+        local_keys,
+        config.bind_address.clone(),
+        peers,
+        banned_addrs,
+        banned_peers,
+        internal_event_sender,
+    )
+    .unwrap_or_else(|e| {
+        panic!("Fatal error: {}", e);
```
zesterer, line 98: Context like `Fatal error when initiating networking: ` would be useful for users.

bee-network/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+/// A trait specifically there to create shorter peer ids for better readability in logs and user interfaces.
+pub trait ShortId
+where
+    Self: ToString,
+{
+    /// The length of the shortened peer id.
+    const SHORT_LENGTH: usize;
+
+    /// Creates a shorter - more readable - id from the original.
+    fn short(&self) -> String;
```
zesterer, line 120: Returning a `Cow<str>` here would let `<PeerId as ShortId>::short` run without allocating.

bee-network/src/peers/banned.rs
```diff
@@ -0,0 +1,84 @@
...
+
+pub type BannedPeerList = BannedList<PeerId>;
+pub type BannedAddrList = BannedList<String>;
+
+#[derive(Clone, Default)]
+pub struct BannedList<T: Hash + Eq>(Arc<DashSet<T>>);
+
+impl BannedList<PeerId> {
+    pub fn new() -> Self {
+        Self(Arc::new(DashSet::with_capacity(DEFAULT_BANNED_PEER_CAPACITY)))
```
zesterer: Any particular reason for a non-zero initial capacity? Bear in mind that if this capacity isn't used, then this means allocation overhead for no reason.

bee-network/src/peers/banned.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub fn contains(&self, value: &T) -> bool {
+        self.0.contains(value)
+    }
+
+    pub fn remove(&self, value: &T) -> bool {
+        self.0.remove(value).is_some()
+    }
+
+    #[cfg(test)]
+    pub fn count(&self) -> usize {
```
zesterer: `len` seems like the more natural name here, although this is only test code. That said, I don't see why this couldn't be useful in non test code.

bee-network/src/peers/list.rs
```diff
@@ -0,0 +1,217 @@
...
+
+use std::{
+    collections::HashMap,
+    sync::{atomic::Ordering, Arc},
+};
+
+const DEFAULT_PEERLIST_CAPACITY: usize = 8;
+
+#[derive(Clone, Default)]
+pub struct PeerList(Arc<RwLock<HashMap<PeerId, (PeerInfo, PeerState)>>>);
```
zesterer, line 20: `BannedList` uses `DashSet` but `PeerList` uses `RwLock<HashMap>`? Is this due to the deadlocking issue?

bee-network/src/peers/list.rs
```diff
@@ -0,0 +1,217 @@
...
+                    Some(kv.0.clone())
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<PeerId>>()
+            .into_iter()
+    }
+
+    pub async fn count_if(&self, predicate: impl Fn(&PeerInfo, &PeerState) -> bool) -> usize {
```
zesterer, line 150: Docs for these functions would be very useful.

zesterer: Also, this sort of thing makes me desperately want monads in Rust :D

bee-network/src/peers/list.rs
```diff
@@ -0,0 +1,217 @@
...
+            return;
+        };
+
+        if remove {
+            let _ = self.0.write().await.remove(id);
+        }
+    }
+
+    #[allow(dead_code)]
+    pub async fn clear(&self) {
```
zesterer, line 175: In what circumstance would this occur? If it only ever occurs with synchronisation, you could change the method receiver to `&mut self` and use `self.0.get_mut().clear();`

bee-network/src/peers/list.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct PeerInfo {
+    pub address: Multiaddr,
+    pub alias: Option<String>,
+    pub relation: PeerRelation,
+}
+
+macro_rules! impl_is_relation {
```
zesterer, line 188: The code this macro effectively expands to is 5 lines long:

```rust
impl PeerInfo {
	pub fn is_known(&self) -> bool { self.relation.is_known() }
	pub fn is_unknown(&self) -> bool { self.relation.is_unknown() }
	pub fn is_discovered(&self) -> bool { self.relation.is_discovered() }
}
```

and yet the macro + macro invocations is 14 lines long and significantly more complicated to read.

I recommend just implementing the methods by hand.

Alternatively, you could `impl Deref<Target = PeerRelation> for PeerInfo { ... }`.

bee-network/src/peers/manager.rs
```diff
@@ -0,0 +1,564 @@
...
+                // of those.
+                for peer_id in peers
+                    .iter_if(|info, state| info.is_known() && state.is_disconnected())
+                    .await
+                {
+                    if let Err(e) = internal_event_sender
+                        .send(InternalEvent::ReconnectScheduled { peer_id })
+                        .map_err(|_| Error::InternalEventSendFailure("ReconnectScheduled"))
+                    {
+                        warn!("{:?}", e)
```
zesterer, line 173: User-visible warning: more context?

bee-network/src/peers/manager.rs
```diff
@@ -0,0 +1,564 @@
...
+        }
+        Command::ConnectPeer { id } => {
+            let local_keys = local_keys.clone();
+            let peers = peers.clone();
+            let banned_addrs = banned_addrs.clone();
+            let banned_peers = banned_peers.clone();
+            let event_sender = event_sender.clone();
+            let internal_event_sender = internal_event_sender.clone();
+
+            tokio::spawn(async move {
```
zesterer, line 247: I must admit, I'm really not happy with how we're forced to use `tokio::spawn` in a bunch of places to spawn tasks that aren't bound to the runtime at all. It's not your fault of course (I've been forced to do the same) but I think I'm going to work on a better solution to this.

bee-network/src/peers/manager.rs
```diff
@@ -0,0 +1,564 @@
...
+        }
+        Command::UpdateRelation { id, relation } => {
+            peers.update_relation(&id, relation).await?;
+        }
+    }
+
+    Ok(())
+}
+
+#[inline]
```
zesterer: - `#[inline]`

bee-network/src/peers/manager.rs
```diff
@@ -0,0 +1,564 @@
...
+            .map_err(|_| Error::EventSendFailure("CommandFailed"))?;
+
+        return Err(e);
+    }
+
+    Ok(())
+}
+
+#[inline]
+async fn send_message(message: Vec<u8>, to: &PeerId, peers: &PeerList) -> Result<(), Error> {
```
zesterer, line 554: Seems like strange function that's more difficult to read, if anything, than the code it wraps.

bee-network/src/peers/manager.rs
```diff
@@ -0,0 +1,564 @@
...
+    peers.send_message(message, to).await
+}
+
+#[inline]
```
zesterer: -`#[inline]`

bee-network/src/protocols/gossip.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Default, Debug, Clone)]
+pub struct GossipProtocol;
+
+impl UpgradeInfo for GossipProtocol {
+    type Info = Vec<u8>;
+    type InfoIter = iter::Once<Self::Info>;
+
+    fn protocol_info(&self) -> Self::InfoIter {
+        iter::once(
+            format!("/iota-gossip/{}/1.0.0", NETWORK_ID.load(Ordering::Relaxed))
```
zesterer, line 28: Global state :\

bee-network/src/transport.rs
```diff
@@ -0,0 +1,31 @@
...
+        .expect("error creating noise keys");
+
+    let tcp = tcp::TokioTcpConfig::new().nodelay(true);
+    let transport = dns::DnsConfig::new(tcp)?;
+
+    Ok(transport
+        .upgrade(upgrade::Version::V1)
+        .authenticate(noise::NoiseConfig::xx(noise_keys).into_authenticated())
+        .multiplex(mplex::MplexConfig::new())
+        // .multiplex(SelectUpgrade::new(yamux::Config::default(), mplex::MplexConfig::new()))
```
zesterer, line 28: Unused?

zesterer (changes requested):

bee-network/examples/common/logger.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use env_logger::fmt::Color;
+use log::{debug, error, info, log_enabled, trace, warn};
+
+use std::io::Write;
+
+pub fn init(level_filter: log::LevelFilter) {
```
samuel-rufi, line 9: Maybe this filter would be helpful for other crates too, should this be moved out?

bee-network/examples/common/utf8msg.rs
samuel-rufi: It makes it quite easy to work with utf8 strings (you can't do things wrong), but agree maybe we could just use the `String` type here.

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+                    // Prevent accepting from banned addresses.
+                    let peer_address_str = peer_address.to_string();
+                    if banned_addrs.contains(&peer_address_str) {
+                        trace!("Ignoring peer. Cause: {} is banned.", peer_address_str);
+                        NUM_LISTENER_EVENT_PROCESSING_ERRORS.fetch_add(1, Ordering::Relaxed);
+                        continue;
+                    }
+
+                    let (peer_id, muxer) = match upgrade.await {
+                        Ok(u) => u,
```
samuel-rufi, line 130: maybe rename "u" -> "peer"?

bee-network/src/conns/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        let mut fused_message_receiver = message_receiver.fuse();
+        let mut buffer = vec![0u8; MSG_BUFFER_SIZE];
+
+        loop {
+            select! {
+                message = fused_message_receiver.next() => {
+                    trace!("Outgoing message channel event.");
+                    if let Some(message) = message {
+                        if let Err(e) = send_message(&mut substream, &message).await {
+                            error!("{:?}", e);
```
samuel-rufi: Also more context to this error.

bee-network/src/peers/list.rs
samuel-rufi: We are using the new 4.0 of dashmap, so the deadlock behaviour should be fixed if I am not mistaken now

bee-network/examples/common/utf8msg.rs
Alex6323: This type has no real purpose except for vaguely reminding the user, that the nodes built with this framework send *some* kind of message over the wire. In this case the example node is only capable of sending `Utf8Message`s.

bee-network/examples/gossip.rs
Alex6323: good point :+1:

Alex6323: however, `bee-node` is not released yet.

bee-network/examples/common/utf8msg.rs
Alex6323: I have no strong opinion on that, and can remove it. @thibault-martinez wdyt?

bee-network/src/config.rs
Alex6323: yea, the "behaviour" will be identical, but are there any benefits of one over the other except for `expect` being easier to read? I thought `unwrap_or_else` should be preferred due to its laziness.

bee-network/src/conns/dial.rs
Alex6323: it does, but the compiler will easily detect that as well and remove the unnecessary nested branch, but you tell me :stuck_out_tongue: 
I wrote it that way because I personally think `If let Ok(connected) ...` is more readable than `If let Ok(true) ....`.

bee-network/src/conns/dial.rs
Alex6323: `transport_id` would be a confusing term, as there is no such thing as a transport id. Would you be fine with just using `id` for the rest of the code?

bee-network/src/peers/manager.rs
Alex6323: removed

bee-network/src/peers/manager.rs
Alex6323: removed

bee-network/src/peers/manager.rs
Alex6323: sure, you're right. The reason is this strange need for duality. There is a `recv_message` function below, so.... How strongly do you want me to change that? I feel like a lot of readers like to have a `send_message` function when there is also a `recv_message` function, but I could be wrong.

bee-network/src/protocols/gossip.rs
Alex6323: yea, you mean the `NETWORK_ID`? That's not really an issue, because it's only set once during initialization. We could think about using a different primitive, that is more appropriate in such a scenario, but the alternative would be to have bigger signatures as this value would have to be passed in a chain, which increases complexity and decreases readibility. Or is your concern headed towards another angle?

bee-network/src/transport.rs
Alex6323: unfortunately I had to do this, because there was an issue with `libp2p` where the network layer couldn't establish a transport layer with a peer using `yamux` as its multiplexer. Only with `Mplex` the code worked as expected. We should, however, come back to this issue at a later point in time, and investigate. That's why I left this line out-commented in there. Once this issue is fixed, instead of thinking long about how to provide both muxers again, I can just uncomment this line and remove the previous.

Alex6323: The reason why I think this is probably something that should be fixed by one of our dependencies is that `libp2p` understands itself as an abstraction layer over the concrete multiplexer being used, or in other words: my user code should NOT depend on the multiplexer I use.

bee-network/examples/common/logger.rs
Alex6323: I don't know what you mean, Sam. The `logger` impl in `bee-common` let's you configure the log level and even allows you to define target filters. So I am actually thinking about removing this logger module entirely.

bee-network/src/conns/manager.rs
Alex6323: the `u` stands for `upgrade` which we are awaiting. Are you also fine with me renaming it to `upgrade`? I think `peer` would be not so fitting, because we not only receive a peer (or to be more exact: its id), but also get a back a *multiplexer*.

bee-network/src/conns/mod.rs
Alex6323: added more context, thx

Alex6323:
@rufsam I don't know why github won't give me a Reply button on your `DashMap 4.0` comment. Anyway....

The reason why I did not change back to `DashMap` is that I want to have a profound reason to introduce dependencies on other crates. If I don't think it's absolutely necessary to achieve my goal I would rather stick to the standard library.

So if you have a compelling reason, why I should use `DashMap` here, I am happy to change back to it :smile: 

Having said that...I am still using DashMap for other lists (as Josh also pointed out), so I guess I'ld like to change that, too (using standard library types, bc. if the standard lib isn't considered even in such circumenstance I'ld have to start wondering, when they're useful at all :stuck_out_tongue: )

bee-network/src/conns/dial.rs
Alex6323: good point, I look into that. Actually there was a reason why I need to switch to `String`, which I really didn't like at the time.

bee-network/src/conns/dial.rs
Alex6323: yea, as already mentioned: `if let Ok(connected) ...` reads better than `if let Ok(true) ...` IMHO

bee-network/src/conns/manager.rs
Alex6323: yea, I see your point, and I am aware of this issue regarding `RwLock`s. The listen addresses are only set at initialization, and usually not during the runtime of the node. I think I'll switch to `OnceCell`. Wdyt?

The reason is that multiple task need to have read access to that `Vec` of `Multiaddr`esses. Doing it non-globally would mean passing a copy to each of those tasks managing the synchronization manually. This would increase code complexity both in terms of code lines and fn signatures.

bee-network/src/conns/manager.rs
Alex6323: this comment is there to remind the reader to take a break and get a coffee or tea before reading on....

Alex6323: jk, it's about to get removed :smile:

bee-network/src/conns/manager.rs
Alex6323: removed

bee-network/src/conns/manager.rs
Alex6323: removed

bee-network/src/conns/mod.rs
Alex6323: good point

bee-network/src/conns/manager.rs
Alex6323: added

bee-network/src/conns/manager.rs
samuel-rufi: Yes that sounds good to me!

bee-network/examples/common/logger.rs
samuel-rufi: Yes, I guess this would be nice.

bee-network/src/conns/dial.rs
Alex6323: anyway, I think reading the full line is still pretty readable, so I make the change :smile:

bee-network/examples/common/utf8msg.rs
Alex6323: After discussion this on Discord, we decided to keep the examples as simple as possible. Therefore it's going to be removed.

bee-network/src/conns/manager.rs
Alex6323: those atomics are used to eventually provide network stats

bee-network/src/conns/manager.rs
Alex6323: yea, maybe. I tried that already, but the order is slightly different due to the fact, that in one case you have the actuall peer id of the peer, and in the other case you start with an address only (mulitaddr).

bee-network/src/peers/list.rs
Alex6323: I switched to `RwLock<HashMap<_, _>>`, whereby RwLock is the async version from tokio. Until DashMap exhibits deadlocks (at it still does so even in version 4.0) I'ld rather not use it. Any objections to that?

bee-network/src/peers/list.rs
Alex6323: they are not exported, though. Do you still think, we should document them?

bee-network/src/peers/list.rs
Alex6323: haha, you're right of course. What you see are the remnants of a bigger refactoring. I agree, that still using the macro doesn't make much sense.

bee-network/src/conns/mod.rs
Alex6323: good catch. I think I forgot to remove it. Out of curiousity: Does it hurt to leave it in case the code gets refactored later on where it might be useful?

Alex6323: Can you also have a look at the `Origin::Inbound` match, where there's an actual loop, but I am (for some reason) not fusing there. I probably wanted to fuse there, but I can't remember anymore :thinking:.

bee-network/src/conns/mod.rs
Alex6323: will change

bee-network/src/conns/manager.rs
Alex6323: Thinking about abstracting this I had the idea of defining a `Filter` trait, where each of those checks are implemented as a concrete filter. Those then can be chained together in different order depending on the usecase. This would be very flexible as new filters could easily be added or removed.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
 	"bee-common/bee-common-derive",
 	"bee-crypto",
 	"bee-runtime",
+	"bee-network",
```
thibault-martinez, line 7: Alphabetic order

bee-network/examples/common/utf8msg.rs
thibault-martinez: Is it still planned to remove it ?

bee-network/examples/common/backend.rs
```diff
@@ -0,0 +1,37 @@
...
+    type Error = Infallible;
+
+    async fn start(_: Self::Config) -> Result<Self, Box<dyn Error>> {
+        Ok(DummyBackend)
+    }
+
+    async fn shutdown(self) -> Result<(), Box<dyn Error>> {
+        Ok(())
+    }
+}
```
thibault-martinez: I would put this file in bee-test to allow reusing it on other crate examples.

bee-network/examples/common/logger.rs
thibault-martinez: Not sure what was the conclusion of this, but we already have this piece of code in bee, I would just use it, otherwise we end up rewriting the same stuff for all crate examples.

bee-network/src/config.rs
thibault-martinez: Laziness is desired when you actually have something potentially expensive to evaluate, this is not the case here. So both are pretty much similar.

bee-network/src/peers/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+pub type DataSender = mpsc::UnboundedSender<Vec<u8>>;
+pub type DataReceiver = mpsc::UnboundedReceiver<Vec<u8>>;
+
+pub fn channel() -> (DataSender, DataReceiver) {
+    mpsc::unbounded_channel()
+}
+
+/// Describes the relation with a peer.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum PeerRelation {
```
thibault-martinez, line 25: I would probably put it in its own module tbh.

bee-network/src/peers/banned.rs
```diff
@@ -0,0 +1,95 @@
...
+        self.0.write().await.remove(value)
+    }
+
+    #[allow(dead_code)]
+    pub async fn len(&self) -> usize {
+        self.0.read().await.len()
+    }
+}
+
+#[cfg(test)]
```
thibault-martinez, line 64: Can we put tests of public API in the `tests` folder ?

bee-network/src/conns/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    internal_event_sender
+        .send(InternalEvent::MessageReceived { message, from: peer_id })
+        .map_err(|_| Error::InternalEventSendFailure("MessageReceived"))?;
+
+    Ok(())
+}
+
+/// Describes direction of an established connection.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Origin {
```
thibault-martinez, line 192: Would put in its own module

bee-network/src/conns/manager.rs
```diff
@@ -0,0 +1,217 @@
...
+        Ok(Self::default())
+    }
+
+    async fn stop(self, _node: &mut N) -> Result<(), Self::Error> {
+        info!("Stopping spawned tasks...");
+        Ok(())
+    }
+}
+
+fn log_inbound_connection_success(peer_id: &PeerId, peer_info: &PeerInfo) {
```
thibault-martinez, line 211: Is this being used once ? If yes, I'd probably not do a function, ends up being more verbose than needed.

bee-network/src/config.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use libp2p::Multiaddr;
+use serde::Deserialize;
+
+use std::str::FromStr;
+
+const DEFAULT_BIND_ADDRESS: &str = "/ip4/0.0.0.0/tcp/15600";
+
+pub const DEFAULT_RECONNECT_INTERVAL_SECS: u64 = 30;
+pub const DEFAULT_CONNECTION_TIMEOUT_SECS: u64 = 10;
```
thibault-martinez, line 12: Why are these public ?

bee-network/examples/common/backend.rs
```diff
@@ -0,0 +1,37 @@
...
+pub struct DummyConfigBuilder;
+
+impl Into<DummyConfig> for DummyConfigBuilder {
+    fn into(self) -> DummyConfig {
+        DummyConfig
+    }
+}
+
+#[async_trait]
+impl StorageBackend for DummyBackend {
```
thibault-martinez: Needs the new `size` trait method

Alex6323: :thumbsup:

PR #274, opened 2021-01-22
introduce `children_to_visit` +  `children_visited` set

samuel-rufi:
# Description of change

Currently, the children to visit are filtered in the `update_future_cone` function. If a children was already processed (cone_index check) it was skipped.

With this PR, the children to visit get filtered first in the `update_messages_referenced_by_milestone` function.
Therefore, the `update_future_cone` won't get anymore triggered for the messages that were already processed by the `update_messages_referenced_by_milestone`.
Also, since the `HashSet` does not allow duplicates, the `update_future_cone`  won't get triggered multiple times for the same children.

This can be considered as a little optimization.

This PR furthermore introduces the `children_visited` HashSet which should actually fix the "Find milestone_cone_updater loop issue".

Find milestone_cone_updater loop issue: https://github.com/iotaledger/bee/issues/272

## Type of change

Not finished yet.

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have checked that new and existing unit tests pass locally with my changes
- [X] I have updated the CHANGELOG.md, if my changes are significant enough

### DIFF
--- a/bee-protocol/src/worker/propagator.rs
+++ b/bee-protocol/src/worker/propagator.rs
@@ -53,10 +53,14 @@ async fn propagate<B: StorageBackend>(
             }
 
             // get OTRSI/YTRSI from parents
-            let parent1_otsri = tangle.otrsi(&parent1).await;
-            let parent2_otsri = tangle.otrsi(&parent2).await;
-            let parent1_ytrsi = tangle.ytrsi(&parent1).await;
-            let parent2_ytrsi = tangle.ytrsi(&parent2).await;
+            let p1m = tangle.get_metadata(&parent1).await;
+            let p2m = tangle.get_metadata(&parent2).await;
+            let p1sepi = tangle.get_solid_entry_point_index(&parent1);
+            let p2sepi = tangle.get_solid_entry_point_index(&parent2);
+            let parent1_otsri = p1sepi.or_else(|| p1m?.otrsi());
+            let parent2_otsri = p2sepi.or_else(|| p2m?.otrsi());
+            let parent1_ytrsi = p1sepi.or_else(|| p1m?.ytrsi());
+            let parent2_ytrsi = p2sepi.or_else(|| p2m?.ytrsi());
 
             // get best OTRSI/YTRSI from parents
             // unwrap() is safe because parents are solid which implies that OTRSI/YTRSI values are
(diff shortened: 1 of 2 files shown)


bee-protocol/src/worker/milestone_cone_updater.rs
```diff
@@ -74,21 +74,22 @@ async fn update_messages_referenced_by_milestone<N: Node>(
...
         }
 
         if tangle.is_solid_entry_point(&message_id) {
             continue;
         }
 
-        // maybe the check below is not necessary; all messages from the most recent cone should be present
+        // maybe this check is not necessary; all messages from the most recent cone should be present
         if let Some((parent1, parent2)) = tangle.get(&message_id).await.map(|m| (*m.parent1(), *m.parent2())) {
             // unwrap() is safe since message is present and so is the metadata
```
Alex6323: but what if that message got evicted between both queries? Unlikely it is, but not impossible. There should be an atomic operation available for this.

thibault-martinez: If it's evicted, it would be pulled from the storage again. If you meant pruned, this is almost impossible.

Alex6323: yea, I meant pruned.

Alex6323: I know it's nearly impossible, but I don't like it anyway. There should be an atomic query to get both informations.

thibault-martinez: I actually think there is!

Alex6323: aaah

bee-protocol/src/worker/milestone_cone_updater.rs
```diff
@@ -74,21 +74,22 @@ async fn update_messages_referenced_by_milestone<N: Node>(
...
             if tangle.get_metadata(&message_id).await.unwrap().cone_index().is_some() {
```
Alex6323, line 95: so this whole algorithm starts with the milestone itself, right? Shouldn't we start with the parents of that milestone? Or is this also the first time, where we set that metadata for a milestone?

Alex6323: forget that :smile:

bee-protocol/src/worker/milestone_cone_updater.rs
```diff
@@ -74,21 +74,22 @@ async fn update_messages_referenced_by_milestone<N: Node>(
 ) where
     N::Backend: StorageBackend,
 {
-    let mut to_visit = vec![message_id];
-    let mut visited = HashSet::new();
+    let mut parents_to_visit = vec![message_id];
```
Alex6323: obviously the milestone we start with doesn't need to be visited. So I'ld suggest initializing this vec with its 2 parents.

samuel-rufi: the `cone_index` currently does not get set in tangle::add_milestone() so this would need to be changed too

samuel-rufi: with the latest commit this is changed

Alex6323 (approved):

PR #321, opened 2021-01-27
Fix cone updater

Alex6323:
# Description of change

Please write a summary of your changes and why you made them. 

Be sure to reference any related issues by adding `fixes issue #`.

## Type of change

- Documentation Fix

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist


bee-protocol/src/worker/confirmer.rs
```diff
@@ -0,0 +1,202 @@
...
+) -> HashSet<MessageId> {
+    let mut parents = vec![parent1, parent2];
+    let mut confirmed = HashSet::new();
+
+    while let Some(parent_id) = parents.pop() {
+        // Our stop conditions. Note that the order of calls is important (from cheap to more expensive) for performance
+        // reasons.
+        if confirmed.contains(&parent_id)
+            || tangle.is_solid_entry_point(&parent_id)
+            || tangle.get_metadata(&parent_id).await.unwrap().flags().is_confirmed()
```
samuel-rufi: we assume that the unwrap() call can never fail right? we should have a comment for this just to describe the situation better

bee-protocol/src/worker/confirmer.rs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            continue;
+        }
+
+        tangle
+            .update_metadata(&parent_id, |metadata| {
+                // TODO: Throw one of those indexes away ;)
+                metadata.set_milestone_index(index);
```
samuel-rufi: in the old "milestone_cone_updater" we didn't set the `milestone_index`.
this is something that is done by bee-ledger since it also traverses the past cone (duplicate work)
this might conflict with bee-ledger traversal if we set it here (so maybe we should remove this).

bee-protocol/src/worker/confirmer.rs
```diff
@@ -0,0 +1,202 @@
...
+
+// NOTE: so once a milestone comes in we have to walk the future cones of the root transactions and update their
+// OTRSI and YTRSI; during that time we need to block the propagator, otherwise it will propagate outdated data.
+async fn update_otrsi_ytrsi<B: StorageBackend>(tangle: &MsTangle<B>, confirmed: HashSet<MessageId>) {
+    let mut to_process = confirmed.into_iter().collect::<Vec<_>>();
+    let mut processed = Vec::new();
+
+    while let Some(parent_id) = to_process.pop() {
+        if let Some(children) = tangle.get_children(&parent_id).await {
+            // Unwrap is safe with very high probability.
```
samuel-rufi: more detailed comment please

bee-protocol/src/worker/confirmer.rs
```diff
@@ -0,0 +1,202 @@
...
+        confirmed.insert(parent_id);
+    }
+
+    debug!("Confirmed {} messages.", confirmed.len());
+
+    confirmed
+}
+
+// NOTE: so once a milestone comes in we have to walk the future cones of the root transactions and update their
+// OTRSI and YTRSI; during that time we need to block the propagator, otherwise it will propagate outdated data.
```
samuel-rufi: "we need to block the propagator" is this implemented?

bee-protocol/src/worker/propagator.rs
```diff
@@ -48,50 +49,66 @@ async fn propagate<B: StorageBackend>(
...
                     } else {
+                        metadata.set_otrsi(child_otrsi);
+                        metadata.set_ytrsi(child_ytrsi);
                         None
                     }
                 })
                 .await
                 .expect("Failed to fetch metadata.");
 
+            // Try to propagate as far as possible into the future.
```
samuel-rufi, line 103: Could it be that the propagator and the confirmation worker trying to update the same future cone? Could it be that the propagator is faster? and we traverse the future cone another time in the confirmation worker?

bee-protocol/src/worker/propagator.rs
```diff
@@ -48,50 +49,66 @@ async fn propagate<B: StorageBackend>(
...
             if let Some(msg_children) = tangle.get_children(&message_id).await {
                 for child in msg_children {
                     children.push(child);
                 }
             }
 
-            let _ = tip_tx.send((*message_id, parent1, parent2, index)).await;
+            // Send child to the tip pool.
+            let _ = tip_tx.send((*message_id, parent1, parent2, ms_index_maybe)).await;
```
samuel-rufi: maybe a `if let Err(err) = tip_tx.send((*message_id, parent1, parent2, ms_index_maybe)).await {}` as we do we other channels would be better here

bee-protocol/src/worker/confirmer.rs
Alex6323: nope, the ledger is not setting it

bee-protocol/src/worker/confirmer.rs
Alex6323: nope. I am pretty sure, however, that not pausing it, leads to a data race.

bee-protocol/src/worker/confirmer.rs
Alex6323: haha, that was a fun one, as it actually occured. Humans an probabilities *sigh*

bee-protocol/src/worker/propagator.rs
Alex6323: sure, I left that line untouched. Will change it.

Alex6323: actually not entirely untouched :joy:

bee-protocol/src/worker/confirmer.rs
Alex6323: wrote a comment about it. Definitely something that nees investigation. thx

bee-ledger/src/worker.rs
```diff
@@ -108,7 +108,7 @@ where
         tangle
             .update_metadata(message_id, |message_metadata| {
                 message_metadata.set_conflict(ConflictReason::None as u8);
-                message_metadata.set_milestone_index(metadata.index);
+                // message_metadata.set_milestone_index(metadata.index);
```
thibault-martinez: You can remove them :)

bee-test/src/rand/metadata.rs
```diff
@@ -3,18 +3,22 @@
 
 use crate::rand::{integer::rand_integer, milestone::rand_milestone_index, option::rand_option};
 
-use bee_tangle::{flags::Flags, metadata::MessageMetadata};
+use bee_tangle::{
+    flags::Flags,
+    metadata::{IndexId, MessageMetadata},
+};
+
+use super::message::rand_message_id;
```
thibault-martinez: Include in the existing import

bee-tangle/src/metadata.rs
```diff
@@ -83,35 +82,27 @@ impl MessageMetadata {
...
+    pub fn set_ytrsi(&mut self, ytrsi: IndexId) {
         self.ytrsi = Some(ytrsi);
     }
 
     pub fn confirmation_timestamp(&self) -> u64 {
         self.confirmation_timestamp
     }
 
-    pub fn solidify(&mut self) {
+    pub fn set_solid(&mut self) {
```
thibault-martinez: Not a big fan of the renaming, `set_*` assumes a parameter (like the other methods). This one doesn't have a parameter and does a bit more than just setting the flag.

bee-tangle/src/metadata.rs
```diff
@@ -183,14 +179,77 @@ impl Packable for MessageMetadata {
...
-            ytrsi: Option::<MilestoneIndex>::unpack(reader)?,
+            otrsi: Option::<IndexId>::unpack(reader)?,
+            ytrsi: Option::<IndexId>::unpack(reader)?,
             conflict: u8::unpack(reader)?,
         })
     }
 }
+
+#[derive(Clone, Copy, Debug)]
+pub struct IndexId(pub MilestoneIndex, pub MessageId);
```
thibault-martinez: Can we make the fields private and have getters ? From an external point of view `.0` and `.1` are kinda confusing, you'd have to always go back to the type to know which one is which one. `.index()` and `.message_id()` would help a lot imo.

bee-protocol/src/worker/propagator.rs
Alex6323: yea, both step on each other's toes, so race conditions are very likely to happen. The propagator and the index updater **must not** run at the same time as they change the same metadata. There has to be a differentiation between a normal message and a milestone message:

Normal Message => run the propagator on that message as usual
Milestone Message => run the updater first, and *then* the propagator (but not before or during)

bee-protocol/src/worker/propagator.rs
Alex6323: added a warning

bee-ledger/src/worker.rs
Alex6323: done

bee-tangle/src/metadata.rs
Alex6323: agree, will change

bee-tangle/src/metadata.rs
Alex6323: I though I did check on what `solidify` was actually doing, and it looked to me like it's just setting booleans and flags.

bee-test/src/rand/metadata.rs
Alex6323: sure, will change

Alex6323: done

bee-tangle/src/metadata.rs
Alex6323: changed

bee-tangle/src/metadata.rs
Alex6323: done

bee-protocol/src/worker/propagator.rs
Alex6323: We will add this later in another PR. Do you agree @thibault-martinez ?

thibault-martinez (approved):

samuel-rufi (approved):

PR #339, opened 2021-01-31
8 parents

thibault-martinez:


bee-protocol/src/worker/propagator.rs
```diff
@@ -62,27 +56,27 @@ async fn propagate<B: StorageBackend>(
...
+
+            let mut parent_otrsis = Vec::new();
+            let mut parent_ytrsis = Vec::new();
+
+            for parent in parents.iter() {
+                let (parent_otrsi, parent_ytrsi) = match tangle.get_solid_entry_point_index(parent).await {
+                    Some(parent_sepi) => (IndexId::new(parent_sepi, *parent), IndexId::new(parent_sepi, *parent)),
+                    None => match tangle.get_metadata(parent).await {
+                        Some(parent_md) => (parent_md.otrsi().unwrap(), parent_md.ytrsi().unwrap()),
+                        None => continue,
```
Alex6323: this `continue` now refers to it's parent for-loop but it used to refer to the while-let. Just use loop labels here to fix this (I love Rust for supporting that)

bee-api/bee-rest-api/src/types.rs
```diff
@@ -170,8 +166,7 @@ impl TryFrom<&Message> for MessageDto {
     fn try_from(value: &Message) -> Result<Self, Self::Error> {
         Ok(MessageDto {
             network_id: value.network_id().to_string(),
-            parent_1_message_id: value.parent1().to_string(),
-            parent_2_message_id: value.parent2().to_string(),
+            parents: vec![value.parents()[0].to_string(), value.parents()[1].to_string()],
```
Alex6323: that's probably a bug. From what @Thoralf-M  pointed out in the discussion on Discord (parents now need to be unique, there are cases where a message only has 1 parent like messages referencing the genesis) the index 1 might not exist, hence this code will panic in such a special situation.

samuel-rufi: okey if so, then we need to adjust also `two_non_lazy_tips()` from the urts.rs. Should I take care of this? @thibault-martinez

bee-message/src/message.rs
```diff
@@ -94,8 +92,16 @@ impl Packable for Message {
     fn unpack<R: Read + ?Sized>(reader: &mut R) -> Result<Self, Self::Error> {
         let network_id = u64::unpack(reader)?;
 
-        let parent1 = MessageId::unpack(reader)?;
-        let parent2 = MessageId::unpack(reader)?;
+        let parents_len = u8::unpack(reader)? as usize;
+
+        if parents_len != 2 {
```
Alex6323: use named constant

bee-api/bee-rest-api/src/types.rs
Thoralf-M: Yes, that was from the point where we thought it will be fixed to 2 parents for the first release

thibault-martinez: @rufsam Please do

bee-message/src/message.rs
thibault-martinez: Sure, I'll replace with a 1..8 range anyway

bee-api/bee-rest-api/src/types.rs
samuel-rufi: The REST API of Hornet still returns 2 tips https://github.com/gohornet/hornet/blob/develop/plugins/restapi/v1/types.go#L36
So is this not implemented yet?
Also, how many `MessageId`s should `tangle::get_messages_to_approve()` return? 1? 8?

Thoralf-M: @rufsam look at the other branch https://github.com/gohornet/hornet/blob/feat/multiple-parents/plugins/restapi/v1/types.go

samuel-rufi: okey thanks, seems like Hornet currently returns 4 tips from the tip pool <https://github.com/gohornet/hornet/blob/04a47e97326bd0b74e4b078a720ad54ebf628bea/pkg/tipselect/urts.go#L329>

bee-ledger/src/white_flag.rs
```diff
@@ -236,17 +236,22 @@ where
...
+                    }
+                }
+
+                match next {
+                    Some(next) => messages_ids.push(*next),
+                    None => {
+                        on_message(storage, message_id, &message, metadata).await?;
+                        visited.insert(*message_id);
+                        messages_ids.pop();
+                    }
```
Alex6323, line 254: this is more of a comment. The code will work **as long** as no one messes with the message builder, or allows to bypass it. Out of interest: Is it possible that a pen tester could create a specially crafted messages that gets deserialized into a Message with an empty parents vec? Because for such messages the `None` case is executed also.

bee-message/src/message.rs
```diff
@@ -17,8 +17,7 @@ pub const MESSAGE_LENGTH_MAX: usize = 32768;
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Message {
     network_id: u64,
-    parent1: MessageId,
-    parent2: MessageId,
+    parents: Vec<MessageId>,
     payload: Option<Payload>,
     nonce: u64,
 }
```
Alex6323, line 26: also more of a comment to think about, then a merge blocker:

A `Message` has to have `1..` parents, not `0..`. Yes, we have guards in place that might always be enough to guarantee, that the `Vec` is nevaaa evaaa empty, but I'ld like to propose this type, that tells the compiler that we truely mean the former than the latter:

```rust
struct NonEmptyParents {
    first: MessageId,
    other: Vec<MessageId>,
} 
```
As an additional benefit: for Messages with only 1 parent there is no extra allocation. And then obviously we'ld implement Iterator on it and some API similar to Vec. I think it is easy to implement and will make the node more robust agains programmer errors. In contrast to that, children are perfectly fine impemented with just a `Vec`, as for a message there indeed can be `0..` of them.

bee-ledger/src/white_flag.rs
thibault-martinez: It's not possible no. You have to announce the number of parents in a message, if you put 0, it is directly rejected.

bee-protocol/src/worker/propagator.rs
```diff
@@ -108,7 +103,7 @@ async fn propagate<B: StorageBackend>(
             }
 
             // Send child to the tip pool.
-            if let Err(e) = tip_tx.send((*message_id, parent1, parent2, ms_index_maybe)).await {
+            if let Err(e) = tip_tx.send((*message_id, parents, ms_index_maybe)).await {
```
Alex6323: tbh, I think we should change this somewhat, because during syncing we don't want to send each just solidified messge to the tip-pool. I think we also want to do that only if we are within BELOW_MAX_DEPTH as in the `index_updater`. This could improve syncing performance.

thibault-martinez: I agree but this task also contains the solidification trigger atm so we would have to extract it. Do you want to do that ?

Alex6323: We could just rename things a bit (`tip_tx`/`tip_rx` -> `solidified_tx`/`solidified_rx` etc) and limit the `tangle.insert_tip` call to happen only when within BELOW_MAX_DEPTH. Wdyt?

Alex6323: I can do the change if you want and push to this PR.

thibault-martinez: Sounds good to me. No side effects by not sending the tips, right ?

samuel-rufi: There aren't any side effects when we don't add tips on synching. Testing currently the PR.

bee-message/src/message.rs
Alex6323: A simpler alternative is to make `NonEmptyParents` a simple wrapper of `Vec`, that ensures it is always initialized with at least one parent. This may be simpler, has the same guarantee, but will always allocate.

bee-tangle/src/urts.rs
```diff
@@ -65,23 +65,23 @@ impl UrtsTipPool {
...
-            self.add_child(*parent1, *hash);
-            self.add_child(*parent2, *hash);
+    fn link_parents_with_child(&mut self, child: &MessageId, parents: &Vec<MessageId>) {
+        let mut linked = Vec::new();
+        for parent in parents {
+            if !linked.contains(parent) {
+                self.add_child(*parent, *child);
+                linked.push(*parent);
+            }
         }
```
Alex6323, line 85: What do you think of doing this instead?
```rust
parents.iter().fold(HashSet::with_capacity(8), |mut linked, parent| {
    if linked.insert(parent) {
      self.add_child(*parent, *child);
    }
    linked
});
```

bee-tangle/src/urts.rs
```diff
@@ -103,12 +103,13 @@ impl UrtsTipPool {
...
-            self.check_retention_rules_for_parent(parent1);
-            self.check_retention_rules_for_parent(parent2);
+    fn check_retention_rules_for_parents(&mut self, parents: &Vec<MessageId>) {
+        let mut checked = Vec::new();
+        for parent in parents {
+            if !checked.contains(parent) {
+                self.check_retention_rules_for_parent(parent);
+                checked.push(*parent);
+            }
         }
```
Alex6323, line 113: same suggestion as above

bee-tangle/src/urts.rs
```diff
@@ -175,23 +176,29 @@ impl UrtsTipPool {
...
+            Some(if non_lazy_tips.len() < self.optimal_num_tips() {
+                non_lazy_tips.iter().map(|t| *t).collect()
+            } else {
+                non_lazy_tips
+                    .iter()
+                    .choose_multiple(&mut rand::thread_rng(), self.optimal_num_tips())
+                    .iter()
+                    .map(|t| **t)
+                    .collect()
+            })
```
Alex6323, line 193: I think this would be easier to read:
```rust
         match self.non_lazy_tips.len() {
            0 => None,
            n if n <= self.optimal_num_tips() => { // no point in using rand if equal
                Some(self.non_lazy_tips.iter().copied().collect())
            }, 
            _ => {
                Some(self.non_lazy_tips
                    .iter()
                    .choose_multiple(&mut rand::thread_rng(), self.optimal_num_tips())
                    .iter()
                    .copied()
                    .copied()
                    .collect()
                    )
            }
        }
```

PR #418, opened 2021-02-19
Use compile time checked parents type

Alex6323:
# Description of change

Based off of Thibault's 'use-parents-type' PR I created an alternative solution:

Advantages:
* Uses only a `Vec` internally instead of a single field and a vec, which makes sorting and deduping message ids easier, which this PR already comes with.
* The instantiated `Parents` type guarantees the correct range of parents (1..=8). This is achieved through a stage-builder pattern.
* The `Parents` type `deref`s to a `Vec`, and impls `From<Vec>`

Disadvantages:
* The stage-builder pattern is relatively verbose, and especially in our case we have 8 stages (1 for each parent). I tried to mitigate this problem a bit by making use of macros.
* The code is a bit weird, because you cannot use loops.

### DIFF
--- a/bee-message/src/parents.rs
+++ b/bee-message/src/parents.rs
@@ -5,7 +5,10 @@ use crate::{Error, MessageId, MESSAGE_ID_LENGTH};
 
 use bee_common::packable::{Packable, Read, Write};
 
-use std::{marker::PhantomData, ops::Deref};
+use std::{
+    marker::PhantomData,
+    ops::{Deref, RangeInclusive},
+};
 
 macro_rules! make_stages {
     ($($parent_n:ident),+) => {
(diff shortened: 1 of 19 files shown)


bee-message/src/parents.rs
```diff
@@ -51,19 +144,106 @@ impl Packable for Parents {
     }
 
     fn unpack<R: Read + ?Sized>(reader: &mut R) -> Result<Self, Self::Error> {
-        let parents_len = u8::unpack(reader)? as usize;
+        let _parents_len = u8::unpack(reader)? as usize;
+
+        // const MESSAGE_PARENTS_RANGE: RangeInclusive<usize> = 1..=8;
+
+        // if !MESSAGE_PARENTS_RANGE.contains(&parents_len) {
```
thibault-martinez, line 151: The check still needs to happen as it is part of the protocol. If there is no check, an invalid message could be accepted by bee but denied by hornet, leading to a consensus diff.

bee-message/src/parents.rs
```diff
@@ -51,19 +144,106 @@ impl Packable for Parents {
...
+        //     return Err(Error::InvalidParentsCount(parents_len));
+        // }
+
+        // NB: The code doesn't depend on `parents_len`, or its correctness anymore. Instead, the
+        // code errors out, if it can not unpack at least 1, or even more than 8 parents. Each
+        // `parents` variable is a distinct type according to the builder stage we're in.
```
thibault-martinez, line 157: You are relying on external factors (remember you can have many things packed one after the other, there could be lots of data in the reader after your message) where the deserialisation of a type should be self-contained.

bee-message/src/parents.rs
Alex6323: you're absolutely right. Preventing the check as a whole is impossible as long as it's serialized somewhere in `Vec` "manner". 
Tbh, I don't care so much for this PR to be merged. I see this PR more like an interesting case study to have a conversation about. I think the proposed `Parents` wrapper is something we should merge, though. We can use it to have the runtime checks in there.

bee-message/src/parents.rs
Alex6323: yes, you're right here as well. And that's something I wondered about actually during implementation. Good point!

Alex6323:
We chose to go with a simple wrapper type  (`struct Parents(Vec<MessageId>)`), that ensures the invariant at runtime. That should be good enough for all intents and purposes.

PR #434, opened 2021-02-24
Add password check

Dr-Electron:
# Description of change

I think a password check for the password tool is quite convenient, because it's easy to make an error.

## Type of change

`bee password` did run successfully with matching and non matching passwords.

## Change checklist

### DIFF
--- a/bee-node/src/tools/mod.rs
+++ b/bee-node/src/tools/mod.rs
@@ -43,7 +43,11 @@ pub fn exec(tool: &Tool) -> Result<(), ToolError> {
         Tool::P2pIdentity(tool) => p2p_identity::exec(tool),
         Tool::Rocksdb(tool) => rocksdb::exec(tool)?,
         Tool::SnapshotInfo(tool) => snapshot_info::exec(tool)?,
-        Tool::Password(tool) => password::exec(tool)?,
+        Tool::Password(tool) => {
+            if let Err(e) = password::exec(tool) {
+                println!("error: {}", e)
+            }
+        }
     }
 
     Ok(())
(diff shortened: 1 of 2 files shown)


bee-node/src/tools/mod.rs
```diff
@@ -43,7 +43,11 @@ pub fn exec(tool: &Tool) -> Result<(), ToolError> {
         Tool::P2pIdentity(tool) => p2p_identity::exec(tool),
         Tool::Rocksdb(tool) => rocksdb::exec(tool)?,
         Tool::SnapshotInfo(tool) => snapshot_info::exec(tool)?,
-        Tool::Password(tool) => password::exec(tool)?,
+        Tool::Password(tool) => {
+            if let Err(e) = password::exec(tool) {
```
thibault-martinez: Just use `?` like the other tools, the error will already be printed in the main.

Dr-Electron: But the main only tries to unwrap and would print a much longer error message, because it fails to unwrap. I though that doesn’t look good only because you used the wrong password the second time.

thibault-martinez: The unwrap is temporary, we'll log the actual error as soon as the logger issue (not being initialised before the node) is fixed.

Dr-Electron: A ok. I wasn’t sure why unwrap was used, so i printed the error here. Will change it 👍🏻

thibault-martinez (approved):

thibault-martinez:
Thanks

PR #464, opened 2021-03-15
Add const generics features to the Packable trait.

thibault-martinez:
# Description of change

Add const generics features to the Packable trait.

## Change checklist


bee-common/bee-common/src/packable.rs
```diff
@@ -26,9 +26,49 @@
     }
 
     /// Reads bytes from the passed reader and unpacks them into an instance.
-    fn unpack<R: Read + ?Sized>(reader: &mut R) -> Result<Self, Self::Error>
+    fn unpack_inner<R: Read + ?Sized, const TRUSTED: bool>(reader: &mut R) -> Result<Self, Self::Error>
```
Alex6323: I think instead of "TRUSTED" I'ld rather use "CHECKED", the reason being that the concept of "trust" is broad and vague while the term "checked" insinuating nothing except that something is ... well .. checked, whether the source being trustful or not.

Alex6323: While essentially you cannot trust any byte (as bit flips can happen anywhere anytime) it still can be reasonable to not check a byte for practicality reasons.

thibault-martinez: Yeah but `TRUSTED` is applied to the Reader, not the bytes. A reader can be trusted (storage) or untrusted (network). But tbh I am tempted to reduce it down to only `T` as it's going to be used a lot in implementations and would be unnecessarily repetitive.

thibault-martinez: Actually, I think `CHECK` is a better name than `TRUSTED` and `CHECKED`. Instead of qualifying the source or the bytes, it just says if we should apply the checks or not.

bee-common/bee-common/CHANGELOG.md
```diff
@@ -19,12 +19,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-<!-- ## 0.4.0 - 2021-XX-XX
+## 0.4.0 - 2021-25-03
```
thibault-martinez:
```suggestion
## 0.4.0 - 2021-26-03
```

zesterer (dismissed):

bee-common/bee-common/src/packable.rs
Alex6323: is fine with me

Alex6323 (dismissed):

Alex6323 (approved):

zesterer (approved):

Adam-Gleave (approved):

samuel-rufi (approved):

PR #481, opened 2021-03-22
Dockerfile to run Bee node in a container

karimodm:
# Description of change

I would have liked to go for a more minimal runtime docker image, I have
experiemented with gcr.io/distroless/base and alpine. They both pose a
challenge due the glibc, stdc++ and ssl library runtimes; I finally opted
to make the Dockerfile more readable and use a minimal Debian image instead.

## Type of change

Ran the resulting image.


thibault-martinez (commented):
Can we maybe put this in `bee-node` as it only makes sense for this crate ?

karimodm:
We could, but the docker build process forbids to include any parent folder as part of the build context. This will make the `docker build` less intuitive as it cannot be run from the inner folder, hence I've decided not to do so.

This being said: `docker build -t bee -f bee-node/Dockerfile .` is not too hard either ;)

Alex6323:
@karimodm I think we should build Bee's dependencies separately from Bee to better make use of Docker's caching abilities. Wdyt?

karimodm:
How would you do that @Alex6323? The problem I see there though is that a `Dockerfile` in my view should be as self-contained and independent as possible; as you want to build on a variety of different platforms and contexts, and always obtain the same as-deterministic-as-possible result. When you do `COPY . /src` and any of the files within changed, you are already invalidating all the following steps' docker cache.

Alex6323:
Hey @karimodm :smiley: I wrote about this in a little bit more detail in the `bee-dev` channel yesterday. Building the dependencies of Bee takes a lot of time, and they certainly change less frequently than the Bee codebase. Hence, in many cases the docker image build process could profit from a prior build stage, that ONLY compiles Bee's dependencies, not Bee itself. If there are no changes in Bee's dependencies, Docker would be able to cache and basically skip this step entirely, thereby speeding up the process significantly. The authors of this crate `https://lib.rs/crates/cargo-chef` report speed-ups of around 4-5x when using their 4 staged process.

karimodm:
I see what you are talking about @Alex6323, good stuff! I've changed the Dockerfile to run the 4-step process you mentioned; see if it's good for you.

Alex6323:
Really nice @karimodm !! That's what I meant. Do we not also need to expose the gossip, dashboard, mqtt ports?

karimodm:
Totally forgot!

bee-node/Dockerfile
```diff
@@ -0,0 +1,58 @@
+############################
+# Prepare dependencies recipe
+############################
+FROM lukemathwalker/cargo-chef as planner
+
+WORKDIR /src
+COPY . .
```
Alex6323, line 7: Are you sure this is going to work? The `WORKDIR` points to a directory that doesn't exist at the target, because you copy to `.`. I suggest doing:
```
COPY . /bee
WORKDIR /bee
```
or just:
```
COPY . .
```

bee-node/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+############################
+# Dependency cache
+############################
+FROM lukemathwalker/cargo-chef as cacher
+
+RUN apt-get update && \
+    apt-get install cmake clang -y
+
+WORKDIR /src
+COPY --from=planner /src/recipe.json recipe.json
```
Alex6323: Instead of `/src` I'ld use `/bee`, otherwise it might be confused with the actual `src` folders in the sub-crates.

bee-node/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+RUN cargo chef cook --release --recipe-path recipe.json
+
+############################
+# Build
+############################
+FROM rust:1 as build
+
+WORKDIR /src
+COPY . .
```
Alex6323: Same as above. Are we sure that `.` at the source wouldn't falsely point to `./src`, which does not exist?

bee-node/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+COPY --from=cacher /src/target target
```
Alex6323:
```
COPY . /bee
WORKDIR /bee

COPY --from=cacher /bee/target target
```

bee-node/Dockerfile
Alex6323: Could you also please add a `.dockerignore` file, which filters out any build artifacts residing in `/target`?

bee-node/Dockerfile
karimodm: Yes, this works; Docker takes care of creating the WORKDIR.

bee-node/Dockerfile
karimodm: Alright, changing the workdir to `/bee`

Alex6323 (approved):
lgtm :+1:

thibault-martinez (approved):

Issue #166, opened 2020-12-31
Make IP and port keys in config more uniform


Dr-Electron:
## Description
In the config, binding IP and port are sometimes configured as two key, sometimes as a single string.
Making the config more uniform would make the syntax easier to remember.

I would suggest the separate key option. Because that way you could use it as one liner.
Example:
```
mqtt = { address = "127.0.0.1", port = 1883, protocol = "tcp" }
```
which is the same as
```
[mqtt] 
address = "127.0.0.1"
port = 1883
protocol = "tcp"
```

## Are you planning to do it yourself in a pull request?
Yes.

Dr-Electron:
Fixed with #483

PR #483, opened 2021-03-22
Fix/uniform config

Dr-Electron:
# Description of change

I tried to make the config more uniform. I used MultiAddr as most people seemed to prefer it over my proposed idea in the issue #166.Furthermore, I used `libp2p` as library. Not sure if I should just use `parity_multiaddr `.
I could also add MQTT with the `tcp/1883` Multiaddr. But if we only need a port for MQTT, that is probably to verbose.
If this gets merged I also need to change the docs, if they also get merged.

fixes issue #166

## Type of change

Tried to run the node with different configs and everything seemed to work.

## Change checklist


bee-node/src/plugins/dashboard/config.rs
```diff
@@ -1,13 +1,15 @@
...
+use libp2p::{multiaddr::Protocol, Multiaddr};
 use serde::Deserialize;
+use std::net::{IpAddr, SocketAddr, ToSocketAddrs};
 
 const DEFAULT_SESSION_TIMEOUT: u64 = 86400;
 const DEFAULT_USER: &str = "admin";
 const DEFAULT_PASSWORD_SALT: &str = "0000000000000000000000000000000000000000000000000000000000000000";
 const DEFAULT_PASSWORD_HASH: &str = "0000000000000000000000000000000000000000000000000000000000000000";
-const DEFAULT_PORT: u16 = 8081;
+const DEFAULT_BIND_ADDRESSES: &str = "/ip4/1.0.0.0/tcp/8081";
```
Alex6323: the `1` should probably be a `0`?

Dr-Electron: Ups. Yes it should be

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +76,44 @@ impl DashboardConfigBuilder {
...
+            .unwrap_or_else(|| DEFAULT_BIND_ADDRESSES.parse().unwrap());
+        let address = multi_addr
+            .iter()
+            .find_map(|x| match x {
+                Protocol::Dns(std::borrow::Cow::Borrowed("localhost")) => Some(
+                    ("localhost", 0)
+                        .to_socket_addrs()
+                        .expect("error resolving 'localhost'")
+                        .nth(0)
+                        .unwrap()
```
Alex6323, line 94: pls add a comment why `unwrap`ing is okay! I suggest:
"Unwrapping here is fine, because `to_socket-addrs()` didn't return an error, thus we can be sure that the iterator contains at least 1 element."

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +76,44 @@ impl DashboardConfigBuilder {
     }
 
     pub fn finish(self) -> DashboardConfig {
+        let multi_addr = self
+            .bind_addresses
+            .unwrap_or_else(|| DEFAULT_BIND_ADDRESSES.parse().unwrap());
```
Alex6323: add a small comment why `unwrap`ing is safe. I suggest:
"We made sure that the default value is valid and therefore parseable.". Alternatively you could use `expect("error parsing default bind addresses")`.

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +76,44 @@ impl DashboardConfigBuilder {
...
+        let address = multi_addr
+            .iter()
+            .find_map(|x| match x {
+                Protocol::Dns(std::borrow::Cow::Borrowed("localhost")) => Some(
```
Alex6323: Should we really limit ourselves to "localhost"? Why would we not allow any domain name for the sake of supporting anybody's (maybe sometimes unusual) configuration?

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +76,44 @@ impl DashboardConfigBuilder {
...
+                _ => None,
+            })
+            .unwrap_or_else(|| panic!("Unsupported address"));
+
+        let port = multi_addr
+            .iter()
+            .find_map(|x| match x {
+                Protocol::Tcp(port) => Some(port),
+                _ => None,
+            })
```
Alex6323: I suggest using `if let` instead of `match` here:
`.find_map(|x| if let Protocol::Tcp(port) { Some(port) } else { None })`

bee-node/src/plugins/dashboard/mod.rs
```diff
@@ -161,11 +160,10 @@ where
                 users.clone(),
             );
 
-            info!("Dashboard available at http://localhost:{}.", config.port());
+            info!("Dashboard available at http://{}.", config.bind_socket_addr());
```
Alex6323: I guess I would print that log only *after* the `warp` server has been spawned, because only then we know what IP the server was actually bound to. If you set `0.0.0.0` in your config, then this log output wouldn't be helpful as it would just replicate that.

bee-node/src/plugins/dashboard/routes.rs
```diff
@@ -122,7 +122,7 @@ pub(crate) fn api_routes(
     allowed_routes
         .and(reverse_proxy_filter(
             "".to_string(),
-            "http://localhost:".to_owned() + &rest_api_config.binding_socket_addr().port().to_string() + "/",
+            "http://localhost:".to_owned() + &rest_api_config.bind_socket_addr().port().to_string() + "/",
```
Alex6323: remove `localhost:` :smile:

bee-api/bee-rest-api/Cargo.toml
```diff
@@ -29,6 +29,7 @@ hex = { version = "0.4" }
 iota-crypto = { git = "https://github.com/iotaledger/crypto.rs", rev = "c3bf565eba62d0b81144174c2ff917bfde282e49", features = ["blake2b"], optional = true }
 log = { version = "0.4", optional = true }
 num_cpus = { version = "1.13", optional = true }
+libp2p = { version = "0.35", default-features = false }
```
Alex6323: Update `libp2p` version to "0.36".

bee-node/src/plugins/dashboard/config.rs
Dr-Electron: I guess the same would apply to port initialization in `bee-api/bee-rest-api/src/endpoints/config.rs `?

bee-api/bee-rest-api/Cargo.toml
Dr-Electron: I guess the same would apply to `bee-node/Cargo.toml`?

bee-node/src/plugins/dashboard/config.rs
Dr-Electron: Als here again: I probably should add the same comments to the rest-api config.rs?

bee-node/src/plugins/dashboard/config.rs
Dr-Electron: You are right. I have never seen anyone use anything else, but then I should not infer from me to others 👍

bee-node/src/plugins/dashboard/config.rs
Alex6323: probably yea :+1:

bee-node/src/plugins/dashboard/config.rs
Alex6323: yea, I guess. It's not a real issue, of course. If there were a good reason this `match` would maybe be expanded later at some point, I'ld say keep it. But I feel like this is not going to happen in this case.

bee-api/bee-rest-api/Cargo.toml
Alex6323: yes :+1:

Alex6323 (approved):
lgtm :zap: :+1:

bee-node/config.example.toml
```diff
@@ -47,8 +47,7 @@ status_interval = 10
 ms_sync_count = 200
 
 [rest_api]
-binding_port          = 14265
-binding_ip_addr       = "0.0.0.0"
+bind_addresses        = "/ip4/0.0.0.0/tcp/14265"
```
samuel-rufi: why is it called `bind_addresses` and not `bind_address`? we only have a single binding address?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -3,10 +3,10 @@
 
 use serde::Deserialize;
 
-use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, ToSocketAddrs};
+use libp2p::{multiaddr::Protocol, Multiaddr};
```
samuel-rufi: just a little thing, but please order the imports in following way:

1) bee-rest-api imports
2) workspace dependencies
3) external dependencies
4) std dependencies

samuel-rufi (changes requested):

bee-node/config.example.toml
thibault-martinez: You're not the only one confused 🙃

bee-api/bee-rest-api/src/endpoints/config.rs
thibault-martinez: With a blank line between sections!

bee-node/config.example.toml
Dr-Electron: I don't know ;) . I just applied the logic from @Alex6323. I think he made the same change somewhere else and I tried to be conform. The idea was, that with 0.0.0.0 you bind on all addresses. At least that's what I remember

Alex6323: I understand the confusion, and I am not against changing it back to `bind_address` (singular) also for the network. It's worth mentioning though, that `0.0.0.0` in context of servers indeed means binding to potentially multiple addresses. For reference, read `https://en.wikipedia.org/wiki/0.0.0.0`.

Alex6323: @Dr-Electron Can you also change it back to singular for the network then?

thibault-martinez: I understand the reasoning, but even in your link they refer to this as a singular address.

Alex6323: Sure. In principle you're right. It's an address, or meta-address, but it is not a valid bind-address in itself. In context of bind-addresses it usually describes a set of valid actual addresses. The reason I changed it in `bee-network` to plural was that if we follow Hornet's configuration, we should allow Bee to bind to multiple addresses, hence the plural. Currently we don't support that in a way that let's you pass an array of valid bind addresses. Instead we can bind to multiple addresses via the `0.0.0.0` meta-address just like Hornet does atm as well. This is the section of Hornet's config:
```json
"p2p": {
    "bindMultiAddresses": [
      "/ip4/0.0.0.0/tcp/16600"
    ],
```
If that makes also sense for other parts of the node in the same way as for the network can be up for debate.

thibault-martinez: Are we planning on actually supporting multiple addresses ? If yes, I'd already make it an array, even if not fully used.

Alex6323: I think it makes sense to support that, yes. Not supporting it would probably even mean that we shouldn't allow `0.0.0.0`.

thibault-martinez: Then let's do plural and array 👍

Alex6323: okay, will add that to my todo list. @Dr-Electron I hope you didn't spend your time on this. We keep the plural in the network, and I will open a PR supporting multiple single multiaddresses. (lol)

Dr-Electron: Ok, so I leave it as it is?

Alex6323: yes, at least for the network leave it as it is.

Dr-Electron: And for the API and dashboard? Because if we apply the same logic, it would also make sense there. I could theoretically have multiple network adapters and bind my dashboard to two of them, but not to all. I guess i will mark this as resolve than?

thibault-martinez (changes requested):
Please fix the conflict

bee-api/bee-rest-api/Cargo.toml
```diff
@@ -29,6 +29,7 @@ hex = { version = "0.4" }
 iota-crypto = { git = "https://github.com/iotaledger/crypto.rs", rev = "c3bf565eba62d0b81144174c2ff917bfde282e49", features = ["blake2b"], optional = true }
 log = { version = "0.4", optional = true }
 num_cpus = { version = "1.13", optional = true }
+libp2p = { version = "0.36", default-features = false }
```
thibault-martinez: Please sort dependencies alphabetically

bee-node/config.example.toml
```diff
@@ -99,7 +98,7 @@ delay   = 60480
 address = "tcp://localhost:1883"
 
 [dashboard]
-port  = 8081
+bind_addresses  = "/ip4/0.0.0.0/tcp/8081"
```
thibault-martinez: As per the other discussion, this should either be singular or an array. I don't think there is any usecase where we would need to bind many addresses for the dashboard so I would go the singular way.

bee-node/config.example.toml
thibault-martinez: Same

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -119,18 +112,41 @@ impl RestApiConfigBuilder {
...
+                        //Unwrapping here is fine, because to_socket-addrs() didn't return an error,
+                        //thus we can be sure that the iterator contains at least 1 element.
+                        .unwrap()
+                        .ip(),
+                ),
+                Protocol::Ip4(ip) => Some(IpAddr::V4(ip)),
+                Protocol::Ip6(ip) => Some(IpAddr::V6(ip)),
+                _ => None,
+            })
+            .unwrap_or_else(|| panic!("Unsupported address"));
```
thibault-martinez: Prefer `expect` instead ?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -119,18 +112,41 @@ impl RestApiConfigBuilder {
...
+
+        let port = multi_addr
+            .iter()
+            .find_map(|x| if let Protocol::Tcp(port) = x { Some(port) } else { None })
+            .unwrap_or_else(|| panic!("Unsupported protocol"));
```
thibault-martinez, line 149: expect ?

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +77,44 @@ impl DashboardConfigBuilder {
...
+                        //Unwrapping here is fine, because to_socket-addrs() didn't return an error,
+                        //thus we can be sure that the iterator contains at least 1 element.
+                        .unwrap()
+                        .ip(),
+                ),
+                Protocol::Ip4(ip) => Some(IpAddr::V4(ip)),
+                Protocol::Ip6(ip) => Some(IpAddr::V6(ip)),
+                _ => None,
+            })
+            .unwrap_or_else(|| panic!("Unsupported address"));
```
thibault-martinez: expect ?

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +77,44 @@ impl DashboardConfigBuilder {
...
+
+        let port = multi_addr
+            .iter()
+            .find_map(|x| if let Protocol::Tcp(port) = x { Some(port) } else { None })
+            .unwrap_or_else(|| panic!("Unsupported protocol"));
```
thibault-martinez: expect ?

thibault-martinez (changes requested):

bee-node/config.example.toml
samuel-rufi: Agree fully, singular or array seems nice.

bee-node/Cargo.toml
```diff
@@ -33,6 +33,8 @@ futures = "0.3"
 fxhash = "0.2"
 hex = "0.4"
 iota-crypto = { version = "0.5.0", features = [ "ed25519", "random", "blake2b" ] }
+jsonwebtoken = { version = "7.2", optional = true }
```
thibault-martinez: Why are you introducing this dependency ?

bee-api/bee-rest-api/Cargo.toml
```diff
@@ -27,6 +27,7 @@ digest = { version = "0.9", optional = true }
 futures = { version = "0.3", optional = true }
 hex = { version = "0.4" }
 iota-crypto = { version = "0.5.0", features = ["blake2b"], optional = true }
+libp2p = { version = "0.36", default-features = false }
```
thibault-martinez: Use https://crates.io/crates/multiaddr if you just need MultiAddr

bee-node/Cargo.toml
```diff
@@ -33,6 +33,8 @@ futures = "0.3"
...
+libp2p = { version = "0.36", default-features = false }
```
thibault-martinez: Use https://crates.io/crates/multiaddr if you just need MultiAddr

thibault-martinez (changes requested):

bee-node/Cargo.toml
Dr-Electron: Tbh, I have no clue. Probably messed up the rebase 😆. Will fix it

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +77,44 @@ impl DashboardConfigBuilder {
     }
 
     pub fn finish(self) -> DashboardConfig {
+        let multi_addr = self
+            .bind_address
+            //We made sure that the default value is valid and therefore parseable.
```
thibault-martinez:
```suggestion
            // We made sure that the default value is valid and therefore parseable.
```

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +77,44 @@ impl DashboardConfigBuilder {
...
+            .unwrap_or_else(|| DEFAULT_BIND_ADDRESS.parse().unwrap());
+        let address = multi_addr
+            .iter()
+            .find_map(|x| match x {
+                Protocol::Dns(address) => Some(
+                    (address.to_string(), 0)
+                        .to_socket_addrs()
+                        .unwrap_or_else(|error| panic!("error resolving '{}':{}", address, error))
+                        .nth(0)
+                        //Unwrapping here is fine, because to_socket-addrs() didn't return an error,
```
thibault-martinez:
```suggestion
                        // Unwrapping here is fine, because to_socket-addrs() didn't return an error,
```

bee-node/src/plugins/dashboard/config.rs
```diff
@@ -74,16 +77,44 @@ impl DashboardConfigBuilder {
...
+                        //thus we can be sure that the iterator contains at least 1 element.
```
thibault-martinez:
```suggestion
                        // thus we can be sure that the iterator contains at least 1 element.
```

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -127,10 +120,33 @@ impl RestApiConfigBuilder {
 
     /// Builds the REST API config.
     pub fn finish(self) -> RestApiConfig {
-        let binding_socket_addr = match self.binding_ip_addr.unwrap_or(DEFAULT_BINDING_IP_ADDR) {
-            IpAddr::V4(ip) => SocketAddr::new(IpAddr::V4(ip), self.binding_port.unwrap_or(DEFAULT_BINDING_PORT)),
-            IpAddr::V6(ip) => SocketAddr::new(IpAddr::V6(ip), self.binding_port.unwrap_or(DEFAULT_BINDING_PORT)),
-        };
+        let multi_addr = self
+            .bind_address
+            //We made sure that the default value is valid and therefore parseable.
```
thibault-martinez:
```suggestion
            // We made sure that the default value is valid and therefore parseable.
```

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -127,10 +120,33 @@ impl RestApiConfigBuilder {
...
+            .unwrap_or_else(|| DEFAULT_BIND_ADDRESS.parse().unwrap());
+        let address = multi_addr
+            .iter()
+            .find_map(|x| match x {
+                Protocol::Dns(address) => Some(
+                    (address.to_string(), 0)
+                        .to_socket_addrs()
+                        .unwrap_or_else(|error| panic!("error resolving '{}':{}", address, error))
+                        .next()
+                        //Unwrapping here is fine, because to_socket-addrs() didn't return an error,
```
thibault-martinez:
```suggestion
                        // Unwrapping here is fine, because to_socket-addrs() didn't return an error,
```

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -127,10 +120,33 @@ impl RestApiConfigBuilder {
...
+                        //thus we can be sure that the iterator contains at least 1 element.
```
thibault-martinez:
```suggestion
                        // thus we can be sure that the iterator contains at least 1 element.
```

thibault-martinez (approved):

PR #488, opened 2021-03-26
Fmt version 2

thibault-martinez:


zesterer (approved):

Alex6323:
Can you motivate this PR in the description? Would make it easier to review it? Is rustfmt version "two" a new thing I missed, or is it around for longer already? I see a lot of changes in the files, so I assume version "one" just didn't format those instances for backwards compatibility reasons with rustfmt 1.0. Is there anything else interesting to know to about this?

thibault-martinez:
Some formatting rules can only make it to version 2 because of breaking changes in implementation. Version two is then version one + all the rules that can't make it to version one, it's more complete. It's not new (2 years old) but we discovered it while stumbling upon this https://github.com/rust-lang/rustfmt/issues/4355#issuecomment-723764276.

Alex6323 (approved):

bingyanglin (approved):
Tbh I like this new fmt version (cleaner information in one line than the previous version).

PR #489, opened 2021-03-26
Add FPC Components

Adam-Gleave:
Adds FPC components using goshimmer implementation as a reference.


bee-vote/src/context.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+
+    /// Add the given `Opinion` to the `VoteContext`.
+    pub(crate) fn add_opinion(&mut self, opinion: Opinion) {
+        self.opinions.push(opinion);
+    }
+
+    /// Retrieve the last formed `Opinion`.
+    pub fn last_opinion(&self) -> Option<Opinion> {
```
m-renaud, line 131: Out of curiosity, is there any situation where it's valid not to have any opinions at all? The `new()` function above enforces that an initial opinion is provided (guaranteed initial opinion), and the `VoteContextBuilder` enforces that the caller uses either `with_initial_opinion` (guaranteed initial opinion)  or `with_initial_opinions` (possible to construct an `Opinions` with no elements).

If it is indeed the case that there **must** be an initial opinion, the VoteContextBuilder could be updated to check the length of the opinions vector and error otherwise (also with `NoInitialOpinions`). This way the API here could be simplified to just return an `Opinion`.

The downside of this is that if there somehow became a way to construct a `VoteContext` that bypassed these checks, the only option here would be to either: a) return a default opinion (is there a "safe" one, protocol wise?) or b) panic (also not great). It seems like the risk here is pretty low though since there are only two possible ways to construct a `VoteContext`.

WDYT?

bee-vote/src/context.rs
```diff
@@ -0,0 +1,252 @@
...
+        self.opinions.last().copied()
+    }
+
+    /// Describes whether this `VoteContext` has been finalized.
```
m-renaud, line 135: Can you also add a comment on how it is determined when a vote is finalized (same opinion for n rounds) and the implications (will not change it's opinion in the future).

bee-vote/src/context.rs
```diff
@@ -0,0 +1,252 @@
...
+                true
+            } else {
+                false
+            }
+        }
+    }
+
+    /// Describes whether the `VoteContext` is new (has not participated in a vote).
+    pub(crate) fn is_new(&self) -> bool {
+        self.liked == LIKED_INITIAL
```
m-renaud: I always get a bad feeling whenever I see floating point equality comparisons 😝  In theory this _should_ be fine because `self.liked` is initialized above to `LIKED_INITIAL` but `self.liked < -0.5_f64` just feels "safer" haha, feel free to ignore this.

bee-vote/src/context.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /// Described whether the `VoteContext` has *just* finished its first round.
+    pub(crate) fn had_first_round(&self) -> bool {
+        self.rounds == 1
+    }
+
+    pub(crate) fn had_fixed_round(
```
m-renaud, line 181: Could you add a comment for what a "fixed round" means?

bee-vote/src/opinion.rs
```diff
@@ -0,0 +1,127 @@
...
+
+impl ops::DerefMut for Opinions {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl Opinions {
+    /// Create a new `Opinions` wrapper from `Vec<Opinion>`.
+    pub fn new(inner: Vec<Opinion>) -> Self {
```
m-renaud, line 109: Related to my other comment about the validity of "no opinions", if it is indeed the case that there should always be at least one then `new()` could check that checks the length (>0). This would mean any errors creating opinions would be discovered at the creation site and not when constructing the `VoteContext` (or asking for the last opinion).

The one place I see where Opinions could be empty is here: https://github.com/iotaledger/bee/pull/489/files?file-filters%5B%5D=.rs#diff-54c04e784e9d4860df9ff084e3b79695ff08fd90ff444706039236e13bf8e1a1R77

But it's also possible that this could just return Vec<Opinion> instead of an Opinions type.

bee-vote/src/statement/opinion.rs
```diff
@@ -0,0 +1,119 @@
...
+    collections::BinaryHeap,
+    ops::{Deref, DerefMut},
+};
+
+/// Length (in bytes) of this statement when serialized.
+pub const OPINION_STATEMENT_LENGTH: usize = 2;
+
+/// Opinion registry statement.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub struct Opinion {
```
m-renaud: I found it a little bit confusing that there were multiple public structs named `Opinion` and `Opinions` in two subdirectories of the same crate, maybe consider renaming them to have a prefix so they're less ambiguous?

bee-vote/src/context.rs
Adam-Gleave: I agree, good catch!

bee-vote/src/context.rs
Adam-Gleave: It is valid for a voter to form no `Opinion`s on an object. These will get skipped in the "liked" calculation at [this line](https://github.com/iotaledger/bee/blob/f1726f92e705a8393be6715b79387aeeef4374ce/bee-vote/src/fpc.rs#L467). What I have realised because of this comment, though, is that the `min_opinions_received` field should be restricted to `> 0`, to stop any nasty panics if a user mistakenly sets it to 0 (this would skip the continue statement and cause a division by zero at [this line](https://github.com/iotaledger/bee/blob/f1726f92e705a8393be6715b79387aeeef4374ce/bee-vote/src/fpc.rs#L474)).

I'll address that!

bee-vote/src/statement/opinion.rs
Adam-Gleave: Yes, that seems fair! I will rename to `OpinionStatement` and `OpinionStatements`. I have already made these aliases in the code, but I agree, it makes sense to make that change everywhere.

bee-vote/src/context.rs
Adam-Gleave: Yep, I'll add comments for each of the areas highlighted in your suggestions 👍

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+            if let Some(selected_count) = queries.get_mut(index) {
+                *selected_count += 1;
+            }
+        }
+
+        let vote_map = Arc::new(RwLock::new(HashMap::<VoteObject, Opinions>::new()));
+        let all_queried_opinions = Arc::new(RwLock::new(Vec::<QueriedOpinions>::new()));
+        let mut query_futures = vec![];
+
+        for (i, opinion_giver) in opinion_givers.iter_mut().enumerate() {
```
m-renaud: Could you use `opinion_givers.iter_mut().zip(queries.iter())`, then you don't need to do the (potentially unsafe "This should never panic, trust me!" haha) `queries.get(i).unwrap()` below, and makes it clear you are only processing corresponding elements in the two vectors (`(opinion_giver, selected_count)`).

Another option would be to have a function that returns `Vec<(OpinionGiver, times_to_count_opinion)>` (or make a new struct, but then you have to come up with a good name :P), then the code here doesn't need to worry at all about the case of the vectors somehow (impossibly?) being a different length; just eliminate any parallel vector skew at construction time.

Yeah, I think I like the second approach more.

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+                        all_queried_opinions.clone(),
+                        opinion_giver,
+                        *selected_count,
+                    ),
+                ));
+            }
+        }
+
+        // Perform all queries.
+        futures::future::join_all(query_futures).await;
```
m-renaud: This whole blocks seems pretty self encapsulated, with the output being the `vote_map` and `all_queried_opinion`, would it be possible to pull this into it's own function?

Also, I think this could mean you could `Arc::try_unwrap(all_queried_opinions).unwrap().into_inner()` inside that function and just return the `Vec`, hiding the `Arc` and `RwLock`s from the caller.

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+
+        // Select queries.
+        let mut queries = vec![0u32; opinion_givers.len()];
+        for _ in 0..self.query_sample_size {
+            let index = rng.sample(dist);
+
+            if let Some(selected_count) = queries.get_mut(index) {
+                *selected_count += 1;
+            }
+        }
```
m-renaud: Could you add a comment describing what this loop does (and maybe pull into its own function)? It took me a few times reading it (and `do_query()` and `QueriedOpinions`) to understand that `selected_count` is a randomized distribution of how many times to count each opinion. Possibly rename to `times_to_count_opinion` too throughout?

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+    async fn do_query(
+        query_ids: &QueryObjects,
+        vote_map: Arc<RwLock<HashMap<VoteObject, Opinions>>>,
+        all_queried_opinions: Arc<RwLock<Vec<QueriedOpinions>>>,
+        opinion_giver: &mut Box<dyn OpinionGiver>,
+        selected_count: u32,
+    ) {
+        // Get opinions from the `OpinionGiver`.
+        let opinions = if let Ok(opinions) = opinion_giver.query(query_ids) {
+            // Sanity check.
```
m-renaud: "Sanity check" doesn't convey much info (and there's a separate point about this term being non-inclusive - https://gist.github.com/seanmhanson/fe370c2d8bd2b3228680e38899baf5cc), could you change it to explain what invariant is being checked? My understanding is:

// Verify that the number of opinions given matches the number of conflict and timestamp queries.

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+    /// Select a number of `OpinionGiver`s and query them for opinions.
+    async fn query_opinions(&self) -> Result<Vec<QueriedOpinions>, Error> {
+        let mut rng = thread_rng();
+
+        let query_ids = self.vote_context_ids().await;
+        if query_ids.conflict_objects.is_empty() && query_ids.timestamp_objects.is_empty() {
+            return Ok(vec![]);
+        }
+
+        // Create opinion givers.
```
m-renaud: ... and validate that there is at least one.

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+        let mut vote_map_guard = vote_map.write().await;
+
+        // Get opinions on a voting object and add to vote map.
+        let mut query_voting_objects = |objects: &Vec<VoteObject>| {
+            for (i, object) in objects.iter().enumerate() {
+                let mut votes = vote_map_guard
+                    .get(object)
+                    .map_or(Opinions::new(vec![]), |opt| opt.clone());
+
+                for _ in 0..selected_count {
```
m-renaud: `selected_count` here is also not immediately clear, after reading this again I think `times_to_count_opinion` is a clearer name.

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+            }
+        }
+
+        QueryObjects {
+            conflict_objects,
+            timestamp_objects,
+        }
+    }
+
+    fn round_thresholds(&self, ctx: &VoteContext) -> (f64, f64) {
```
m-renaud: wdyt about renaming to `round_opinion_change_thresholds()`?

Could you comment this? To make sure I understand:

/// Return the uniform range used to select the threshold at which a node will change it's opinion based on votes.

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+        all_queried_opinions.write().await.push(queried_opinions);
+    }
+
+    /// Get the IDs of all `VoteContext`s currently in the voting pool.
+    async fn vote_context_ids(&self) -> QueryObjects {
+        let context_guard = self.contexts.read().await;
+        let mut conflict_objects = vec![];
+        let mut timestamp_objects = vec![];
+
+        for (_, context) in context_guard.iter() {
```
m-renaud: I think you can use `context_guard.values()` here since the key is ignored.

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+                    .get(object)
+                    .map_or(Opinions::new(vec![]), |opt| opt.clone());
+
+                for _ in 0..selected_count {
+                    votes.push(opinions[i]);
+                }
+
+                queried_opinions.opinions.insert(*object, opinions[i]);
+
+                if vote_map_guard.contains_key(object) {
```
m-renaud: To remove the `unwrap()` I think the following would work:

```
if let Some(opinions) = vote_map_guard.get_mut(object) {
  *opinions = votes;
} else {
  vote_map_guard.insert(*object, votes);
}
```

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+            opinion_giver_id: opinion_giver.id().to_string(),
+            opinions: HashMap::new(),
+            times_counted: selected_count,
+        };
+
+        let mut vote_map_guard = vote_map.write().await;
+
+        // Get opinions on a voting object and add to vote map.
+        let mut query_voting_objects = |objects: &Vec<VoteObject>| {
+            for (i, object) in objects.iter().enumerate() {
```
m-renaud: I think you can use the `.zip()` approach here to remove the index access.

bee-vote/src/context.rs
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+}
+
+/// Builder pattern struct for instantiating a `VoteContext`.
+pub struct VoteContextBuilder {
+    /// Object of the voting (conflict or timestamp), and associated ID.
+    object: VoteObject,
+    /// Opinions held by this context on the vote object.
+    opinions: Option<Opinions>,
```
pvdrz: Hi, is there a difference between `opinions` being `None` and `opinions` being `Some(Opinions(vec![])`?

Adam-Gleave: No – it gives a nice way to produce an error on building with no opinions through ok_or, but there is no other functional difference. We could definitely achieve the same thing just through an empty check.

bee-vote/src/fpc.rs
Adam-Gleave: Should be fine! Agreed with encapsulating the `Arc` here.

bee-vote/src/context.rs
pvdrz: IMHO I'd avoid the `Option` to have less indirection. But I leave it to your judgement :)

bee-vote/src/fpc.rs
Adam-Gleave: Second approach sounds best here 👍

bee-vote/src/fpc.rs
```diff
@@ -0,0 +1,587 @@
...
+                let mut votes = vote_map_guard
+                    .get(object)
+                    .map_or(Opinions::new(vec![]), |opt| opt.clone());
```
pvdrz: Even though `vec![]` doesn't do allocations maybe using `map_or_else` instead would be better to avoid creating the `default` even if it is not used.

bee-vote/src/fpc.rs
Adam-Gleave: Sure, I can definitely change that.

bee-vote/src/fpc.rs
Adam-Gleave: Thanks for the heads-up on that one – I'll add a more descriptive comment.

bee-vote/src/fpc.rs
Adam-Gleave: Good catch, I'll do that.

bee-vote/src/fpc.rs
Adam-Gleave: Agreed, don't know how I missed that one!

bee-vote/src/fpc.rs
Adam-Gleave: I think maybe something like `calculate_round_thresholds()`? The function doesn't change anything (and takes a shared reference to `self`), but instead uses the context to return the correct set of thresholds for the current round.

bee-vote/src/fpc.rs
Adam-Gleave: Agreed, much cleaner.

bee-vote/src/context.rs
m-renaud: The one reason I can think of to use `Option<Opinions>` is if you wanted to differentiate errors for "opinions never set" (programmer error since `with_initial_opinion(s)` was never called on the builder), or "opinions set empty" (logic/node error? since an empty opinion set was generated when it should be non-empty).

I personally lean towards having more descriptive errors when possible, but no strong opinion here.

bee-vote/src/fpc.rs
Adam-Gleave: Renamed to `opinion_select_count`, which is a bit briefer but I think still makes it clear that it describes the number of times an opinion gets counted/selected.

bee-vote/src/fpc.rs
Adam-Gleave: Settled with the first approach for now actually, because I feel it's a little easier to read and reason around (especially since this code is pretty opaque anyway, and I don't really want to introduce another struct).

bee-vote/src/context.rs
Adam-Gleave: I've removed the `Option`, since it feels a bit more ergonomic that way.

bee-vote/src/fpc.rs
```diff
@@ -405,88 +400,86 @@ where
...
         let vote_map = Arc::new(RwLock::new(HashMap::<VoteObject, Opinions>::new()));
         let all_queried_opinions = Arc::new(RwLock::new(Vec::<QueriedOpinions>::new()));
         let mut query_futures = vec![];
 
-        for (i, opinion_giver) in opinion_givers.iter_mut().enumerate() {
-            // This should never panic, since `queries.len()` == `opinion_givers.len()`
-            let selected_count = queries.get(i).unwrap();
-
-            if *selected_count > 0 {
+        for (opinion_giver, opinion_select_count) in opinion_givers
```
pvdrz: I'd add a tiny comment explaining why those two have the same length.

bee-vote/src/fpc.rs
```diff
@@ -512,28 +505,31 @@ where
...
+        let mut query_voting_objects = |vote_objects: &Vec<VoteObject>| {
+            for (vote_object, opinion) in vote_objects
+                .iter()
+                .zip(opinions.iter()) 
+            {
                 let mut votes = vote_map_guard
-                    .get(object)
-                    .map_or(Opinions::new(vec![]), |opt| opt.clone());
+                    .get(vote_object)
+                    .map_or_else(|| Opinions::new(vec![]), |votes| votes.clone());
```
pvdrz, line 515: I'd reserve the capacity to hold `opinion_select_count` items in the vector upfront.

bee-vote/src/fpc.rs
```diff
@@ -544,13 +540,44 @@ where
...
+
+            // Make sure enough opinions were received for the round to be considered valid.
+            if voted_count < self.min_opinions_received {
+                continue;
+            }
+
+            contexts_guard
+                .get_mut(object)
+                .unwrap()
+                .set_liked(liked_sum / voted_count as f64);
```
pvdrz, line 564: could it happen that all opinions are unknown and this turns out to be a division by zero?

Adam-Gleave: Nice catch, yes that's possible. I'll add a check to make sure `min_opinions_received` cannot be set to 0, therefore if `voted_count` is 0 this will be skipped.

Adam-Gleave:
Switching back to draft for now.

PR #500, opened 2021-04-06
Add password module

thibault-martinez:
# Description of change

Will be used by the password tool and the dashboard/api.

## Change checklist


bee-common/bee-common/src/password.rs
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! A module that provides password utilities.
+
+use argon2::{self, Config};
+use rand::Rng;
+
+/// Generates a salt to be used for password hashing.
+pub fn generate_salt() -> [u8; 32] {
+    rand::thread_rng().gen::<[u8; 32]>()
```
zesterer: From the docs:

> ThreadRng, provided by the thread_rng function, is a handle to a thread-local CSPRNG with periodic seeding from OsRng. Because this is local, it is typically much faster than OsRng. It should be secure, though the paranoid may prefer OsRng.

How paranoid are we? `rand::rngs::OsRng::default().gen()` might be better if that answer is 'very' (if this isn't hot code, which I assume it isn't, then there's an even stronger case for it).

I'm pretty sure the type hint can be removed too.

thibault-martinez: Thanks, will replace, definitely not a hot path!

bee-common/bee-common/src/password.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/// Hashes a password together with a salt.
+pub fn password_hash(password: &[u8], salt: &[u8]) -> Result<Vec<u8>, String> {
```
zesterer: Why not propagate the underlying `rust-argon2` error? By turning it into a string, you remove any real possibility of the caller deciding on a strategy to handle the error (or making the error message more useful for the end user). Also, it means unnecessary allocation.

thibault-martinez: Yeah I asked myself about that but I somehow wanted to keep the possibility of switching the internal password lib without causing too much damage 🤔

zesterer: Well you could always just realias it locally

zesterer (approved):

samuel-rufi (approved):

Adam-Gleave (approved):

PR #506, opened 2021-04-12
Add bee-pow crate

thibault-martinez:
## Change checklist


bee-pow/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "bee-pow"
+version = "0.1.0"
+authors = ["IOTA Stiftung"]
+edition = "2018"
+description = "Provides Proof of Work utilities from the IOTA protocol"
```
Alex6323: "from" -> "for"?

thibault-martinez: Indeed

bee-pow/tests/score.rs
```diff
@@ -0,0 +1,30 @@
...
+use bee_pow::score::compute_pow_score;
+
+#[test]
+fn score() {
+    let message: [u8; 21] = [
+        0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x2c, 0x20, 0x57, 0x6f, 0x72, 0x6c, 0x64, 0x21, 0x5e, 0xe6, 0xaa, 0xaa, 0xaa,
+        0xaa, 0xaa, 0xaa,
+    ];
+
+    assert!((compute_pow_score(&message) - 937.2857142857143).abs() < f64::EPSILON);
```
Adam-Gleave, line 17: Maybe a comment here about where the magic number comes from – it's not immediately obvious (to me at least!)

thibault-martinez: That's just random for the tests, but I'll add links to where I took the tests from.

Adam-Gleave: Yep, makes sense after following the link!

bee-pow/README.md
```diff
@@ -0,0 +1 @@
+# bee-pow
```
Alex6323, line 1: I think that we should - at least for our released crates - write proper READMEs and crate-level documentation. Other Rust libraries in our ecosystem do that already nicely. It will make our libraries much more accessible to the developer community.

bee-pow/src/providers/constant.rs
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+}
+
+/// A nonce provider that return constant nonces.
+pub struct Constant {
+    value: u64,
+}
+
+impl NonceProvider for Constant {
```
Alex6323: Why do we not just `impl NonceProvider for u64`. Is the extra `Constant` type really necessary?

bee-pow/src/providers/miner.rs
```diff
@@ -0,0 +1,146 @@
...
+use std::{
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread,
+};
+
+const DEFAULT_NUM_WORKERS: usize = 1;
+/// https://oeis.org/A002391
```
Alex6323: Doesn't have to be a doc comment, because it's not exported. The link is fine and works, but an explanation why we need it would be helpful.

bee-pow/src/providers/miner.rs
```diff
@@ -0,0 +1,146 @@
...
+const LN_3: f64 = 1.098_612_288_668_109_8;
+
+/// Errors occurring when computing nonces with the `Miner` nonce provider.
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("The worker has been cancelled.")]
+    Cancelled,
+}
```
Alex6323: Probably seems overly nitpicky (but hey...it's monday). but despite the error not having many variants I'ld stick to what we do in many other crates, and that is: having a dedicated `error.rs` file. The thing is if you stick to such a structure consistently then orienting within a codebase becomes really easy.

Adam-Gleave (dismissed):

bee-pow/src/providers/miner.rs
```diff
@@ -0,0 +1,146 @@
...
+/// Builder for the `Miner` nonce provider.
+#[derive(Default)]
+pub struct MinerBuilder {
+    num_workers: Option<usize>,
+}
+
+impl MinerBuilder {
+    /// Sets the desired number of workers for the `Miner` nonce provider.
+    pub fn with_num_workers(mut self, num_workers: usize) -> Self {
+        self.num_workers = Some(num_workers);
```
Alex6323: using `Option::replace` instead of assignment reads better IMHO.

bee-pow/src/providers/miner.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+impl Miner {
+    fn worker(
+        done: Arc<AtomicBool>,
+        pow_digest: TritBuf<T1B1Buf>,
+        start_nonce: u64,
+        target_zeros: usize,
+    ) -> Result<u64, Error> {
+        assert!(target_zeros <= HASH_LENGTH);
```
Alex6323: how likely is this assert to ever cause a panic? I wonder if we should make this `debug_assert`. If `target_zeros` is something the user controls in the config, then I think we should catch this there already instead of here.

bee-pow/src/providers/miner.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        while !done.load(Ordering::Relaxed) {
+            for (i, buffer) in buffers.iter_mut().enumerate() {
+                let nonce_trits = b1t6::encode::<T1B1Buf>(&(nonce + i as u64).to_le_bytes());
+                buffer[pow_digest.len()..pow_digest.len() + nonce_trits.len()].copy_from(&nonce_trits);
+                hasher.add(buffer.clone());
+            }
+
+            for (i, hash) in hasher.hash_batched().enumerate() {
+                let trailing_zeros = hash.iter().rev().take_while(|t| *t == Btrit::Zero).count();
```
Alex6323, line 130: `take_while(Btrit::Zero)`

bee-pow/src/providers/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+    type Provider: NonceProvider<Builder = Self>;
+
+    /// Creates a new nonce provider builder.
+    fn new() -> Self;
+
+    /// Constructs the nonce provider from the builder.
+    fn finish(self) -> Self::Provider;
+}
+
+/// A trait descriving how a nonce is provided.
```
Alex6323: typo: descri**b**ing

bee-pow/src/providers/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+/// A trait to build nonce providers.
+pub trait NonceProviderBuilder: Default + Sized {
+    /// The type of the built nonce provider.
+    type Provider: NonceProvider<Builder = Self>;
+
+    /// Creates a new nonce provider builder.
+    fn new() -> Self;
+
+    /// Constructs the nonce provider from the builder.
+    fn finish(self) -> Self::Provider;
```
Alex6323, line 20: I propose that the `finish` method returns the `Arc<AtomicBool>` to cancel the miner. The `nonce` method signature then becomes simpler only taking `bytes` and `target_score`.

bee-pow/src/score.rs
```diff
@@ -0,0 +1,46 @@
...
+use crypto::hashes::{blake2b::Blake2b256, Digest};
+
+/// Computes the Proof of Work score of given bytes.
+pub fn compute_pow_score(bytes: &[u8]) -> f64 {
+    pow_score(&pow_hash(bytes), bytes.len())
+}
+
+/// Returns the Proof of Work hash of given bytes.
+pub fn pow_hash(bytes: &[u8]) -> TritBuf<T1B1Buf> {
+    assert!(bytes.len() >= std::mem::size_of::<u8>());
```
Alex6323: if we don't control the bytes length, then it's probably better to return a `Result`, rather than panicking here. If we do, then `debug_assert!` should be enough I think.

bee-pow/src/score.rs
```diff
@@ -0,0 +1,46 @@
...
+    pow_input.push(Btrit::Zero);
+    pow_input.push(Btrit::Zero);
+    pow_input.push(Btrit::Zero);
+
+    curl.digest(&pow_input).unwrap()
+}
+
+/// Returns the Proof of Work score of a Proof of Work hash.
+pub fn pow_score(pow_hash: &Trits<T1B1>, len: usize) -> f64 {
+    let zeros = pow_hash.iter().rev().take_while(|t| *t == Btrit::Zero).count() as u32;
```
Alex6323: I've seen this done before. Maybe replace it with an inline flagged function `count_zeros` or more high-level even like `get_difficulty`.

bee-pow/src/providers/miner.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+impl NonceProvider for Miner {
+    type Builder = MinerBuilder;
+    type Error = Error;
+
+    fn nonce(&self, bytes: &[u8], target_score: f64, done: Option<Arc<AtomicBool>>) -> Result<u64, Self::Error> {
+        let mut nonce = 0;
+        let mut pow_digest = TritBuf::<T1B1Buf>::new();
+        let done = done.unwrap_or_else(|| Arc::new(AtomicBool::new(false)));
```
Alex6323: See my proposal regarding the `Arc<AtomicBool>`. I think it would be a better design if the `Arc<AtomicBool>` would be created by the builder. We then wouldn't need to create it here, if the user didn't hand us one.

Alex6323 (changes requested):

bee-pow/src/providers/mod.rs
thibault-martinez: But that would mean the Provider is only usable once, or we would have to reset the bool but it's a kinda weird API

bee-pow/src/providers/miner.rs
thibault-martinez: `take_while` takes an `FnMut`.

bee-pow/src/providers/miner.rs
thibault-martinez: That's the error variant specific to the miner provider.

bee-pow/src/providers/mod.rs
Alex6323: Hm, I don't think it would be a weird API tbh. The `Arc<AtomicBool` already says, that it's shared between the threads and the user, and anyone can change its value. So if the user wants to reuse the provider s/he can also expect that the `Arc<AtomicBool>` (which I maybe would wrap in a convenience type) has been reset.

bee-pow/src/providers/miner.rs
Alex6323: What errors can this crate produce? Would it make sense to have a top-level `error.rs` which contains one variant called `MinerProviderCancelled` or simlar?

thibault-martinez: That's not really possible, it's a trait with an error variant so that people implementing their provider provide their own set of errors.

bee-pow/README.md
Adam-Gleave: I agree – it's something I've tried to do with the FPC stuff, and I think it would be useful to have everywhere. Some summary comments and perhaps an example of how to use the crate functionality would go a log way.

bee-pow/src/providers/miner.rs
Alex6323: alright. I understand!

bee-pow/src/providers/constant.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Contains a nonce provider that returns a given constant nonce.
+
+use crate::providers::{NonceProvider, NonceProviderBuilder};
+
+const DEFAULT_VALUE: u64 = 0;
```
m-renaud: Nit: `DEFAULT_NONCE_VALUE` ?

bee-pow/src/providers/miner.rs
```diff
@@ -0,0 +1,157 @@
...
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread,
+};
+
+const DEFAULT_NUM_WORKERS: usize = 1;
+// Precomputed natural logarithm of 3 for performance reasons.
+// See https://oeis.org/A002391.
+const LN_3: f64 = 1.098_612_288_668_109_8;
```
m-renaud: Maybe a silly question, but where does the last `_8` come from? Looking at the decimal expansion the next two digits are `69` which if rounding should go to `7`, not `8`, right?

thibault-martinez: Nice catch, I have no idea how that happened :)

thibault-martinez: Ahhh it was actually clippy, the value with a 7 is too low/precise to be represented as an f64. I just removed it, it's precise enough anyway.

bee-pow/src/providers/constant.rs
thibault-martinez: Done, thanks

bee-pow/README.md
thibault-martinez: I won't have time to do that and we need to merge. I added a simple sentence for now and we'll improve later.

Alex6323: okay :+1:

Alex6323 (dismissed):

Adam-Gleave (dismissed):

samuel-rufi (dismissed):

Alex6323 (approved):
I like the type names xD

Adam-Gleave (approved):

samuel-rufi (approved):

PR #523, opened 2021-04-19
Add ledger types

thibault-martinez:
## Change checklist


bee-ledger/src/types/conflict.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// The referenced Utxo cannot be found.
+    InputUtxoNotFound = 3,
+    /// The sum of the inputs and output values does not match.
+    InputOutputSumMismatch = 4,
+    /// The unlock block signature is invalid.
+    InvalidSignature = 5,
+    /// The dust allowance for the address is invalid.
+    InvalidDustAllowance = 6,
+    /// The semantic validation failed.
+    SemanticValidationFailed = 255,
```
Adam-Gleave, line 23: Does this need any more specifics? I imagine there's more than one reason that semantic validation can fail.

bee-ledger/src/types/error.rs
```diff
@@ -0,0 +1,45 @@
...
+    NegativeBalance(i64),
+    /// Balance overflow.
+    #[error("Balance overflow: {0}")]
+    BalanceOverflow(i128),
+    /// Balance diff overflow.
+    #[error("Balance diff overflow: {0}")]
+    BalanceDiffOverflow(i128),
+    /// Option error happened
+    #[error("Option error happened")]
+    Option,
```
Adam-Gleave: I feel like this should have a different name, since `Option` is such a commonly used type. Also maybe a more detailed description – what is an Option error?

bee-ledger/src/types/conflict.rs
thibault-martinez: The more specific semantic errors are actually all the other variants. This is kinda just a placeholder in case something unexpected happens. Also, this is part of the protocol so we can't really change it.

bee-ledger/src/types/error.rs
thibault-martinez: I know! This bothered me a lot, it comes from packing/unpacking an option: https://github.com/iotaledger/bee/blob/dev/bee-common/bee-common/src/packable.rs#L123

bee-ledger/src/types/conflict.rs
Adam-Gleave: Makes sense, I wondered if there was a reason for it 👍

thibault-martinez: I can add a comment anyway

bee-ledger/src/types/error.rs
Adam-Gleave: Maybe split into `PackOption`/`UnpackOption` variants (or something similar) would make it clearer, if that's doable? Plus a clearer comment 🙂

zesterer (dismissed):

bee-ledger/src/types/balance.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub fn apply_diff(self, diff: &BalanceDiff) -> Result<Self, Error> {
+        let amount = (self.amount as i64)
+            .checked_add(diff.amount())
+            .ok_or_else(|| Error::BalanceOverflow(self.amount as i128 + diff.amount() as i128))?;
+        let dust_allowance = (self.dust_allowance() as i64)
+            .checked_add(diff.dust_allowance())
+            .ok_or_else(|| Error::BalanceOverflow(self.dust_allowance() as i128 + diff.dust_allowance() as i128))?;
+        let dust_outputs = (self.dust_outputs as i64)
+            .checked_add(diff.dust_outputs())
+            .ok_or_else(|| Error::BalanceOverflow(self.dust_outputs as i128 + diff.dust_outputs() as i128))?;
```
Alex6323, line 60: `self.amount/dust_allowance/dust_outputs as i64` will wrap if those values contain values that `> u64::MAX/2`. Are we sure that isn't possible. If not, then better use `as i128` conversions.

Alex6323 (changes requested):
Review pt1

bee-ledger/src/types/receipt.rs
```diff
@@ -0,0 +1,114 @@
...
+    included_in: MilestoneIndex,
+}
+
+impl Receipt {
+    /// Creates a new `Receipt`.
+    pub fn new(inner: ReceiptPayload, included_in: MilestoneIndex) -> Self {
+        Self { inner, included_in }
+    }
+
+    /// Returns the inner receipt of the `Receipt`.
```
Alex6323, line 28: /// Returns the payload of the `Receipt`.

bee-ledger/src/types/receipt.rs
```diff
@@ -0,0 +1,114 @@
...
+    input::Input,
+    milestone::MilestoneIndex,
+    output::Output,
+    payload::{receipt::ReceiptPayload, Payload},
+};
+
+/// A type that wraps a receipt and the index of the milestone in which it was included.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Receipt {
+    inner: ReceiptPayload,
```
Alex6323, line 18: consider renaming `inner` to `payload` or `receipt_payload`

bee-ledger/src/types/balance.rs
thibault-martinez: Yeah it's all already syntactically bound by the number of IOTA tokens but I guess it wouldn't hurt to use i128, will do

bee-ledger/src/types/receipt.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl Receipt {
+    /// Creates a new `Receipt`.
+    pub fn new(inner: ReceiptPayload, included_in: MilestoneIndex) -> Self {
+        Self { inner, included_in }
+    }
+
+    /// Returns the inner receipt of the `Receipt`.
+    pub fn inner(&self) -> &ReceiptPayload {
```
Alex6323, line 29: I'ld suggest renaming this API. I'ld only name it `inner` if we'ld also find it logical to deref to this "inner" type, which in this case we probably don't.

bee-ledger/src/types/receipt.rs
```diff
@@ -0,0 +1,114 @@
...
+        };
+
+        for funds in self.inner().funds() {
+            migrated_amount = migrated_amount.checked_add(funds.output().amount()).ok_or_else(|| {
+                Error::InvalidMigratedFundsAmount(migrated_amount as u128 + funds.output().amount() as u128)
+            })?;
+        }
+
+        if migrated_amount > IOTA_SUPPLY {
+            return Err(Error::InvalidMigratedFundsAmount(migrated_amount as u128));
```
Alex6323: It would probably be neat, if the error provided the maxium allowed amount, but I leave it up to you to change this.

bee-ledger/src/types/receipt.rs
```diff
@@ -0,0 +1,114 @@
...
+            output => return Err(Error::UnsupportedOutputKind(output.kind())),
+        };
+
+        let created_amount = consumed_treasury_output
+            .inner()
+            .amount()
+            .checked_sub(migrated_amount)
+            .ok_or_else(|| {
+                Error::InvalidMigratedFundsAmount(
+                    consumed_treasury_output.inner().amount() as u128 - migrated_amount as u128,
```
Alex6323: this will also underflow for `u128`. Did you meant to write `i128`?

Alex6323 (changes requested):
Review pt2

bee-ledger/src/types/receipt.rs
thibault-martinez: Good catch

bee-ledger/src/types/receipt.rs
thibault-martinez: I don't really like payload as it gives the idea that it's an actual `Payload` type and then can be generic. I like `inner` because `Receipt` is just a wrapper type.

bee-ledger/src/types/receipt.rs
thibault-martinez: I'll see what I can do

bee-ledger/src/types/receipt.rs
thibault-martinez: I think it's fine

bee-ledger/src/types/balance.rs
thibault-martinez: Actually it's not possible, the checked operations wouldn't work anymore

thibault-martinez: Also, the `self.amount/dust_allowance/dust_outputs` have to be the result of a previous call to `apply_diff ` and we start with all `0` so this just can't happen.

PR #525, opened 2021-04-19
Add bee-network crate with initial types

Alex6323:


bee-network/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Networking functionality and types for nodes and clients participating in the IOTA protocol built on top of
+//! `libp2p`.
```
thibault-martinez, line 5: Oh no, 121 chars :(

thibault-martinez: Gotta remove a word :trollface:

Alex6323: yea, too bad :sweat_smile:

thibault-martinez (dismissed):

bee-network/src/peer/info.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Represents a known peer.
+    Known,
+    /// Represents an unknown peer.
+    Unknown,
+}
+
+impl PeerRelation {
+    /// Returns whether the peer is known.
+    pub fn is_known(&self) -> bool {
+        self.eq(&Self::Known)
```
zesterer: Why not `*self == Self::Known`, out of interest?

bee-network/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+#[doc(inline)]
+pub use libp2p::{
+    core::identity::{ed25519::Keypair, PublicKey},
+    multiaddr::Protocol,
+    Multiaddr, PeerId,
+};
+
+/// Creates a (shorter) peer alias from a peer id.
+#[macro_export]
+macro_rules! alias {
```
zesterer, line 24: Can this not be a function?

bee-network/src/peer/info.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /// Returns whether the peer is unknown.
+    pub fn is_unknown(&self) -> bool {
+        self.eq(&Self::Unknown)
```
samuel-rufi: instead of calling `self.eq(&Self::Unknown)`, what about calling `!is_known()` here? what would be best-practice?

thibault-martinez: That would probably not play very well when (if) we introduce new enum variants

samuel-rufi (dismissed):

bee-network/src/peer/info.rs
Alex6323: No particular reason. Can't even remember why I opted to go with less syntactic sugar here :shrug: I guess people like `==` better, so I'll probably change. Should I? Please put a :+1:  or :-1: if you guys have a preferance.

pvdrz: I'd use `matches!` in case someone decides that the variants should have fields in the future

thibault-martinez: `matches!` is probably my favourite thing <3

bee-network/src/lib.rs
Alex6323: It could ... of course, but in this particular case I decided against it to free the user from having to care about whether s/he works with an owned or borrowed peer id.

bee-network/src/peer/info.rs
Alex6323: That's right. It looks like we will be having more enum variants in the future (`Discovered` possibly). Another good practice and alternative here is to use the `matches!` macro.

pvdrz: `matches!`-gang assemble

bee-network/src/peer/info.rs
Alex6323: Haha, that's true. Thibault loves the `matches!` macro. His specialty is to do something like `!matches!`.  I like that you can even put guards in there.

Adam-Gleave: +1 for `matches!` ✊

pvdrz (dismissed):

bee-network/src/peer/info.rs
Alex6323: Done

bee-network/src/lib.rs
Alex6323: Further context: it's often used in debugging scenarios: 
`debug!("Peer {} didn't play by the rulez!", alias!(very_bad_peer))`.

Adam-Gleave (dismissed):

thibault-martinez (dismissed):

pvdrz (dismissed):

pvdrz (dismissed):

thibault-martinez (approved):

Adam-Gleave (approved):

PR #527, opened 2021-04-19
Add bee-tangle crate

zesterer:
# Description of change

Moved `bee-tangle` from the `chrysalis-pt-2` branch

## Type of change

`bee-tangle` has existed in `chrysalis-pt-2` for a while now and has been well-tested.

## Change checklist


bee-tangle/src/event.rs
```diff
@@ -0,0 +1,39 @@
...
+pub struct ConfirmedMilestoneChanged {
+    /// The index of the milestone
+    pub index: MilestoneIndex,
+    /// The milestone data
+    pub milestone: Milestone,
+}
+
+/// An event that indicates that a snapshot milestone has had an index change.
+#[derive(Clone)]
+pub struct SnapshotMilestoneIndexChanged(pub MilestoneIndex);
```
pvdrz: It is probably not worth but this could use named fields instead for the sake of consistency

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,303 @@
...
+        self.ymrsi = Some(ymrsi);
+    }
+
+    /// Get the reference timestamp (seconds from the unix epoch) of this message.
+    pub fn reference_timestamp(&self) -> u64 {
+        self.reference_timestamp
+    }
+
+    /// Solidify this message at the current system time.
+    pub fn solidify(&mut self) {
```
pvdrz: I'd rename this as `mark_as_solid` or something so no-one thinks this magically triggers solidification somehow.

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /// Reference this message with the given timestamp.
+    pub fn reference(&mut self, timestamp: u64) {
+        self.flags.set_referenced(true);
+        self.reference_timestamp = timestamp;
+    }
+
+    /// Get the conflict state of this message.
+    pub fn conflict(&self) -> u8 {
```
pvdrz: Is there any reason for the conflict value to not be a newtype?

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,303 @@
...
+    pub fn set_conflict(&mut self, conflict: u8) {
+        self.conflict = conflict;
+    }
+}
+
+/// An error that may occur when manipulating message metadata.
+#[derive(Debug)]
+pub enum MessageMetadataError {
+    /// An IO error occurred.
+    Io(std::io::Error),
```
pvdrz, line 151: If this variant comes from `Packable` maybe we could use `Packable::Error`  instead

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,303 @@
...
+        self.0
+    }
+
+    /// Get the message ID of this `IndexId`.
+    pub fn id(&self) -> MessageId {
+        self.1
+    }
+
+    /// Update this `IndexId` with a new milestone index.
+    pub fn update(&mut self, index: MilestoneIndex) {
```
pvdrz: `set/update_index` might be better name?

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        Ok(Self(index, id))
+    }
+}
+
+/// An error that may occur when manipulating message indices.
+#[derive(Debug)]
+pub enum IndexIdError {
+    /// An IO error occurred.
+    Io(std::io::Error),
```
pvdrz, line 292: same question about `Packable`

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+    }
+
+    async fn insert_approver(&self, msg: MessageId, approver: MessageId) -> Result<(), Self::Error> {
+        trace!("Attempted to insert approver for message {:?}", msg);
+        self.storage.insert(&(msg, approver), &()).await
+    }
+
+    async fn update_approvers(&self, msg: MessageId, approvers: &[MessageId]) -> Result<(), Self::Error> {
+        trace!("Attempted to update approvers for message {:?}", msg);
+        // self.storage.insert(&msg, approvers).await
```
pvdrz: old code I guess??

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+        Ok(())
+    }
+}
+
+/// A Tangle wrapper designed to encapsulate milestone state.
+pub struct MsTangle<B> {
+    config: TangleConfig,
+    pub(crate) inner: Tangle<MessageMetadata, StorageHooks<B>>,
+    milestones: Mutex<HashMap<MilestoneIndex, Milestone>>,
+    pub(crate) solid_entry_points: Mutex<HashMap<SolidEntryPoint, MilestoneIndex>>,
```
pvdrz: I'd rather not expose this `Mutex` to the whole crate because ti makes deadlocks really hard to find.

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+            snapshot_index: Default::default(),
+            pruning_index: Default::default(),
+            entry_point_index: Default::default(),
+            tip_pool: Mutex::new(UrtsTipPool::default()),
+        }
+    }
+
+    /// Shut down the tangle, terminating any and all worker tasks.
+    pub async fn shutdown(self) {
+        // TODO: Write back changes by calling self.inner.shutdown().await
```
pvdrz, line 133: Is this TODO intended to be here after merge?

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+        //     }
+        // }
+        //
+        // opt
+        self.inner.insert(hash, message, metadata).await
+    }
+
+    /// Add a milestone to the tangle.
+    pub async fn add_milestone(&self, idx: MilestoneIndex, milestone: Milestone) {
+        // TODO: only insert if vacant
```
pvdrz, line 148: same question about `TODO`

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+                metadata.set_omrsi(IndexId::new(idx, *milestone.message_id()));
+                metadata.set_ymrsi(IndexId::new(idx, *milestone.message_id()));
+            })
+            .await;
+        self.inner
+            .hooks()
+            .insert_milestone(idx, &milestone)
+            .await
+            .unwrap_or_else(|e| info!("Failed to insert message {:?}", e));
+        // TODO can there be a race condition between 2 ops ?
```
pvdrz: I'd add a log entry before and after acquiring the lock just to be sure.

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+        self.milestones.lock().await.insert(idx, milestone);
+    }
+
+    /// Remove a milestone from the tangle.
+    pub async fn remove_milestone(&self, index: MilestoneIndex) {
+        self.milestones.lock().await.remove(&index);
```
pvdrz, line 167: Same here.

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+
+            Some(message_id)
+        } else {
+            None
+        }
+    }
+
+    /// Get the milestone from the tangle that corresponds to the given milestone index.
+    pub async fn get_milestone(&self, index: MilestoneIndex) -> Option<Milestone> {
+        self.milestones.lock().await.get(&index).cloned()
```
pvdrz, line 191: Same here.

bee-tangle/src/ms_tangle.rs
pvdrz: Maybe we could have a wrapper for this. @Adam-Gleave might have some opinions about this.

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+    pub fn get_solid_milestone_index(&self) -> MilestoneIndex {
+        self.solid_milestone_index.load(Ordering::Relaxed).into()
+    }
+
+    /// Update the latest solid milestone index.
+    pub fn update_solid_milestone_index(&self, new_index: MilestoneIndex) {
+        self.solid_milestone_index.store(*new_index, Ordering::Relaxed);
+
+        // TODO: Formalise this a little better
+        let new_len = ((1000.0 + self.get_sync_threshold() as f32 * 500.0) as usize)
```
pvdrz, line 244: maybe letting the cast after the product is faster. Could it happen that this overflows in that case?

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,1145 @@
...
+        self.tip_pool.lock().await.reduce_tips();
+    }
+
+    /// Return the number of non-lazy tips.
+    pub async fn non_lazy_tips_num(&self) -> usize {
+        self.tip_pool.lock().await.non_lazy_tips().len()
+    }
+}
+
+// #[cfg(test)]
```
pvdrz: Is this intended to be commented?

bee-tangle/src/tangle.rs
```diff
@@ -0,0 +1,535 @@
...
+
+/// A foundational, thread-safe graph datastructure to represent the IOTA Tangle.
+pub struct Tangle<T, H = NullHooks<T>>
+where
+    T: Clone,
+{
+    // Global Tangle Lock. Remove this as and when it is deemed correct to do so.
+    vertices: TRwLock<HashMap<MessageId, Vertex<T>>>,
+
+    pub(crate) cache_queue: Mutex<LruCache<MessageId, (), DefaultHashBuilder>>,
```
pvdrz: same question about `pub(crate)` mutexes

bee-tangle/src/tangle.rs
```diff
@@ -0,0 +1,535 @@
...
+    pub fn hooks(&self) -> &H {
+        &self.hooks
+    }
+
+    async fn insert_inner(
+        &self,
+        message_id: MessageId,
+        message: Message,
+        metadata: T,
+        prevent_eviction: bool,
```
pvdrz, line 140: could this argument be a constant?

bee-tangle/src/tangle.rs
```diff
@@ -0,0 +1,535 @@
...
+
+        let msg = self
+            .insert_inner(message_id, message.clone(), metadata.clone(), !exists)
+            .await;
+
+        self.vertices
+            .write()
+            .await
+            .get_mut(&message_id)
+            .expect("Just-inserted message is missing")
```
pvdrz, line 193: could this happen if something runs between `insert_inner` and this?

pvdrz, line 195: could this happen if something runs between `insert_inner` and this?

bee-tangle/src/tangle.rs
```diff
@@ -0,0 +1,535 @@
...
+                    }
+                } else {
+                    break;
+                }
+            }
+        }
+    }
+}
+
+// #[cfg(test)]
```
pvdrz: Should this be here?

bee-tangle/src/traversal.rs
```diff
@@ -0,0 +1,170 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Collection of Tangle traversal functions.
+
+// TODO: Refactor all of this into methods on `Tangle`.
```
pvdrz, line 6: Is this `TODO` intended for this PR?

bee-tangle/src/traversal.rs
```diff
@@ -0,0 +1,170 @@
...
+                .get_vertex(&message_id)
+                .await
+                .as_ref()
+                .and_then(|v| v.message_and_metadata().cloned());
+            match msg_meta {
+                Some((msg, meta)) => {
+                    if matches(message_id, msg.clone(), meta).await {
+                        apply(&message_id, &msg, &meta);
+
+                        for parent in msg.parents().iter() {
```
pvdrz: could we use `extend_from_slice` or preallocate space in `parents` to avoid multiple allocation calls?

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,197 @@
...
+                metadata.children.insert(child);
+                metadata.time_first_child = Some(Instant::now());
+                entry.insert(metadata);
+            }
+        }
+    }
+
+    fn check_retention_rules_for_parent(&mut self, parent: &MessageId) {
+        if self.non_lazy_tips.len() > MAX_LIMIT_NON_LAZY as usize
+            || self.tips.get(parent).unwrap().children.len() as u8 > MAX_NUM_CHILDREN
```
pvdrz: could we `self.tip.get(parent)` once? also, why is it ok to unwrap this?

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,197 @@
...
+                    .iter()
+                    .map(|t| **t)
+                    .collect()
+            })
+        }
+    }
+
+    pub(crate) fn optimal_num_tips(&self) -> usize {
+        // TODO: hardcoded at the moment
+        4
```
pvdrz, line 184: Insert joke about IEEE standards here :trollface:

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    pub(crate) fn reduce_tips(&mut self) {
+        let mut to_remove = Vec::new();
+        for (tip, metadata) in &self.tips {
+            if let Some(age) = metadata.time_first_child {
+                if age.elapsed().as_secs() as u8 > MAX_AGE_SECONDS_AFTER_FIRST_CHILD {
```
pvdrz: is `u8` large enough to hold this elapsed time? or is the truncation intentional?

bee-tangle/src/vertex.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use std::sync::Arc;
+
+#[derive(Clone)]
+pub struct Vertex<T>
+where
+    T: Clone,
+{
+    message: Option<(MessageRef, T)>,
+    children: (VecSet<MessageId>, bool), // Exhaustive flag
```
pvdrz, line 16: I'd split this for readability

bee-tangle/src/vertex.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.eviction_blocks -= 1;
+        assert!(self.eviction_blocks >= 0);
+    }
+
+    pub(crate) fn can_evict(&self) -> bool {
+        self.eviction_blocks == 0
+    }
+}
+
+// #[cfg(test)]
```
pvdrz: Should this be commented?

thibault-martinez:
Any idea why CI didn't run on this PR ?

bee-tangle/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "bee-tangle"
+version = "0.1.0-alpha"
+authors = ["IOTA Stiftung"]
+edition = "2018"
+description = ""
```
thibault-martinez: A one sentence description would be nice

bee-tangle/README.md
```diff
@@ -0,0 +1 @@
+# bee-tangle
```
thibault-martinez: And put the same sentence here as well

bee-tangle/src/config.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2020 IOTA Stiftung
```
thibault-martinez: Can you update all headers to "2020-2021" please ?

bee-tangle/src/ms_tangle.rs
thibault-martinez: It's very old code, let's not keep it any longer.

bee-tangle/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+dashmap = "4.0"
+futures = "0.3"
+log = "0.4"
+lru = "0.6"
+rand = "0.8"
+ref-cast = "1.0"
+serde = { version = "1.0", features = ["derive" ] }
+thiserror = "1.0"
+tokio = { version = "1.4", features = ["sync", "time"] }
+hashbrown = "0.11"
```
Adam-Gleave: "H" is after "T" 😄

bee-tangle/src/ms_tangle.rs
Adam-Gleave: If we're not sure something bad could happen here, probably best if we do something along those lines in case we happen to run into it. Would save a lot of time debugging. Something like the wrapper we were using for `wallet.rs` may be a little heavyweight, though.

bee-tangle/src/traversal.rs
```diff
@@ -0,0 +1,170 @@
...
+use crate::{
+    tangle::{Hooks, Tangle},
+    MessageRef,
+};
+
+use bee_message::MessageId;
+
+use std::{collections::HashSet, future::Future};
+
+// /// A Tangle walker that - given a starting vertex - visits all of its ancestors that are connected through
```
Adam-Gleave: Old code?

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    async fn tip_score<B: StorageBackend>(&self, tangle: &MsTangle<B>, message_id: &MessageId) -> Score {
+        // in case the tip was pruned by the node, consider tip as lazy
+        if !tangle.contains(message_id).await {
+            return Score::Lazy;
+        }
+
+        let smi = *tangle.get_solid_milestone_index();
+        let omrsi = *tangle.omrsi(&message_id).await.unwrap().index();
```
Adam-Gleave: Maybe document these `unwrap`s.

bee-tangle/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "bee-tangle"
+version = "0.1.0-alpha"
```
samuel-rufi: do we keep alpha here?

bee-tangle/CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+
+### Security -->
```
samuel-rufi, line 20: Maybe add some release notes as we have done in bee-network or bee-rest-api

bee-tangle/src/traversal.rs
```diff
@@ -0,0 +1,170 @@
...
+                None => {
+                    missing_apply(&message_id);
+                }
+            }
+            visited.insert(message_id);
+        }
+    }
+}
+
+// TODO reimplement with multiple parents
```
samuel-rufi: should we remove old code like this?

bee-tangle/src/urts.rs
samuel-rufi: you're right this must be changed, would this work?
`if age.elapsed().as_secs() > MAX_AGE_SECONDS_AFTER_FIRST_CHILD as u64`

bee-tangle/CHANGELOG.md
thibault-martinez, line 20: Prepare a release

bee-tangle/Cargo.toml
thibault-martinez: Remove alpha

bee-tangle/src/vec_set.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(dead_code)]
```
thibault-martinez: What's not being used and why not to remove it ?

bee-tangle/src/urts.rs
samuel-rufi: Line 73 makes sure the parent is available, so it's safe to unwrap()
You can use something like that as comment:

"For every tip we add to the pool we call `add_child()`. `add_child()` makes sure that the parents of the tip are present in the pool. Since `check_retention_rules_for_parent()` will be called after `add_child()` we can be sure that the parents do exist. Therefore, unwrapping the parents here is safe."

bee-tangle/src/solid_entry_point.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! The solid_entry_point module defined the [SolidEntryPoint] type which represents
+//! an already solidified message in the tangle.
```
thibault-martinez: I feel like this definition is a bit off. The other one you use later in more correct.

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    fn check_retention_rules_for_parent(&mut self, parent: &MessageId) {
+        if self.non_lazy_tips.len() > MAX_LIMIT_NON_LAZY as usize
+            || self.tips.get(parent).unwrap().children.len() as u8 > MAX_NUM_CHILDREN
+            || self
+                .tips
+                .get(parent)
+                .unwrap()
```
samuel-rufi, line 110: Also this unwrap is safe, for the same reason:

"For every tip we add to the pool we call `add_child()`. `add_child()` makes sure that the parents of the tip are present in the pool. Since `check_retention_rules_for_parent()` will be called after `add_child()` we can be sure that the parents do exist. Therefore, unwrapping the parents here is safe."

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,197 @@
...
+                .time_first_child
+                .unwrap()
```
samuel-rufi, line 112: Also this unwrap is safe since of the add_child function:

"In case the parents of the tip were not present in the pool, `add_child()` will add them. Furthermore it will set `metadata.time_first_child`. `time_first_child` therefore will always be "some" and unwrapping here is safe."

bee-tangle/src/urts.rs
samuel-rufi: Also these unwraps are safe since we only add `solid` tips to the tip pool
All of the solid tips do have omris/ymrsi values set. So unwrapping should be safe.
We need this as a comment here

bee-tangle/src/urts.rs
zesterer: image

bee-tangle/src/vertex.rs
zesterer: I'd rather not. The exhaustiveness flag is always relevant when accessing `children`, so I want to make it as clear as possible that they're mutually-entangled bits of state, and more difficult to forget about the flag.

zesterer:
> Any idea why CI didn't run on this PR ?

bee-tangle/src/metadata.rs
zesterer: I'm not sure. @thibault-martinez do you have opinions on this?

bee-tangle/src/metadata.rs
zesterer: What do you mean by this? There's no such associated type defined that would let us do what I think you're asking for

bee-tangle/src/urts.rs
zesterer: @rufsam Do you know what the original reason for choosing `u8` here was?

bee-tangle/src/vertex.rs
zesterer: I'm going to remove it entirely, it no longer tests things we care about.

bee-tangle/README.md
zesterer: > The distributed DAG that sits at the heart of IOTA's Bee node

How does this sound?

bee-tangle/src/traversal.rs
zesterer: @thibault-martinez Do you want this code kept?

bee-tangle/src/urts.rs
zesterer: @rufsam Do you think you could come up with a comment-worth justification for them? It seems like that explanation is relying on some very non-local reasoning

bee-tangle/src/urts.rs
samuel-rufi: if i remember correctly `age.elapsed().as_secs() as u8` was choosen only to be able to compare it with `MAX_AGE_SECONDS_AFTER_FIRST_CHILD` which is u8

bee-tangle/src/vec_set.rs
zesterer: `new` is unused. I think a `new` method is pretty standard for a data structure like this though. I can remove it for now though given that the rest of the API isn't really fleshed out.

bee-tangle/src/solid_entry_point.rs
zesterer: What do you think would be a better definition?

bee-tangle/Cargo.toml
zesterer: Already removed

bee-tangle/CHANGELOG.md
zesterer: Already done this too

bee-tangle/src/urts.rs
samuel-rufi: Maybe comment something like:

"The tip pool only works with solid tips. Therefore, all tips added to the pool can be considered to solid. The solid flag will be set together with `omrsi` and `ymrsi` values. Therefore, when a message is solid, `omrsi` and `ymrsi` values are available. Therefore, unwrapping here is safe."

bee-tangle/src/event.rs
thibault-martinez: Actually for consistency, all events don't have names fields :p

bee-tangle/src/metadata.rs
thibault-martinez: There is actually a type in the ledger crate but we couldn't use it because of a circular dependency. Since, the ledger types have been put in a module that doesn't use the tangle so maybe we can start using it ? It's called `ConflictReason`

bee-tangle/README.md
thibault-martinez: Good

bee-tangle/src/traversal.rs
thibault-martinez: No

bee-tangle/src/solid_entry_point.rs
thibault-martinez: You wrote an actual good definition for the type just below.

bee-tangle/src/urts.rs
zesterer: I've switched it to `u64` now.

bee-tangle/src/urts.rs
zesterer: Cheers

bee-tangle/src/event.rs
zesterer: I just changed it to add named fields... Do you want me to change it back?

pvdrz: well either one :P

bee-tangle/src/metadata.rs
pvdrz: Maybe I'm missing something: https://github.com/iotaledger/bee/blob/dev/bee-common/bee-common/src/packable.rs#L11

bee-tangle/src/urts.rs
pvdrz: I'd add a compact version of this as a comment. Thanks Sam!

bee-tangle/src/vertex.rs
pvdrz: So, maybe a new struct?

bee-tangle/src/ms_tangle.rs
zesterer: I'll see if I can lower the visibility

zesterer: Looks like I can. No idea why it was made `pub(crate)` to begin with.

bee-tangle/src/ms_tangle.rs
zesterer: Yes, it's still relevant (in general, but the implementation *currently* doesn't expose a method like this)

bee-tangle/src/ms_tangle.rs
zesterer: Yes, still relevant

bee-tangle/src/ms_tangle.rs
zesterer: This is actually no longer relevant because the implementation has changed such that there are no longer lock dependencies (i.e: there's no chance of a deadlock).

bee-tangle/src/ms_tangle.rs
zesterer: As mentioned before: this is not a problem because a) the mutex is not exposed and b) there is no way for a caller to hold the lock twice because each method is effectively atomic.

bee-tangle/src/ms_tangle.rs
zesterer: As mentioned before: this is not a problem because a) the mutex is not exposed and b) there is no way for a caller to hold the lock twice because each method is effectively atomic.

bee-tangle/src/ms_tangle.rs
zesterer: I can promise that there are far, far more expensive things to care about on this code path than an extra cast :grin: (and besides, this code only gets run intermittently anyway).

bee-tangle/src/tangle.rs
zesterer: Lowered visibility

bee-tangle/src/tangle.rs
zesterer: Possibly, but there's not much point. There are database accesses, many atomic operations, and hashing operations on this path. A boolean check isn't going to matter. All this aside: the code is not exposed beyond this crate and is invoked from within this crate so LLVM is free to inline & optimise it if it so desires. I'd suggest that leaving it up to LLVM is the more sensible route rather than unintentionally inflating the code cache when it's not optimal to do so by monomorphising multiple instances of what amounts to the same code. The optimiser knows best!

bee-tangle/src/tangle.rs
zesterer: This is actually the race condition that I fixed last week by adding a flagging system that prevents eviction. So, no.

bee-tangle/src/tangle.rs
zesterer: Going to be replacing these with a more comprehensive set of tests anyway, so I'll remove it.

bee-tangle/src/traversal.rs
zesterer: It is not. It's still relevant, but not something to resolve now.

bee-tangle/src/traversal.rs
zesterer: Now that `Parents` has been changed, yes

bee-tangle/src/metadata.rs
zesterer: Yes, but how would you have access to it here?

zesterer: Like... there's no point because the definition of `Packable` relies on `Read` and `Write`. It's implicitly dependent on `std::io::Error` no matter what, so there's no need to obscure the definition of the error by reaching that type through a convoluted path involving the associated type.

bee-tangle/src/vertex.rs
zesterer: Why? This is accessed from about 2 locations, all within this file. I'd rather not obscure the definition of an already private field for nebulous gains. The fact of the matter is that `children` is by itself a meaningful name for the values, and it doesn't make sense to (correctly) consider either of them in isolation. Ergo, a tuple.

bee-tangle/src/metadata.rs
pvdrz: well that's a good point

bee-tangle/src/vertex.rs
pvdrz: sure

bee-tangle/tests/traversal.rs
```diff
@@ -0,0 +1,110 @@
+// Copyright 2020 IOTA Stiftung
```
thibault-martinez: Remove

bee-tangle/tests/helper.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright 2020 IOTA Stiftung
```
thibault-martinez: Remove

thibault-martinez (changes requested):

bee-tangle/src/event.rs
zesterer: @thibault-martinez Do you have an opinion on this?

bee-tangle/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "bee-tangle"
+version = "0.1.0"
+authors = ["IOTA Stiftung"]
+edition = "2018"
+description = "The distributed DAG that sits at the heart of IOTA's Bee node"
```
Alex6323: I'ld prefer a more technical, and less metaphorical description. I suggest:
"Implementation of the Tangle, the core data structure for reaching consensus in the IOTA distributed ledger."

bee-tangle/README.md
```diff
@@ -0,0 +1,3 @@
+# `bee-tangle`
+
+The distributed DAG that sits at the heart of IOTA's Bee node.
```
Alex6323: same here :smiley:

bee-tangle/src/flags.rs
```diff
@@ -0,0 +1,81 @@
...
+
+bitflags! {
+    /// Flags representing the state of a message.
+    #[derive(Default, Serialize)]
+    pub struct Flags: u8 {
+        /// The message is solid (confirmed)
+        const SOLID = 0b0000_0001;
+        /// The message is a milestone
+        const MILESTONE = 0b0000_0010;
+        /// The message has been referenced by something
```
Alex6323: by something? I think that "something" is a milestone, although I find this terminology weird, but I think this is the current IOTA speak. @thibault-martinez might correct me on this.

bee-tangle/src/flags.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    /// Return whether the flags indicate that the message is valid.
+    pub fn is_valid(&self) -> bool {
+        self.contains(Flags::VALID)
+    }
+
+    /// Set the valid flag for this message.
+    pub fn set_valid(&mut self, is_valid: bool) {
+        self.set(Flags::VALID, is_valid);
+    }
```
Alex6323, line 63: docs: pls use 3rd person

bee-tangle/src/flags.rs
thibault-martinez: Yes, referenced by a milestone. I don't really like that either but that's how it is...

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,304 @@
...
+            arrival_timestamp,
+            solidification_timestamp,
+            reference_timestamp,
+            omrsi,
+            ymrsi,
+            conflict,
+        }
+    }
+
+    /// Create metadata that corresponds to a just-arrived message using the current system time.
```
Alex6323, line 55: Docs: 3rd person. Main reasons:
* consistency with the rest of the codebase
* consistency with the ecosystem or at least `std`

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,304 @@
...
+            solidification_timestamp: u64::unpack_inner::<R, CHECK>(reader)?,
+            reference_timestamp: u64::unpack_inner::<R, CHECK>(reader)?,
+            omrsi: Option::<IndexId>::unpack_inner::<R, CHECK>(reader)?,
+            ymrsi: Option::<IndexId>::unpack_inner::<R, CHECK>(reader)?,
+            conflict: ConflictReason::unpack_inner::<R, CHECK>(reader)?,
+        })
+    }
+}
+
+/// An identifier for message indices.
```
Alex6323: Since I added this type a long time ago: A more accurate description would probably be:
"A type that is used to associate two particular interesting Cone Root Indexes with a message in the Tangle, i.e. the Oldest Cone Root Index (OCRI), and the Youngest Cone Root Index (YCRI)"
Once merged I probably open a PR to suggest the renaming of this type to "ConeRootIndex".

bee-tangle/src/solid_entry_point.rs
```diff
@@ -0,0 +1,80 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A SolidEntryPoint is a [`MessageId`](crate::MessageId) of a message even if we do not have them
```
Alex6323: The path `(crate::MessageId)` is not necessary anymore (I believe)

bee-tangle/src/unconfirmed_message.rs
```diff
@@ -0,0 +1,53 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use bee_common::packable::{Packable, Read, Write};
+use bee_message::MessageId;
+
+use std::ops::Deref;
+
+/// A type representing an unconfirmed message.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct UnconfirmedMessage(MessageId);
```
Alex6323: `#[repr(transparent)]`

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,204 @@
...
+// C1: the maximum allowed delta value for the YMRSI of a given message in relation to the current SMI before it
+// gets lazy.
+const YMRSI_DELTA: u32 = 8;
+// C2: the maximum allowed delta value between OMRSI of a given message in relation to the current SMI before it
+// gets semi-lazy.
+const OMRSI_DELTA: u32 = 13;
+// M: the maximum allowed delta value between OMRSI of a given message in relation to the current SMI before it
+// gets lazy.
+/// The maximum permitted difference between a message's OMRSI and the current SMI before it is considered lazy.
+pub const BELOW_MAX_DEPTH: u32 = 15;
```
Alex6323: That needs to be read from the config and - I suggest - stored globally. Otherwise the user's configuration will have no effect.

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub(crate) async fn insert<B: StorageBackend>(
+        &mut self,
+        tangle: &MsTangle<B>,
+        message_id: MessageId,
+        parents: Vec<MessageId>,
+    ) {
+        if let Score::NonLazy = self.tip_score::<B>(tangle, &message_id).await {
+            self.non_lazy_tips.insert(message_id);
+            self.tips.insert(message_id, TipMetadata::new());
+            for parent in parents.iter() {
```
Alex6323: why not just `&parents`? :trollface:

bee-tangle/src/urts.rs
Alex6323: please put your reasoning in the code :rocket:

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        let smi = *tangle.get_solid_milestone_index();
+
+        // The tip pool only works with solid tips. Therefore, all tips added to the pool can be considered to solid.
+        // The solid flag will be set together with omrsi and ymrsi values. Therefore, when a message is solid, omrsi
+        // and ymrsi values are available. Therefore, unwrapping here is fine.
+        let omrsi = *tangle.omrsi(&message_id).await.unwrap().index();
+        let ymrsi = *tangle.ymrsi(&message_id).await.unwrap().index();
+
+        if (smi - ymrsi) > YMRSI_DELTA {
```
Alex6323: Since we're operating on `u32`s here, I'ld suggest the following equivalent inequation:
`if smi > YMRSI_DELTA + ymrsi`
This way we won't ever have an underflow no matter how weird the circumstances.

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,204 @@
...
+            return Score::Lazy;
+        }
+
+        if (smi - omrsi) > BELOW_MAX_DEPTH {
```
Alex6323: same argument as above

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,204 @@
...
+            return Score::Lazy;
+        }
+
+        if (smi - omrsi) > OMRSI_DELTA {
```
Alex6323: same argument as above

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,204 @@
...
+        let non_lazy_tips = &self.non_lazy_tips;
+        if non_lazy_tips.is_empty() {
+            None
+        } else {
+            Some(if non_lazy_tips.len() < self.optimal_num_tips() {
+                non_lazy_tips.iter().copied().collect()
+            } else {
+                non_lazy_tips
+                    .iter()
+                    .choose_multiple(&mut rand::thread_rng(), self.optimal_num_tips())
```
Alex6323, line 174: I'ld store the `Rng` in the `UrtsTipPool` type so we can reuse it. It will make this function require `&mut self` though. Please check if that could be an issue.

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,204 @@
...
+                    to_remove.push(*tip);
+                }
+            }
+        }
+        for tip in to_remove {
+            self.tips.remove(&tip);
+            self.non_lazy_tips.remove(&tip);
+        }
+    }
+}
```
Alex6323: That method can be improved a lot performance-wise.
(1) store the `to_remove` Vec with a reasonable capacity in the `UrtsTipPool` type, so we don't need to re-allocate whenever we run this method (but then don't forget to clean it at the end of the method :trollface: )
(2) make a comment that we should use the `retain` API once it's in stable Rust.

Alex6323: If you don't want to make this change atm, then maybe open an issue, and at least give the `Vec` a default capacity of the average tip-pool size :smile:

bee-tangle/tests/helper.rs
Alex6323: For me it's still there. Maybe it's on my end. Dunno.

Alex6323 (commented):
Just minor things. Nothing serious.

bee-tangle/Cargo.toml
zesterer: I think this is kind of bike-shedding, but I'll tweak the wording

bee-tangle/src/flags.rs
zesterer: Umm? I've not used anything but third-person here.

bee-tangle/src/metadata.rs
zesterer: I'm really not sure what you're talking about, you'll have to clarify.

bee-tangle/src/metadata.rs
zesterer: It's not clear to me: what does 'interesting' mean in this context? It feels like an insufficient definition without that being expanded upon.

bee-tangle/src/unconfirmed_message.rs
zesterer: Why? `repr(transparent)` has a very specific meaning that does not apply here.

bee-tangle/src/urts.rs
zesterer: I believe it is in the config. Perhaps this is the default? I'm not sure. I didn't write and, quite frankly, I'm grasping at straws when trying to understand it in the first place due to lack of comments and overuse of acronyms.

bee-tangle/src/flags.rs
Alex6323: I mean "Sets ..., Returns ..."

bee-tangle/src/urts.rs
zesterer: This aside, I have a list of things I think need to change about `Parents`... but that's for another day

bee-tangle/src/metadata.rs
Alex6323: `Creates....`

bee-tangle/src/urts.rs
zesterer: Going to refactor this whole section tbh, I'm not sure why it was written like this in the first place

bee-tangle/src/metadata.rs
Alex6323: "interesting" here means, that only two particular of the many "CRI"s are worth considering when it comes to tip selection. If you imagine a Cone with a particular milestone index at its root, then the only two interesting messages when you want to attach this cone to a previous one (in order to be accepted by the network) are the ones:
(a) the one message that is attached to the closest previous milestone (in terms of its index)
(b) the one message that is attached to the furthest previous milestone (in terms of its index)
Hence, only those two messages (within that cone!) are interesting.

bee-tangle/src/urts.rs
zesterer: I'm more than a little nervous about this. It sounds like it could be exploitable by an attacker to favour choosing certain tips. Regular reseeding is important. I'm not sure what possible purpose this would serve other than leaving things more open to exploitation than is otherwise necessary?

bee-tangle/src/unconfirmed_message.rs
Alex6323: Isn't the only reason we have this wrapper to satisfy the type system? I see no reason why this can't have the same memory layout as a MessageId.

bee-tangle/src/urts.rs
Alex6323: @rufsam Can you not just use the config setting here?

bee-tangle/src/urts.rs
zesterer: Before doing something like this, I'd like to get a few things straight:

1) Mutable state is bad, and it becomes even worse when persisted. Even in a seemingly innocent case like this, it's mental overhead that wouldn't otherwise be necessary for a maintainer to have to deal with. Something like 'don't forget to clean it at the end' is an implicit contract that can't be checked by the compiler, and so gets hoisted on to the developer tasked with maintaining the thing.

2) Modern allocators are fast. *Really fast*. Especially for small allocations like this. Unless you're telling me that `to_remove` grows to multiple thousands of elements in size every time it's invoked, this is going to be small fry compared to the other hot code we're running in the node. We shouldn't over-complicate things for small fry.

3) `retain` is already stable (as of Rust 1.0): https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#1398-1400

bee-tangle/src/urts.rs
Alex6323: you're right, Josh. But it seems like it's reseeding automatically:
https://docs.rs/rand/0.8.3/rand/rngs/struct.ThreadRng.html

bee-tangle/src/urts.rs
Alex6323: re: 3. Did I write `retain`? My bad. I meant `drain_filter`. 
https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter

Alex6323: re: 2. The current impl is still inefficient as it will reallocate the Vec several times. 6 times on average with current speeds. But this is going to increase in the future.

Alex6323: re: 1. In this case it's not that bad, because this function is the only place where this state is mutated.

Alex6323: Also. This code is part of the hot-path as the tip-set changes with every incoming message, hence the need to keep the tip-pool at bay, meaning it needs to be reduced constantly. This is exactly the type of code where you'ld want to reduce unnecessary allocations.

bee-tangle/src/flags.rs
zesterer: I'm not sure what you mean. "Returns [x]" is very common terminology in `std`: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_raw_parts

bee-tangle/src/metadata.rs
zesterer: Okay, but what do I put in the doc comment? I'm not an expert on this area, so you're going to be the better one to come up with a comment, unless you think the first one is sufficient?

bee-tangle/src/flags.rs
thibault-martinez: He wants the ending 's' in "returnS"

bee-tangle/src/unconfirmed_message.rs
zesterer: `repr(transparent)` is used when you want to *guarantee* that the representations are the same for a specific reason (i.e: you're unsafely transmuting between types). If the compiler chooses a different representation to the one you expect, it's because it knows something that you don't and we should let it choose that representation if it so desires. `repr(transparent)` just provides the compiler with fewer options to generate efficient code and should be avoided unless there's a specific reason for it.

bee-tangle/src/urts.rs
zesterer: In which case, I don't see there being a particular advantage to making this change.

bee-tangle/src/urts.rs
zesterer:
```rust
    pub(crate) fn reduce_tips(&mut self) {
        let non_lazy_tips = &mut self.non_lazy_tips;
        self.tips.retain(|tip, metadata| metadata
            .time_first_child
            .filter(|age| age.elapsed().as_secs() > MAX_AGE_SECONDS_AFTER_FIRST_CHILD as u64)
            .map(|_| non_lazy_tips.remove(&tip))
            .is_none());
    }
```

Here's a better implementation that avoids unnecessary mutable state

bee-tangle/src/metadata.rs
Alex6323: "A type that is used to associate two particular Cone Root Indexes with a message in the Tangle, i.e. the Oldest Cone Root Index (OCRI), and the Youngest Cone Root Index (YCRI) to decide on the eligibility of a message as a tip."

Alex6323: But since we are still using the old terms atm throughout the codebase, pls replace OCRI->OMRSI and YCRI->YMRSI for now. We'll change that later! :smile:

bee-tangle/src/metadata.rs
zesterer: I think this is bike-shedding and a very weird hill to die on. It's definitely not a rule we stick to anywhere else in Bee. It's not like `std` has consistency on this, even.

https://doc.rust-lang.org/std/fs/fn.read.html

"Read" vs "Reads"

bee-tangle/src/flags.rs
zesterer: See https://github.com/iotaledger/bee/pull/527#discussion_r618753923

bee-tangle/src/unconfirmed_message.rs
Alex6323: Alright, you're the expert on this, so I leave it up to you. But as someone who has worked with this type in the last couple of days quite a lot I know that this type is nothing but a `MessageId`, so maybe in the future it becomes important that we can treat is as such. That is what I suggested to tell the compiler. However, it's your choice.

bee-tangle/src/urts.rs
Alex6323: I does...but it's not equivalent code, because it ties the `retain` to the success of the removal in `non_lazy_tips`, which the former code didn't do.

Alex6323: oh no. it's fine :joy:

bee-tangle/src/urts.rs
Alex6323: alright

bee-tangle/tests/helper.rs
thibault-martinez: Remove this file

bee-tangle/tests/traversal.rs
```diff
@@ -0,0 +1,110 @@
+// Copyright 2020-2021 IOTA Stiftung
```
thibault-martinez: Remove this file

thibault-martinez (changes requested):

bee-ledger/src/types/conflict.rs
```diff
@@ -22,3 +27,43 @@ pub enum ConflictReason {
...
+            0 => Self::None,
+            1 => Self::InputUtxoAlreadySpent,
+            2 => Self::InputUtxoAlreadySpentInThisMilestone,
+            3 => Self::InputUtxoNotFound,
+            4 => Self::InputOutputSumMismatch,
+            5 => Self::InvalidSignature,
+            6 => Self::InvalidDustAllowance,
+            255 => Self::SemanticValidationFailed,
+            x => {
+                return Err(Error::new(
```
thibault-martinez: Can you add a new variant to the Error type in the crate instead please ?

bee-tangle/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+license = "Apache-2.0"
+keywords = ["iota", "tangle", "bee", "framework"]
+homepage = "https://www.iota.org"
+
+[dependencies]
+bee-common = { path = "../bee-common/bee-common" }
+bee-message = { path = "../bee-message", features = ["serde"] }
+bee-runtime = { path = "../bee-runtime" }
+bee-storage = { path = "../bee-storage/bee-storage" }
+bee-ledger = { path = "../bee-ledger", features = ["serde"] }
```
thibault-martinez: You'll need to put the versions to release

bee-tangle/src/event.rs
```diff
@@ -0,0 +1,45 @@
...
+/// An event that indicates that the latest milestone has changed.
+#[derive(Clone)]
+pub struct LatestMilestoneChanged {
+    /// The index of the milestone
+    pub index: MilestoneIndex,
+    /// The milestone data
+    pub milestone: Milestone,
+}
+
+/// An event that indicates that a solid milestone has changed.
```
thibault-martinez:
```suggestion
/// An event that indicates that the solid milestone has changed.
```

bee-tangle/src/event.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Clone)]
+pub struct SolidMilestoneChanged {
+    /// The index of the milestone
+    pub index: MilestoneIndex,
+    /// The milestone data
+    pub milestone: Milestone,
+}
+
+/// An event that indicates that a confirmed milestone has changed.
```
thibault-martinez:
```suggestion
/// An event that indicates that the confirmed milestone has changed.
```

bee-tangle/src/event.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Clone)]
+pub struct ConfirmedMilestoneChanged {
+    /// The index of the milestone
+    pub index: MilestoneIndex,
+    /// The milestone data
+    pub milestone: Milestone,
+}
+
+/// An event that indicates that a snapshot milestone has had an index change.
```
thibault-martinez:
```suggestion
/// An event that indicates that the snapshot milestone has changed.
```

bee-tangle/src/event.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Clone)]
+pub struct SnapshotMilestoneIndexChanged {
+    /// The index of the milestone
+    pub index: MilestoneIndex,
+}
+
+/// An event that indicates that a pruning milestone has had an index change.
```
thibault-martinez:
```suggestion
/// An event that indicates that the pruning milestone has changed.
```

bee-tangle/src/flags.rs
```diff
@@ -0,0 +1,81 @@
...
+use bee_common::packable::{Packable, Read, Write};
+
+use bitflags::bitflags;
+use serde::Serialize;
+
+bitflags! {
+    /// Flags representing the state of a message.
+    #[derive(Default, Serialize)]
+    pub struct Flags: u8 {
+        /// The message is solid (confirmed).
```
thibault-martinez:
```suggestion
        /// The message is solid.
```
solid != confirmed

bee-tangle/src/lib.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A crate that contains foundational building blocks for the IOTA Tangle.
+
+#![deny(missing_docs)]
+
+/// Types used for tangle configuration.
+pub mod config;
+/// Types that represent milestone events.
```
thibault-martinez:
```suggestion
/// Types that represent tangle events.
```
Will not necessarily always be related to milestones.

bee-tangle/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+pub mod event;
+/// Milestone message flags.
```
thibault-martinez:
```suggestion
/// Message flags.
```
These are not specific to milestones.

bee-tangle/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+pub mod flags;
+/// Milestone message data, including message flags.
```
thibault-martinez:
```suggestion
/// Message metadata, including message flags.
```
These are not specific to milestones.

bee-tangle/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+pub mod metadata;
+/// Milestone-enabled tangle type.
+pub mod ms_tangle;
+/// Types used to represent SEPs (Solid Entry Points).
+pub mod solid_entry_point;
+/// Types used for interoperation with a node's storage layer.
+pub mod storage;
+/// Common tangle traversal functionality.
+pub mod traversal;
+/// Types used to represent unconfirmed (unsolid) messages.
```
thibault-martinez:
```suggestion
/// Types used to represent unconfirmed messages.
```
solid != confirmed

thibault-martinez (changes requested):

bee-tangle/src/event.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/// An event that indicates that the snapshot milestone has changed.
+#[derive(Clone)]
+pub struct SnapshotMilestoneIndexChanged {
+    /// The index of the milestone
+    pub index: MilestoneIndex,
+}
+
+/// An event that indicates that a pruning milestone has changed.
```
thibault-martinez:
```suggestion
/// An event that indicates that the pruning milestone has changed.
```

bee-tangle/CHANGELOG.md
```diff
@@ -0,0 +1,26 @@
...
+
+## 0.1.0 - 2021-04-22
```
thibault-martinez: Change date to today

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    fn packed_len(&self) -> usize {
+        self.flags.packed_len()
+            + self.milestone_index.packed_len()
+            + self.arrival_timestamp.packed_len()
+            + self.solidification_timestamp.packed_len()
+            + self.reference_timestamp.packed_len()
+            + self.omrsi.packed_len()
+            + self.ymrsi.packed_len()
+            + 0u8.packed_len()
```
thibault-martinez:
```suggestion
            + self.conflict.packed_len()
```

bee-tangle/src/metadata.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    fn pack<W: Write>(&self, writer: &mut W) -> Result<(), Self::Error> {
+        self.flags.pack(writer)?;
+        self.milestone_index.pack(writer)?;
+        self.arrival_timestamp.pack(writer)?;
+        self.solidification_timestamp.pack(writer)?;
+        self.reference_timestamp.pack(writer)?;
+        self.omrsi.pack(writer)?;
+        self.ymrsi.pack(writer)?;
+        (self.conflict as u8).pack(writer)?;
```
thibault-martinez:
```suggestion
        self.conflict.pack(writer)?;
```

thibault-martinez (changes requested):

bee-tangle/src/unconfirmed_message.rs
thibault-martinez: Sorry for only noticing that now but could you rename this to `UnreferencedMessage` please ? It was renamed on some other places and it would be more consistent. Also the module name :)

thibault-martinez (changes requested):

bee-tangle/src/metadata.rs
zesterer: Ah, nice catch

bee-tangle/src/unconfirmed_message.rs
zesterer: Sure

bee-tangle/src/solid_entry_point.rs
thibault-martinez: Missing word ?

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,421 @@
...
+    type Error = B::Error;
+
+    async fn get(&self, msg: &MessageId) -> Result<Option<(Message, MessageMetadata)>, Self::Error> {
+        trace!("Attempted to fetch message {:?}", msg);
+        let message = self.storage.fetch(msg).await?;
+        let meta = self.storage.fetch(msg).await?;
+        Ok(message.zip(meta))
+    }
+
+    async fn insert(&self, msg: MessageId, tx: Message, metadata: MessageMetadata) -> Result<(), Self::Error> {
```
thibault-martinez:
```suggestion
    async fn insert(&self, msg_id: MessageId, msg: Message, metadata: MessageMetadata) -> Result<(), Self::Error> {
```

thibault-martinez: For this and all the others I'd use `msg_id` instead of `msg` when it's actually an id, otherwise it's a bit confusing. I'll let you decide if you want to change but at least change `tx` here as it's wrong.

bee-tangle/src/solid_entry_point.rs
zesterer: What word is missing?

bee-tangle/src/ms_tangle.rs
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    /// Get the configuration of this tangle.
+    pub fn config(&self) -> &TangleConfig {
+        &self.config
+    }
+
+    /// Insert a message into the tangle.
+    pub async fn insert(&self, message: Message, hash: MessageId, metadata: MessageMetadata) -> Option<MessageRef> {
+        // TODO this has been temporarily moved to the processor.
```
thibault-martinez: Remove all this comment, very old code.

bee-tangle/src/solid_entry_point.rs
```diff
@@ -0,0 +1,80 @@
...
+//! without needing the full tangle history.
+
+use bee_common::packable::{Packable, Read, Write};
+use bee_message::MessageId;
+
+use ref_cast::RefCast;
+
+use core::{convert::AsRef, ops::Deref};
+
+/// A SolidEntryPoint is a [`MessageId`] of a message even if we do not have them
```
thibault-martinez: Missing word ?

bee-tangle/src/solid_entry_point.rs
```diff
@@ -0,0 +1,80 @@
...
+/// or their past in the database. They often come from a snapshot file and allow a node to solidify
+/// without needing the full tangle history.
+///
+/// This is a type wrapper around a [`MessageId`](crate::MessageId) to differentiate it from a
+/// non-solidified message.
+///
+/// Spec: #iota-protocol-rfc
+/// <https://github.com/iotaledger/protocol-rfcs/blob/master/text/0005-white-flag/0005-white-flag.md>
```
thibault-martinez: Remove these two lines (I know it was already there).

bee-tangle/src/solid_entry_point.rs
```diff
@@ -0,0 +1,80 @@
...
+use ref_cast::RefCast;
+
+use core::{convert::AsRef, ops::Deref};
+
+/// A SolidEntryPoint is a [`MessageId`] of a message even if we do not have them
+/// or their past in the database. They often come from a snapshot file and allow a node to solidify
+/// without needing the full tangle history.
+///
+/// This is a type wrapper around a [`MessageId`](crate::MessageId) to differentiate it from a
+/// non-solidified message.
```
thibault-martinez: That's not a correct sentence. Just remove it.

bee-tangle/src/solid_entry_point.rs
```diff
@@ -0,0 +1,80 @@
...
+///
+/// Spec: #iota-protocol-rfc
+/// <https://github.com/iotaledger/protocol-rfcs/blob/master/text/0005-white-flag/0005-white-flag.md>
+#[derive(RefCast)]
+#[repr(transparent)]
```
thibault-martinez, line 19: I know there was a discussion about this, should it stay then ?

bee-tangle/src/tangle.rs
```diff
@@ -0,0 +1,468 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{vertex::Vertex, MessageRef};
+
+use bee_message::{Message, MessageId};
+
+use async_trait::async_trait;
+// use dashmap::{mapref::entry::Entry, DashMap};
```
thibault-martinez: Remove

bee-tangle/src/tangle.rs
```diff
@@ -0,0 +1,468 @@
...
+        Ok(())
+    }
+}
+
+/// A foundational, thread-safe graph datastructure to represent the IOTA Tangle.
+pub struct Tangle<T, H = NullHooks<T>>
+where
+    T: Clone,
+{
+    // Global Tangle Lock. Remove this as and when it is deemed correct to do so.
```
thibault-martinez: Are we still planning on removing it ?

bee-tangle/src/tangle.rs
```diff
@@ -0,0 +1,468 @@
...
+                    true
+                } else {
+                    false
+                }
+            })
+        } else {
+            self.contains_inner(message_id).await
+        };
+
+        // If the tangle already contains the tx, do no more work
```
thibault-martinez:
```suggestion
        // If the tangle already contains the message, do no more work
```

thibault-martinez: Maybe do a pass and replace tx with msg please, this is wrong

bee-tangle/src/urts.rs
```diff
@@ -0,0 +1,197 @@
...
+                Score::Lazy
+            } else if smi > omrsi + OMRSI_DELTA {
+                Score::SemiLazy
+            } else {
+                Score::NonLazy
+            }
+        }
+    }
+
+    pub fn two_non_lazy_tips(&self) -> Option<Vec<MessageId>> {
```
thibault-martinez:
```suggestion
    pub fn non_lazy_tips(&self) -> Option<Vec<MessageId>> {
```
It doesn't return only two anymore

thibault-martinez (changes requested):

bee-tangle/src/solid_entry_point.rs
thibault-martinez: Sentence was weird to me, I'd better understand something like `A SolidEntryPoint is a [`MessageId`](crate::MessageId) of a message that is solid even if we do not have them...`

bee-tangle/src/solid_entry_point.rs
zesterer: It's needed for `RefCast`

thibault-martinez: Thanks

thibault-martinez (approved):

Adam-Gleave (approved):

Alex6323 (approved):

pvdrz (approved):
I think this needs more memes but we can leave that for coordicide.

PR #569, opened 2021-05-04
Add THE `Packable` trait

### LINKED ISSUE #533: Rewrite the `Packable` trait to make it `no_std` compatible.
## Description

Rewrite the `Packable` trait to make it `no_std` compatible.

## Motivation

Right now the `Packable` trait inside `bee-common` depends explicitly on several types from `std::io` whose migration to `core` seems unlikely. 

## Requirements

Write a list of what you want this feature to do.

1. Be able to use `Packable` in a `no_std` crate.
2. Still be able to work with `std::io` if available.
3. Be able to `#[derive(Packable)]` if no additional checks are needed to pack a type. 

## Open questions (optional)

- It seems there is no standard alternative to `std::io` (no pun intended). There is [genio](https://crates.io/crates/genio) but its development stopped and it is being redesigned. Do we want to have our own general purpose `io` library as part of `bee-common`?

- The naive way to get out would be to introduce two new traits `Read` and `Write` and add either an associated type or generic type parameter to `Packable`, `Read` and `Write` for the error:
  -  Using an associated type restricts us to having to provide a single error type for every implementation beforehand. Which would mean we would have to also provide our own readers and writers. All of this conflicts heavily with (2).
  -  On the other hand using a generic type does not have any of those problems but sacrifices usability because you cannot do `true.packed_len()` but have to write `Packable::<E>::packed_len(&true)` instead. It is also more fragile because an user could provide different `packed_len`s for different error types. We could move `packed_len` to another trait `PackedSized` with no generics and make `Packable<E>: PackedSized` as a workaround.

- Another option would be to move away from the fallible IO approach and make all the `Packable` functions infallible. It would be the "packer's" responsibility to provide a `Vec` or slice with enough space and with the appropriate lifetime to pack and unpack the desired value.  
- Is it feasible to provide our own `serde` serializer and deserializer given that it is already `no_std` and remove `Packable` completely?

Yes

pvdrz:
Fixes #533.


thibault-martinez (changes requested):
Can you remove the bee-tmp crate that should not be required anymore please ?

bee-common/src/lib.rs
```diff
@@ -1,4 +1,4 @@
 // Copyright 2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
```
thibault-martinez: Please add `#![warn(missing_docs)]` so we're not missing any public doc

bee-common/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "bee-common"
+version = "0.1.0"
```
thibault-martinez:
```suggestion
version = "1.0.0-alpha"
```

bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+
+macro_rules! impl_packable_for_int {
+    ($ty:ty) => {
+        impl Packable for $ty {
+            fn pack<P: Packer>(&self, packer: &mut P) -> Result<(), P::Error> {
+                packer.pack_bytes(&self.to_le_bytes())
+            }
+
+            fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, U::Error> {
+                let bytes: &[u8; std::mem::size_of::<Self>()] = unpacker.unpack_exact_bytes()?;
```
thibault-martinez:
```suggestion
                let bytes: &[u8; core::mem::size_of::<Self>()] = unpacker.unpack_exact_bytes()?;
```

bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn pack<P: Packer>(&self, packer: &mut P) -> Result<(), P::Error> {
+        for item in self.iter() {
+            item.pack(packer)?;
+        }
+
+        Ok(())
+    }
+
+    fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, U::Error> {
+        use std::mem::MaybeUninit;
```
thibault-martinez:
```suggestion
        use core::mem::MaybeUninit;
```

thibault-martinez (changes requested):

bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+    };
+}
+
+impl_packable_for_int!(u8);
+impl_packable_for_int!(u16);
+impl_packable_for_int!(u32);
+impl_packable_for_int!(u64);
+impl_packable_for_int!(u128);
```
thibault-martinez: Can you conditionally exclude if u128 is not available please ? We had issues in the past because of that.
https://github.com/iotaledger/bee/blob/dev/bee-common/bee-common/src/packable.rs#L204
https://github.com/iotaledger/bee/blob/dev/bee-common/bee-common/build.rs

bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, U::Error> {
+        Ok(u64::unpack(unpacker)? as usize)
+    }
+}
+
+impl_packable_for_int!(i8);
+impl_packable_for_int!(i16);
+impl_packable_for_int!(i32);
+impl_packable_for_int!(i64);
+impl_packable_for_int!(i128);
```
thibault-martinez: Same

thibault-martinez (changes requested):

Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
 members = [
-  "bee-tmp"
+  "bee-common",
+  "bee-common-derive"
```
thibault-martinez: I wonder if we should bundle them together like we did in the dev branch. Just thinking that it can get a bit too much in the future if we have other derive crates.

bee-common-derive/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+authors = ["IOTA Stiftung"]
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = { version = "1.0", features = ["full", "extra-traits"] }
+quote = "1.0"
+proc-macro2 = "1.0"
```
thibault-martinez: Please order them alphanumerically :)

bee-common-derive/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "bee-common-derive"
+version = "1.0.0-alpha"
+authors = ["IOTA Stiftung"]
+edition = "2018"
```
thibault-martinez: description field ?

bee-common/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A crate that provides common functionalities shared across multiple crates within the Bee framework, and for
+//! applications built on-top.
```
thibault-martinez:
```suggestion
//! applications built on-top.

```

bee-common/build.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright 2020-2021 IOTA Stiftung
```
thibault-martinez:
```suggestion
// Copyright 2021 IOTA Stiftung
```

bee-common/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "bee-common"
+version = "1.0.0-alpha"
+authors = ["IOTA Stiftung"]
+edition = "2018"
```
thibault-martinez: description field ?

thibault-martinez (changes requested):

Cargo.toml
pvdrz: whatever you think is better is fine for me

bee-common-derive/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 IOTA Stiftung
+// Copyright 2020-2021 IOTA Stiftung
```
thibault-martinez: I am sorry, maybe one of my comments was misinterpreted, but it should be 2021 only :(

pvdrz: [`sd`](https://github.com/chmln/sd) is my best friend
```bash
sd '2020-2021' '2020' $(fd --type file)
```

thibault-martinez: Well... you replaced with 2020 :trollface:

bee-common-derive/src/packable.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    match struct_fields {
+        Fields::Named(FieldsNamed { named, .. }) => {
+            let (labels, types): (Vec<&Ident>, Vec<&Type>) = named
+                .iter()
+                .map(|field| (field.ident.as_ref().unwrap(), &field.ty))
+                .unzip();
+
+            pack = quote! {
+                #(self.#labels.pack(packer)?;) *
```
zesterer: Perhaps use the fully-qualified UFCS to avoid possible ambiguity errors?

bee-common-derive/src/lib.rs
pvdrz:
```bash
sd '2020' '2021' $(fd --type file)
```

bee-common-derive/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+mod packable;
+
+#[proc_macro_derive(Packable, attributes(packable))]
+pub fn packable(input: TokenStream) -> TokenStream {
+    let DeriveInput { ident, data, attrs, .. } = parse_macro_input!(input);
+
+    let (pack, unpack) = match data {
+        Data::Struct(data_struct) => packable::gen_struct_bodies(data_struct.fields),
+        Data::Enum(data_enum) => {
+            let ty = packable::parse_attr::<Type>("ty", &attrs).unwrap();
```
Alex6323: should we add a comment about this 'unwrap'?

bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+            item.pack(packer)?;
+        }
+
+        Ok(())
+    }
+
+    fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, U::Error> {
+        use core::mem::MaybeUninit;
+
+        // Safety: an uninitialized array of unitialized stuff is actually initialized.
```
Alex6323: Haha. That explanation was confusing to me at first, maybe to others as well, I don't know. Maybe an easier to understand version of it would be: "an uninitialized array of `MaybeUninit`s (don't require initialization) is safe to be considered initialized" ?

bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+        let mut array = unsafe { MaybeUninit::<[MaybeUninit<T>; N]>::uninit().assume_init() };
+
+        for item in array.iter_mut() {
+            let unpacked = T::unpack(unpacker)?;
+            // Safety: each `item` is only visited once so we are never overwritting values that
```
Alex6323: typo: overwri_t_ing.

bee-common/tests/packable.rs
```diff
@@ -0,0 +1,141 @@
...
+fn packed_non_zero_bytes_are_truthy() {
+    let mut packer = VecPacker::default();
+    42u8.pack(&mut packer).unwrap();
+
+    let bytes = packer.vec.leak();
+
+    let mut unpacker = SliceUnpacker::new(bytes);
+    let result = bool::unpack(&mut unpacker).unwrap();
+
+    assert!(result);
```
Alex6323: maybe instead of `result` this variable is better named `is_true`?

bee-common/src/packable/mod.rs
pvdrz: I like that

bee-common-derive/src/lib.rs
pvdrz: yes maybe I should add better error reporting in general.

bee-common-derive/tests/packable/dup_id_enum.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(unused_imports)]
```
thibault-martinez:
```suggestion
#![allow(unused_imports)]

```
For this one and all the others :p

Cargo.toml
thibault-martinez: I'd rather bundle them yes

bee-common/src/packable/unpacker.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
```
thibault-martinez:
```suggestion
// SPDX-License-Identifier: Apache-2.0

```

Cargo.toml
Alex6323: agree on bundling them

bee-common/bee-common-derive/tests/packable/fail/dup_id_enum.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#![allow(unused_imports)]
+
+use bee_common::packable::Packable;
+use bee_common_derive::Packable;
+
+#[derive(Packable)]
+#[packable(ty = u8)]
+pub enum OptI32 {
+    #[packable(id = 0)]
```
zesterer: `id` seems like a strange name... Perhaps `tag`?

bee-common/bee-common-derive/tests/packable/fail/invalid_ty_enum.stderr
```diff
@@ -0,0 +1,33 @@
+error[E0605]: non-primitive cast: `i32` as `[u8; 32]`
```
zesterer: This test seems like it's going to fail frequently, especially because I hear the 'mismatched types' error is due to change quite soon.

bee-common/bee-common/src/packable/error.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use core::fmt::Display;
+
+use super::Packable;
+
+/// A type that represents errors with the unpacking format as well as with the unpacking process itself.
+pub trait UnpackError: Sized {
```
zesterer: Should this have a `Error` bound too (or, at least, `Debug`)?

bee-common/bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+            let unpacked = T::unpack(unpacker)?;
+            // Safety: each `item` is only visited once so we are never overwriting nor dropping
+            // values that are already initialized.
+            unsafe {
+                item.as_mut_ptr().write(unpacked);
+            }
+        }
+
+        // Safety: We traversed the whole array and initialized every item.
+        Ok(unsafe { (&array as *const [MaybeUninit<T>; N] as *const Self).read() })
```
zesterer, line 212: A `// TODO` to switch to https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.array_assume_init

bee-common/bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, U::Error> {
+        use core::mem::MaybeUninit;
+
+        // Safety: an uninitialized array of `MaybeUninit`s is safe to be considered initialized.
+        let mut array = unsafe { MaybeUninit::<[MaybeUninit<T>; N]>::uninit().assume_init() };
```
zesterer, line 198: A `// TODO` to switch to https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.uninit_array

bee-common/bee-common/src/packable/unpacker.rs
```diff
@@ -0,0 +1,20 @@
...
+/// A type that can unpack any value that implements `Packer`.
+pub trait Unpacker {
+    /// The error type that can be returned if some error occurs while unpacking.
+    type Error: UnpackError;
+
+    /// Unpack a sequence of bytes from the `Unpacker`.
+    fn unpack_bytes(&mut self, slice: &mut [u8]) -> Result<(), Self::Error>;
+
+    /// Unpack a statically-sized sequence of bytes from the `Unpacker`.
+    fn unpack_exact_bytes<const N: usize>(&mut self) -> Result<[u8; N], Self::Error> {
```
zesterer: This function is presumably only used for numeric types, and those are implemented with a macro. Why not put this impl in the macro? I can't think of anything else for which it would be useful.

bee-common/bee-common-derive/src/packable.rs
```diff
@@ -0,0 +1,250 @@
...
+
+            unpack = quote! {
+                Ok(Self(#(<#types>::unpack(unpacker)?), *))
+            };
+
+            packed_len = quote! {
+                0 #(+ <#types>::packed_len(&self.#indices)) *
+            }
+        }
+        Fields::Unit => {
```
zesterer: There's something weirdly annoying to me that `struct MyStruct();` doesn't count as a unit struct but `struct MyStruct;` does.

zesterer (dismissed):

bee-common/bee-common-derive/tests/packable/fail/invalid_ty_enum.stderr
pvdrz: I'd actually like to have a custom error for this saying that you can only use integers or smth.

zesterer (dismissed):

bee-common/bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+pub enum UnpackOptionError<T: Packable> {
+    /// The tag found while unpacking is not valid.
+    UnknownTag(u8),
+    /// A semantic error for the underlying type was raised.
+    Inner(T::Error),
+}
+
+/// Options are packed and unpacked using `0u8` as the prefix for `None` and `1u8` as the prefix
+/// for `Some`.
+impl<T: Packable> Packable for Option<T> {
```
zesterer, line 147: Does deriving not work here?

pvdrz: can you derive for types that are defined in `core`/`std`?

zesterer (dismissed):

bee-common/bee-common/src/packable/mod.rs
zesterer: Fair point, I'm not using my brain today.

thibault-martinez (changes requested):
Tests are failing

bee-common/bee-common-derive/tests/lib.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::path::Path;
+
+use trybuild::TestCases;
```
thibault-martinez, line 4: 3rd parties before `std`

bee-common/bee-common-derive/tests/packable/fail/duplicated_tag_enum.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#![allow(unused_imports)]
+
+use bee_common::packable::Packable;
+
+use core::convert::Infallible;
+
+#[derive(Packable)]
+#[packable(tag_ty = u8)]
+#[packable(error = Infallible)]
```
thibault-martinez, line 12: Wherever possible, I'd put attributes together like `#[packable(tag_ty = u8, error = Infallible)]`

bee-common/bee-common-derive/tests/packable/fail/duplicated_tag_enum.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(unused_imports)]
```
thibault-martinez, line 4: What is not being used ?

bee-common/bee-common-derive/tests/packable/fail/invalid_tag_enum.stderr
```diff
@@ -0,0 +1,5 @@
+error: All variants of an enum that derives `Packable` require a `#[packable(tag = ...)]` attribute.
```
thibault-martinez: This error doesn't feel right ? All variants actually have a `#[packable(tag = ...)]` ?

bee-common/bee-common-derive/tests/packable/fail/invalid_ty_enum.stderr
```diff
@@ -0,0 +1,5 @@
+error: Tags for enums can only be unisigned, sized integers.
```
thibault-martinez: Typo `unisigned `

bee-common/bee-common-derive/tests/packable/fail/missing_tag_enum.stderr
```diff
@@ -0,0 +1,5 @@
+error: All variants of an enum that derives `Packable` require a `#[packable(tag = ...)]` attribute.
```
thibault-martinez, line 1:
```suggestion
error: All variants of an enum that derive `Packable` require a `#[packable(tag = ...)]` attribute.
```

thibault-martinez (changes requested):

bee-common/bee-common-derive/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+proc-macro = true
+
+[[test]]
+name = "tests"
+path = "tests/lib.rs"
+
+[package.metadata.cargo-udeps.ignore]
+development = ["bee-common"]
+
+[dev-dependencies]
```
thibault-martinez: Can we put `dev-dependencies` after `dependencies` please ?

bee-common/bee-common/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = ["iota", "tangle", "bee", "framework", "common"]
+homepage = "https://www.iota.org"
+
+[features]
+alloc = []
+io = []
+
+[dev-dependencies]
```
thibault-martinez: same

bee-common/bee-common/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+bee-common = { path = ".", features = ["alloc", "io"] }
```
thibault-martinez: Is that actually necessary ?

thibault-martinez (changes requested):

bee-common/bee-common-derive/src/lib.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use proc_macro::{self, TokenStream};
+use proc_macro_error::{abort, proc_macro_error};
+use quote::{quote, ToTokens};
+use syn::{parse_macro_input, Data, DeriveInput, Type};
+
+mod packable;
```
thibault-martinez: Alright I'm being too picky now, but for consistency with the rest of the codebase, can we put `mod`s before `use`s please ?

bee-common/bee-common-derive/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+#[proc_macro_derive(Packable, attributes(packable))]
+pub fn packable(input: TokenStream) -> TokenStream {
+    let DeriveInput {
+        ident,
+        data,
+        attrs,
+        generics,
+        ..
+    } = parse_macro_input!(input);
+    // Check if the type has an `error` attribute.
```
thibault-martinez: Doesn't really "check" it though

bee-common/bee-common-derive/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    let error_type = packable::parse_attr::<Type>("error", &attrs).map(ToTokens::into_token_stream);
+
+    match data {
+        Data::Struct(data) => {
+            // Use `Infallible` if there was no error attribute.
+            let error_type = error_type.unwrap_or_else(|| quote!(core::convert::Infallible));
+            // Generate the implementation for the struct.
+            let (pack, packed_len, unpack) = packable::gen_bodies_for_struct(data.fields);
+            packable::gen_impl(&ident, &generics, error_type, pack, unpack, packed_len).into()
```
thibault-martinez: OCD triggered: `pack, packed_len, unpack` not in the same order as passed to `gen_impl`. Preferred order: `packed_len, pack, unpack`.

bee-common/bee-common-derive/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        }
+        Data::Enum(data) => {
+            // Verify that the enum has a `"tag_ty"` attribute for the type of the tag.
+            let tag_ty = packable::parse_attr::<Type>("tag_ty", &attrs).unwrap_or_else(|| {
```
thibault-martinez: I'm not a big fan of `tag_ty` because I like expressive attributes, shortening make it harder to read IMO, as a comparison serde doesn't shorten anything, I'd prefer the full `tag_type` tbh. Another thing is that `type` is a keyword so we took the habit of calling these tag bytes `kind`, not sure if we want to also call it that here, we should discuss it.

thibault-martinez (changes requested):

bee-common/bee-common/src/packable/packer.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+/// A type that can pack any value that implements `Packer`.
```
thibault-martinez: I don't understand this sentence

thibault-martinez: Replace `Packer` by `it` ?

bee-common/bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,209 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A module that provides a `Packable` trait to serialize and deserialize types.
+
+#[cfg(feature = "alloc")]
+mod alloc;
```
thibault-martinez: Why is this under a feature btw ? no_std libraries can depend on alloc

bee-common/bee-common/src/packable/io.rs
```diff
@@ -0,0 +1,24 @@
...
+extern crate std;
+
+use super::{Packer, Unpacker};
+
+use std::io::{self, Read, Write};
+
+impl<W: Write> Packer for W {
+    type Error = io::Error;
+
+    fn pack_bytes(&mut self, bytes: &[u8]) -> Result<(), Self::Error> {
```
thibault-martinez: I wonder if we'll ever have the need for async pack/unpack, like async versions of Read/Write 🤔

thibault-martinez (changes requested):

bee-common/bee-common/src/packable/packer.rs
pvdrz: It should be `Packable`

bee-common/bee-common/src/packable/mod.rs
pvdrz: is `alloc` always available when you use `no_std`?

bee-common/bee-common/src/packable/io.rs
pvdrz: image

bee-common/bee-common-derive/tests/packable/fail/duplicated_tag_enum.rs
pvdrz: `Infallible` is not used but i don't want the stderr file to include stuff not related to the actual error.

bee-common/bee-common/src/packable/mod.rs
thibault-martinez: From the doc `While #![no_std] binaries using alloc still require nightly Rust, #![no_std] library crates can use the alloc crate in stable Rust.` https://doc.rust-lang.org/edition-guide/rust-next/alloc.html

bee-common/bee-common-derive/tests/packable/fail/invalid_tag_enum.stderr
pvdrz: so the thing is that I'm parsing `#[packable(tag = <INTEGER>)]`. Maybe I should parse `#[packable(tag = <EXPR>)]` if the first one fails and return an error.

bee-common/bee-common-derive/tests/packable/fail/duplicated_tag_enum.rs
thibault-martinez: But it's used line 12, because it's part of a macro it complains ?

bee-common/bee-common-derive/tests/packable/fail/missing_tag_enum.stderr
pvdrz: but the enum is the one deriving packable. Maybe my grammar is wrong?

bee-common/bee-common/Cargo.toml
pvdrz: the tests require it :/

bee-common/bee-common-derive/src/lib.rs
pvdrz: there is no rest of the codebase :trollface: (JK)

bee-common/bee-common-derive/src/lib.rs
pvdrz: I'll use `tag_type` for now. The thing with kind is that you don't know what you should put there.

bee-common/bee-common/Cargo.toml
thibault-martinez: Ah maybe that's because of the features ? Because the crates is automatically added for tests of the same crate

pvdrz:
> Tests are failing

I think it is because one of the error messages changed in nightly but not in stable.

bee-common/bee-common/src/packable/mod.rs
pvdrz: great :D

bee-common/bee-common-derive/tests/packable/fail/duplicated_tag_enum.rs
pvdrz: yes because macro expansion doesn't happen because of the error.

thibault-martinez: Ahhhhh I see, thanks

bee-common/bee-common-derive/tests/packable/fail/duplicated_tag_enum.rs
pvdrz: hmmm... This comment made me realize something. 

Most libraries would do `#[packable(tag_ty = "u8", error = "Infallible")]` because the quotes give a clear delimitation of each argument. But we avoid the quotes at all, which means we don't know when an argument ends because they can have arbitrary syntax. 

So if we have  `#[packable(tag_ty = u8, other_arg = (foo, trait T), error = Infallible)]`, how do we parse that? 
- If we were to skip tokens until we reach a `,` we will end up at `trait T`.
- If we try to parse `(foo, trait T)` we would fail because it is not a valid term in Rust.

Maybe we should just use quotes for arguments.

bee-common/bee-common-derive/tests/packable/pass/error_coercion.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+impl From<PickyError> for PickyAndByteError {
+    fn from(err: PickyError) -> Self {
+        Self(err)
+    }
+}
+
+
```
Thoralf-M: Why two empty lines here?

bee-common/bee-common-derive/tests/packable/pass/unit_struct.rs
```diff
@@ -0,0 +1,19 @@
...
+#[derive(Packable)]
+#[packable(error = Infallible)]
+pub struct Unit;
+
+#[derive(Packable)]
+#[packable(error = Infallible)]
+pub struct RoundUnit();
+
+fn main() {}
+
```
Thoralf-M: Only file with a newline at the end, align!

bee-common/bee-common-derive/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+proc-macro-error = "1.0"
+proc-macro2 = "1.0"
+quote = "1.0"
+syn = { version = "1.0", features = ["full", "extra-traits"] }
+
+[dev-dependencies]
+bee-common = { path = "../bee-common" }
+
+trybuild = { version = "1.0", features = ["diff"] }
+rustversion = { version = "1.0" }
```
thibault-martinez, line 32: R before T :p

bee-common/bee-common-derive/tests/packable/fail/missing_tag_type_enum.rs
```diff
@@ -0,0 +1,19 @@
...
+#![allow(unused_imports)]
+
+use bee_common::packable::Packable;
+
+use core::convert::Infallible;
+
+#[derive(Packable)]
+#[packable(error = Infallible)]
+pub enum OptI32 {
+    #[packable(id = 0)]
```
thibault-martinez: Irrelevant to the test but why is this `id` here ?

bee-common/bee-common/tests/lib.rs
```diff
@@ -1,4 +1,4 @@
 // Copyright 2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-fn main() {}
+mod packable;
```
thibault-martinez: Is this file needed ?

bee-common/bee-common-derive/src/packable.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// The returned streams correspond to the branches necessary to implement `pack`, `packed_len`
+    /// and `unpack` for this variant.
+    fn consume_for_variant(self, tag: impl ToTokens, tag_ty: impl ToTokens) -> (TokenStream, TokenStream, TokenStream) {
+        let Self {
+            pattern,
+            pack,
+            packed_len,
+            unpack,
+        } = self;
+        // Following the example from `new` and assuming that the tag for this variant is `tag` and
```
Adam-Gleave: I would put a newline here for consistency.

bee-common/bee-common/tests/lib.rs
pvdrz: yes, if we want more stuff in bee-common

bee-common/bee-common-derive/tests/packable/fail/missing_tag_type_enum.rs
pvdrz: cuz someone forgot to change it :P let me fix it

bee-common/bee-common/tests/packable.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
```
thibault-martinez, line 2:
```suggestion
// SPDX-License-Identifier: Apache-2.0

```

bee-common/bee-common/src/packable/io.rs
thibault-martinez: Should this be `AsRef<[u8]>` ?

bee-common/bee-common/src/packable/io.rs
```diff
@@ -0,0 +1,24 @@
...
+        self.write_all(bytes)
+    }
+}
+
+impl<R: Read> Unpacker for R {
+    type Error = io::Error;
+
+    fn unpack_bytes(&mut self, slice: &mut [u8]) -> Result<(), Self::Error> {
```
thibault-martinez: `AsMut<[u8]>` ?

bee-common/bee-common/tests/lib.rs
thibault-martinez: I don't follow, everything at the root of `tests` is executed with or without lib.rs

bee-common/bee-common/src/packable/io.rs
pvdrz: well it could yes.

bee-common/bee-common/src/packable/packer.rs
```diff
@@ -0,0 +1,47 @@
...
+pub trait Packer {
+    /// The error type that can be returned if some error occurs while packing.
+    type Error;
+
+    /// Pack a sequence of bytes into the `Packer`.
+    fn pack_bytes(&mut self, bytes: &[u8]) -> Result<(), Self::Error>;
+}
+
+#[derive(Default)]
+/// A `Packer` backed by a `Vec<u8>`.
```
thibault-martinez, line 20: Not consistent with the rest of your code, put the comment before the attribute

bee-common/bee-common/src/packable/packer.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
```
thibault-martinez, line 2:
```suggestion
// SPDX-License-Identifier: Apache-2.0

```

bee-common/bee-common/src/packable/packer.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// Pack a sequence of bytes into the `Packer`.
+    fn pack_bytes(&mut self, bytes: &[u8]) -> Result<(), Self::Error>;
+}
+
+#[derive(Default)]
+/// A `Packer` backed by a `Vec<u8>`.
+pub struct VecPacker(Vec<u8>);
+
+impl VecPacker {
```
thibault-martinez, line 24: Add a `new` method pls

bee-common/bee-common/tests/lib.rs
pvdrz: oh wow, then yes we don't need it :P

bee-common/bee-common-derive/src/packable.rs
```diff
@@ -0,0 +1,377 @@
...
+
+            fn pack<P: bee_common::packable::Packer>(&self, packer: &mut P) -> Result<(), P::Error> {
+                #pack_body
+            }
+
+            fn unpack<U: bee_common::packable::Unpacker>(unpacker: &mut U) -> Result<Self, bee_common::packable::UnpackError<Self::Error, U::Error>> {
+                #unpack_body
+            }
+
+            fn packed_len(&self) -> usize {
```
thibault-martinez, line 318: For consistency with the rest of your code, put packed_len after pack

bee-common/bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+/// A type that can be packed and unpacked.
+///
+/// Almost all basic sized types implement this trait. This trait can be derived using the
+/// `bee_common_derive::Packable` macro. If you need to implement this trait manually, use the provided
+/// implementations as a guide.
+pub trait Packable: Sized {
+    /// The error type that can be returned if some semantic error occurs while unpacking.
+    ///
+    /// It is recommended to use `core::convert::Infallible` if this kind of error cannot happen or
+    /// `UnpackEnumError` when implementing this trait for an enum.
```
thibault-martinez: UnknownTagError ?

bee-common/bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    type Error;
+    /// Pack this value into the given `Packer`.
+    fn pack<P: Packer>(&self, packer: &mut P) -> Result<(), P::Error>;
+    /// Unpack this value from the given `Unpacker`.
+    fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, UnpackError<Self::Error, U::Error>>;
+    /// The size of the value in bytes after being packed.
+    fn packed_len(&self) -> usize;
```
thibault-martinez, line 37: Can you put all packed_len right after pack ?

bee-common/bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+        impl Packable for $ty {
+            type Error = Infallible;
+
+            fn pack<P: Packer>(&self, packer: &mut P) -> Result<(), P::Error> {
+                packer.pack_bytes(&self.to_le_bytes())
+            }
+
+            fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, UnpackError<Self::Error, U::Error>> {
+                let mut bytes = [0u8; core::mem::size_of::<Self>()];
+                unpacker.unpack_bytes(&mut bytes).map_err(UnpackError::Unpacker)?;
```
thibault-martinez: Do we really need the map ?

bee-common/bee-common/src/packable/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    type Error = Infallible;
+
+    /// Booleans are packed as `u8` integers following Rust's data layout.
+    fn pack<P: Packer>(&self, packer: &mut P) -> Result<(), P::Error> {
+        (*self as u8).pack(packer)
+    }
+
+    /// Booleans are unpacked if the byte used to represent them is non-zero.
+    fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, UnpackError<Self::Error, U::Error>> {
+        Ok(unpacker.unpack_infallible::<u8>().map_err(UnpackError::Unpacker)? != 0)
```
thibault-martinez: Shouldn't `unpack_infallible ` be a method of packable instead ?
Also, what's the point of calling `unpack_infallible` since we still have to map it to the same type as if we used unpack_bytes ?

bee-common/bee-common-derive/tests/packable/pass/error_coercion.rs
```diff
@@ -0,0 +1,90 @@
...
+        let value = unpacker.unpack_infallible::<u8>()?;
+
+        if value == 42 {
+            Ok(Self(value))
+        } else {
+            Err(UnpackError::Packable(PickyError(value)))
+        }
+    }
+
+    fn packed_len(&self) -> usize {
```
thibault-martinez: before unpack

bee-common/bee-common/src/packable/unpacker.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use super::{Packable, UnpackError};
+
+use core::convert::Infallible;
+
+/// A type that can unpack any value that implements `Packer`.
```
thibault-martinez: Packable ?

bee-common/bee-common/src/packable/mod.rs
pvdrz: > Shouldn't `unpack_infallible ` be a method of packable instead ?

Yes but I couldn't figure out how to do that because it should only work for `Packable<Error = Infallible>` and I didn't want to add another trait just for that.
> Also, what's the point of calling `unpack_infallible` since we still have to map it to the same type as if we used unpack_bytes ?

That you don't need to map the `Packable::Error` error too because it doesn't happen. Otherwise you'd have to do something like
```rust
match unpack_error {
    UnpackError::Unpacker(err) => UnpackError::Unpacker(err),
    UnpackError::Packable(infallible) => match infallible {},
}
```
an alternative could be adding special methods for `UnpackError` if the packable error is `Infallible`.

bee-common/bee-common/src/packable/mod.rs
pvdrz: we do not!

thibault-martinez: Maybe do a pass and see if we can remove some other ?

thibault-martinez (approved):

Thoralf-M (approved):

Adam-Gleave (approved):

PR #571, opened 2021-05-06
Allow Docker image to build

cscetbon:
# Description of change

Move the Dockerfile to the root level allowing Docker to finds all dependencies it needs and successfully build the Docker image.

## Type of change

- Bug fix

Ran the command `docker build -t bee .` and added it to the README as well for all users.


Alex6323:
@cscetbon Hey thanks for the PR, and sorry for the late response, but we're kinda busy with a lot of stuff. We decided internally to keep the dockerfile in `bee-node`, and resolve the dependencies in a different way.

PR #575, opened 2021-05-07
Add pruning logic to cpt2

Alex6323:


bee-ledger/src/workers/consensus/worker.rs
```diff
@@ -240,38 +241,35 @@ where
...
         let (tx, rx) = mpsc::unbounded_channel();
 
         let tangle = node.resource::<MsTangle<N::Backend>>();
+        let tangle_config = node.resource::<TangleConfig>();
         let storage = node.storage();
         let bus = node.bus();
 
         validate_ledger_state(&*storage).await?;
 
-        // TODO should be done in config directly ?
```
thibault-martinez, line 253: Maybe this comment confused you but I was going to do this another way as part of my ledger refactoring/cleaning. I prefer changing the conf values within the config `finish` than here with an override method. You'll probably have a tiny little conflict, sorry :)

Alex6323: yea, that's fine!

Alex6323:
Will be merged into dev instead.

PR #580, opened 2021-05-10
Ledger workers

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-ledger/CHANGELOG.md
+++ b/bee-ledger/CHANGELOG.md
@@ -19,7 +19,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-## 0.3.0 - 2021-05-10
+## 0.3.0 - 2021-05-11
+
+### Added
 
 ### Removed
 
--- a/bee-ledger/CHANGELOG.md
+++ b/bee-ledger/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-## 0.3.0 - 2021-05-11
+## 0.3.0 - 2021-05-12
 
 ### Added
 
--- a/bee-ledger/CHANGELOG.md
+++ b/bee-ledger/CHANGELOG.md
@@ -23,6 +23,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `SnapshotWorker` and associated types and operations;
+- `ConsensusWorker` (White Flag) and associated types and operations;
+- `BalanceDiffs::{negate, negated, output_add, output_sub}`;
+- `impl core::fmt::{Debug, Display} for Unspent`;
+
 ### Removed
 
 - Moved `ConflictReason` to bee-tangle;
--- a/bee-ledger/src/lib.rs
+++ b/bee-ledger/src/lib.rs
@@ -3,7 +3,7 @@
 
 //! A crate that contains all types and features required to compute and maintain the ledger state.
 
-//#![warn(missing_docs)]
+// #![warn(missing_docs)]
 
 pub mod types;
 #[cfg(feature = "workers")]
--- a/bee-ledger/src/lib.rs
+++ b/bee-ledger/src/lib.rs
@@ -3,6 +3,8 @@
 
 //! A crate that contains all types and features required to compute and maintain the ledger state.
 
-#![deny(missing_docs, warnings)]
+//#![warn(missing_docs)]
 
 pub mod types;
+#[cfg(feature = "workers")]
+pub mod workers;
--- a/bee-ledger/src/lib.rs
+++ b/bee-ledger/src/lib.rs
@@ -3,7 +3,7 @@
 
 //! A crate that contains all types and features required to compute and maintain the ledger state.
 
-// #![warn(missing_docs)]
+#![warn(missing_docs)]
 
 pub mod types;
 #[cfg(feature = "workers")]
--- a/bee-ledger/src/lib.rs
+++ b/bee-ledger/src/lib.rs
@@ -3,7 +3,7 @@
 
 //! A crate that contains all types and features required to compute and maintain the ledger state.
 
-#![warn(missing_docs)]
+#![deny(missing_docs)]
 
 pub mod types;
 #[cfg(feature = "workers")]
--- a/bee-ledger/src/types/balance_diff.rs
+++ b/bee-ledger/src/types/balance_diff.rs
@@ -3,7 +3,11 @@
 
 use crate::types::Error;
 
-use bee_message::{address::Address, constants::IOTA_SUPPLY};
+use bee_message::{
+    address::Address,
+    constants::IOTA_SUPPLY,
+    output::{Output, DUST_THRESHOLD},
+};
 
 use std::collections::{
     hash_map::{IntoIter, Iter, IterMut},
@@ -100,6 +104,49 @@ impl BalanceDiffs {
         self.0.get(address)
     }
 
+    /// Negates a `BalanceDiffs`.
+    pub fn negate(&mut self) {
+        for (_, diff) in self.iter_mut() {
+            diff.negate();
+        }
+    }
+
+    pub fn output_add(&mut self, output: &Output) -> Result<(), Error> {
+        match output {
+            Output::SignatureLockedSingle(output) => {
+                self.amount_add(*output.address(), output.amount())?;
+                if output.amount() < DUST_THRESHOLD {
+                    self.dust_outputs_inc(*output.address())?;
+                }
+            }
+            Output::SignatureLockedDustAllowance(output) => {
+                self.amount_add(*output.address(), output.amount())?;
+                self.dust_allowance_add(*output.address(), output.amount())?;
+            }
+            output => return Err(Error::UnsupportedOutputKind(output.kind())),
+        }
+
+        Ok(())
+    }
+
+    pub fn output_sub(&mut self, output: &Output) -> Result<(), Error> {
+        match output {
+            Output::SignatureLockedSingle(output) => {
+                self.amount_sub(*output.address(), output.amount())?;
+                if output.amount() < DUST_THRESHOLD {
+                    self.dust_outputs_dec(*output.address())?;
+                }
+            }
+            Output::SignatureLockedDustAllowance(output) => {
+                self.amount_sub(*output.address(), output.amount())?;
+                self.dust_allowance_sub(*output.address(), output.amount())?;
+            }
+            output => return Err(Error::UnsupportedOutputKind(output.kind())),
+        }
+
+        Ok(())
+    }
+
     /// Adds a given amount to a given address.
     pub fn amount_add(&mut self, address: Address, amount: u64) -> Result<(), Error> {
         let entry = self.0.entry(address).or_default();
@@ -168,13 +215,6 @@ impl BalanceDiffs {
     /// Creates a mutable iterator over the balance diffs.
     pub fn iter_mut(&mut self) -> IterMut<'_, Address, BalanceDiff> {
         self.0.iter_mut()
-    }
-
-    /// Negates a `BalanceDiffs`.
-    pub fn negate(&mut self) {
-        for (_, diff) in self.iter_mut() {
-            diff.negate();
-        }
     }
 }
 
--- a/bee-ledger/src/types/balance_diff.rs
+++ b/bee-ledger/src/types/balance_diff.rs
@@ -111,6 +111,13 @@ impl BalanceDiffs {
         }
     }
 
+    /// Creates a new negated version of a `BalanceDiffs`.
+    pub fn negated(&self) -> Self {
+        let mut new = self.clone();
+        new.negate();
+        new
+    }
+
     /// Adds an output to a `BalanceDiffs`.
     pub fn output_add(&mut self, output: &Output) -> Result<(), Error> {
         match output {
--- a/bee-ledger/src/types/balance_diff.rs
+++ b/bee-ledger/src/types/balance_diff.rs
@@ -111,6 +111,7 @@ impl BalanceDiffs {
         }
     }
 
+    /// Adds an output to a `BalanceDiffs`.
     pub fn output_add(&mut self, output: &Output) -> Result<(), Error> {
         match output {
             Output::SignatureLockedSingle(output) => {
@@ -129,6 +130,7 @@ impl BalanceDiffs {
         Ok(())
     }
 
+    /// Subtracts an output from a BalanceDiffs`.
     pub fn output_sub(&mut self, output: &Output) -> Result<(), Error> {
         match output {
             Output::SignatureLockedSingle(output) => {
--- a/bee-ledger/src/types/balance_diff.rs
+++ b/bee-ledger/src/types/balance_diff.rs
@@ -131,7 +131,7 @@ impl BalanceDiffs {
                 self.amount_add(*output.address(), output.amount())?;
                 self.dust_allowance_add(*output.address(), output.amount())?;
             }
-            output => return Err(Error::UnsupportedOutputKind(output.kind())),
+            Output::Treasury(_) => return Err(Error::UnsupportedOutputKind(output.kind())),
         }
 
         Ok(())
@@ -150,7 +150,7 @@ impl BalanceDiffs {
                 self.amount_sub(*output.address(), output.amount())?;
                 self.dust_allowance_sub(*output.address(), output.amount())?;
             }
-            output => return Err(Error::UnsupportedOutputKind(output.kind())),
+            Output::Treasury(_) => return Err(Error::UnsupportedOutputKind(output.kind())),
         }
 
         Ok(())
--- a/bee-ledger/src/workers/consensus/merkle_hasher.rs
+++ b/bee-ledger/src/workers/consensus/merkle_hasher.rs
@@ -1,0 +1,105 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use bee_message::MessageId;
+
+use crypto::hashes::{Digest, Output};
+
+use std::marker::PhantomData;
+
+/// Leaf domain separation prefix.
+const LEAF_HASH_PREFIX: u8 = 0x00;
+/// Node domain separation prefix.
+const NODE_HASH_PREFIX: u8 = 0x01;
+
+/// Computes the largest power of two inferior to `n`.
+fn largest_power_of_two(n: u32) -> usize {
+    1 << (32 - n.leading_zeros() - 1)
+}
+
+/// A Merkle hasher based on a digest function.
+pub(crate) struct MerkleHasher<D> {
+    marker: PhantomData<D>,
+}
+
+impl<D: Default + Digest> MerkleHasher<D> {
+    /// Creates a new Merkle hasher.
+    pub(crate) fn new() -> Self {
+        Self { marker: PhantomData }
+    }
+
+    /// Returns the digest of the empty hash.
+    fn empty(&mut self) -> Output<D> {
+        D::digest(&[])
+    }
+
+    /// Returns the digest of a Merkle leaf.
+    fn leaf(&mut self, message_id: MessageId) -> Output<D> {
+        let mut hasher = D::default();
+
+        hasher.update([LEAF_HASH_PREFIX]);
+        hasher.update(message_id);
+        hasher.finalize()
+    }
+
+    /// Returns the digest of a Merkle node.
+    fn node(&mut self, message_ids: &[MessageId]) -> Output<D> {
+        let mut hasher = D::default();
+        let (left, right) = message_ids.split_at(largest_power_of_two(message_ids.len() as u32 - 1));
+
+        hasher.update([NODE_HASH_PREFIX]);
+        hasher.update(self.digest_inner(left));
+        hasher.update(self.digest_inner(right));
+        hasher.finalize()
+    }
+
+    /// Returns the digest of a list of hashes as an `Output<D>`.
+    fn digest_inner(&mut self, message_ids: &[MessageId]) -> Output<D> {
+        match message_ids.len() {
+            0 => self.empty(),
+            1 => self.leaf(message_ids[0]),
+            _ => self.node(message_ids),
+        }
+    }
+
+    /// Returns the digest of a list of hashes as a `Vec<u8>`.
+    pub(crate) fn digest(&mut self, message_ids: &[MessageId]) -> Vec<u8> {
+        self.digest_inner(message_ids).to_vec()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    use crypto::hashes::blake2b::Blake2b256;
+
+    use std::str::FromStr;
+
+    #[test]
+    fn tree() {
+        let mut hashes = Vec::new();
+
+        for hash in [
+            "52fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649",
+            "81855ad8681d0d86d1e91e00167939cb6694d2c422acd208a0072939487f6999",
+            "eb9d18a44784045d87f3c67cf22746e995af5a25367951baa2ff6cd471c483f1",
+            "5fb90badb37c5821b6d95526a41a9504680b4e7c8b763a1b1d49d4955c848621",
+            "6325253fec738dd7a9e28bf921119c160f0702448615bbda08313f6a8eb668d2",
+            "0bf5059875921e668a5bdf2c7fc4844592d2572bcd0668d2d6c52f5054e2d083",
+            "6bf84c7174cb7476364cc3dbd968b0f7172ed85794bb358b0c3b525da1786f9f",
+        ]
+        .iter()
+        {
+            hashes.push(MessageId::from_str(hash).unwrap());
+        }
+
+        let hash = MerkleHasher::<Blake2b256>::new().digest(&hashes);
+
+        assert_eq!(
+            hex::encode(hash),
+            "bf67ce7ba23e8c0951b5abaec4f5524360d2c26d971ff226d3359fa70cdb0beb"
+        )
+    }
+}
--- a/bee-ledger/src/workers/consensus/merkle_hasher.rs
+++ b/bee-ledger/src/workers/consensus/merkle_hasher.rs
@@ -13,6 +13,7 @@ const LEAF_HASH_PREFIX: u8 = 0x00;
 const NODE_HASH_PREFIX: u8 = 0x01;
 
 /// Computes the largest power of two less than or equal to `n`.
+/// Undefined behaviour: 0 is not a valid value for `n`.
 fn largest_power_of_two(n: u32) -> usize {
     1 << (32 - n.leading_zeros() - 1)
 }
@@ -55,9 +56,9 @@ impl<D: Default + Digest> MerkleHasher<D> {
 
     /// Returns the digest of a list of hashes as an `Output<D>`.
     fn digest_inner(&mut self, message_ids: &[MessageId]) -> Output<D> {
-        match message_ids.len() {
-            0 => self.empty(),
-            1 => self.leaf(message_ids[0]),
+        match message_ids {
+            [] => self.empty(),
+            [message_id] => self.leaf(*message_id),
             _ => self.node(message_ids),
         }
     }
@@ -79,9 +80,7 @@ mod tests {
 
     #[test]
     fn tree() {
-        let mut hashes = Vec::new();
-
-        for hash in [
+        let hashes = [
             "52fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649",
             "81855ad8681d0d86d1e91e00167939cb6694d2c422acd208a0072939487f6999",
             "eb9d18a44784045d87f3c67cf22746e995af5a25367951baa2ff6cd471c483f1",
@@ -91,9 +90,8 @@ mod tests {
             "6bf84c7174cb7476364cc3dbd968b0f7172ed85794bb358b0c3b525da1786f9f",
         ]
         .iter()
-        {
-            hashes.push(MessageId::from_str(hash).unwrap());
-        }
+        .map(|hash| MessageId::from_str(hash).unwrap())
+        .collect::<Vec<_>>();
 
         let hash = MerkleHasher::<Blake2b256>::new().digest(&hashes);
 
--- a/bee-ledger/src/workers/consensus/merkle_hasher.rs
+++ b/bee-ledger/src/workers/consensus/merkle_hasher.rs
@@ -12,7 +12,7 @@ const LEAF_HASH_PREFIX: u8 = 0x00;
 /// Node domain separation prefix.
 const NODE_HASH_PREFIX: u8 = 0x01;
 
-/// Computes the largest power of two inferior to `n`.
+/// Computes the largest power of two less than or equal to `n`.
 fn largest_power_of_two(n: u32) -> usize {
     1 << (32 - n.leading_zeros() - 1)
 }
--- a/bee-ledger/src/workers/consensus/metadata.rs
+++ b/bee-ledger/src/workers/consensus/metadata.rs
@@ -3,8 +3,8 @@
 
 use crate::types::{BalanceDiffs, ConsumedOutput, CreatedOutput};
 
-use bee_ledger_types::types::ConflictReason;
 use bee_message::{milestone::MilestoneIndex, output::OutputId, MessageId};
+use bee_tangle::ConflictReason;
 
 use std::collections::HashMap;
 
--- a/bee-ledger/src/workers/consensus/metadata.rs
+++ b/bee-ledger/src/workers/consensus/metadata.rs
@@ -9,7 +9,6 @@ use bee_tangle::ConflictReason;
 use std::collections::HashMap;
 
 /// White flag metadata of a milestone confirmation.
-#[derive(Default)]
 pub struct WhiteFlagMetadata {
     /// Index of the confirmed milestone.
     pub(crate) index: MilestoneIndex,
@@ -36,7 +35,14 @@ impl WhiteFlagMetadata {
     pub fn new(index: MilestoneIndex) -> WhiteFlagMetadata {
         WhiteFlagMetadata {
             index,
-            ..Self::default()
+            referenced_messages: 0,
+            excluded_no_transaction_messages: Vec::new(),
+            excluded_conflicting_messages: Vec::new(),
+            included_messages: Vec::new(),
+            created_outputs: HashMap::new(),
+            consumed_outputs: HashMap::new(),
+            balance_diffs: BalanceDiffs::new(),
+            merkle_proof: Vec::new(),
         }
     }
 
--- a/bee-ledger/src/workers/consensus/metadata.rs
+++ b/bee-ledger/src/workers/consensus/metadata.rs
@@ -1,0 +1,47 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::types::{BalanceDiffs, ConsumedOutput, CreatedOutput};
+
+use bee_ledger_types::types::ConflictReason;
+use bee_message::{milestone::MilestoneIndex, output::OutputId, MessageId};
+
+use std::collections::HashMap;
+
+/// White flag metadata of a milestone confirmation.
+#[derive(Default)]
+pub struct WhiteFlagMetadata {
+    /// Index of the confirmed milestone.
+    pub(crate) index: MilestoneIndex,
+    /// The number of messages which were referenced by the confirmed milestone.
+    pub(crate) referenced_messages: usize,
+    /// The messages which were excluded because they did not include a transaction.
+    pub(crate) excluded_no_transaction_messages: Vec<MessageId>,
+    /// The messages which were excluded because they were conflicting with the ledger state.
+    pub(crate) excluded_conflicting_messages: Vec<(MessageId, ConflictReason)>,
+    // The messages which mutate the ledger in the order in which they were applied.
+    pub(crate) included_messages: Vec<MessageId>,
+    /// The outputs created within the confirmed milestone.
+    pub(crate) created_outputs: HashMap<OutputId, CreatedOutput>,
+    /// The outputs consumed within the confirmed milestone.
+    pub(crate) consumed_outputs: HashMap<OutputId, (CreatedOutput, ConsumedOutput)>,
+    /// The balance diffs occurring within the confirmed milestone.
+    pub(crate) balance_diffs: BalanceDiffs,
+    /// The merkle proof of the milestone.
+    pub(crate) merkle_proof: Vec<u8>,
+}
+
+impl WhiteFlagMetadata {
+    /// Creates a new `WhiteFlagMetadata`.
+    pub fn new(index: MilestoneIndex) -> WhiteFlagMetadata {
+        WhiteFlagMetadata {
+            index,
+            ..Self::default()
+        }
+    }
+
+    /// Returns the merkle proof of a `WhiteFlagMetadata`.
+    pub fn merkle_proof(&self) -> &[u8] {
+        &self.merkle_proof
+    }
+}
--- a/bee-ledger/src/workers/consensus/state.rs
+++ b/bee-ledger/src/workers/consensus/state.rs
@@ -1,0 +1,85 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    types::{Balance, Unspent},
+    workers::{
+        error::Error,
+        storage::{self, StorageBackend},
+    },
+};
+
+use bee_message::{
+    address::Address,
+    constants::IOTA_SUPPLY,
+    output::{self, dust_outputs_max},
+};
+use bee_storage::access::AsStream;
+
+use futures::StreamExt;
+
+async fn validate_ledger_unspent_state<B: StorageBackend>(storage: &B, treasury: u64) -> Result<(), Error> {
+    let mut supply: u64 = 0;
+    let mut stream = AsStream::<Unspent, ()>::stream(storage)
+        .await
+        .map_err(|e| Error::Storage(Box::new(e)))?;
+
+    while let Some((output_id, _)) = stream.next().await {
+        let output = storage::fetch_output(storage, &*output_id)
+            .await?
+            .ok_or(Error::MissingUnspentOutput(output_id))?;
+
+        let amount = match output.inner() {
+            output::Output::SignatureLockedSingle(output) => output.amount(),
+            output::Output::SignatureLockedDustAllowance(output) => output.amount(),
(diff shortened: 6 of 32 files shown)


Alex6323:
Are we in a hurry with this? Because otherwise I would wait until we merged the pruning logic to cpt2, and include it in this PR? It's not quite a small addition to `bee-ledger` invalidating a few files in this PR already. And if this gets merged to `dev` and therefore removed from `cpt2` it will make things more difficult for me. Wdyt?

bee-ledger/CHANGELOG.md
```diff
@@ -19,7 +19,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
-## 0.3.0 - 2021-05-10
+## 0.3.0 - 2021-05-11
+
+### Added
```
zesterer, line 24: Unused `Added`?

bee-ledger/src/lib.rs
```diff
@@ -3,6 +3,8 @@
 
 //! A crate that contains all types and features required to compute and maintain the ledger state.
 
-#![deny(missing_docs, warnings)]
+#![warn(missing_docs)]
```
zesterer: Any reason for downgrading to `warn`?

bee-ledger/src/types/balance_diff.rs
```diff
@@ -100,6 +104,51 @@ impl BalanceDiffs {
...
 
+    /// Negates a `BalanceDiffs`.
+    pub fn negate(&mut self) {
+        for (_, diff) in self.iter_mut() {
+            diff.negate();
+        }
+    }
+
+    /// Adds an output to a `BalanceDiffs`.
+    pub fn output_add(&mut self, output: &Output) -> Result<(), Error> {
```
zesterer, line 122: Is `Output` not trivially copyable?

bee-ledger/src/types/balance_diff.rs
```diff
@@ -100,6 +104,51 @@ impl BalanceDiffs {
...
+                self.amount_add(*output.address(), output.amount())?;
+                if output.amount() < DUST_THRESHOLD {
+                    self.dust_outputs_inc(*output.address())?;
+                }
+            }
+            Output::SignatureLockedDustAllowance(output) => {
+                self.amount_add(*output.address(), output.amount())?;
+                self.dust_allowance_add(*output.address(), output.amount())?;
+            }
+            output => return Err(Error::UnsupportedOutputKind(output.kind())),
```
zesterer: Perhaps explicitly mention the other variants so that an error is produced here if more are added?

bee-ledger/src/workers/consensus/worker.rs
```diff
@@ -0,0 +1,328 @@
...
+                }
+
+                if !tangle.is_confirmed() {
+                    continue;
+                }
+
+                if should_snapshot(&tangle, MilestoneIndex(*ledger_index), depth, &snapshot_config) {
+                    // if let Err(e) = snapshot(snapshot_config.path(), event.index - depth) {
+                    //     error!("Failed to create snapshot: {:?}.", e);
+                    // }
```
Adam-Gleave, line 314: Looks like this is TODO going by the content of `pruning/mod.rs`, might be worth marking it.

bee-ledger/src/workers/consensus/worker.rs
```diff
@@ -0,0 +1,328 @@
...
+                }
+
+                if should_prune(&tangle, MilestoneIndex(*ledger_index), delay, &pruning_config) {
+                    // if let Err(e) = prune_database(&tangle, MilestoneIndex(*event.index - delay)) {
+                    //     error!("Failed to prune database: {:?}.", e);
+                    // }
```
Adam-Gleave, line 321: Same here.

bee-ledger/src/workers/consensus/merkle_hasher.rs
```diff
@@ -0,0 +1,105 @@
...
+use crypto::hashes::{Digest, Output};
+
+use std::marker::PhantomData;
+
+/// Leaf domain separation prefix.
+const LEAF_HASH_PREFIX: u8 = 0x00;
+/// Node domain separation prefix.
+const NODE_HASH_PREFIX: u8 = 0x01;
+
+/// Computes the largest power of two inferior to `n`.
```
zesterer: This comment is wrong. It should read 'inferior to/less than or equal to `n`'.

bee-ledger/src/workers/consensus/metadata.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub(crate) consumed_outputs: HashMap<OutputId, (CreatedOutput, ConsumedOutput)>,
+    /// The balance diffs occurring within the confirmed milestone.
+    pub(crate) balance_diffs: BalanceDiffs,
+    /// The merkle proof of the milestone.
+    pub(crate) merkle_proof: Vec<u8>,
+}
+
+impl WhiteFlagMetadata {
+    /// Creates a new `WhiteFlagMetadata`.
+    pub fn new(index: MilestoneIndex) -> WhiteFlagMetadata {
```
zesterer, line 35: If the 'standard' way to create this type (`::new`) requires an index, does it make sense to derive `Default` on it?

bee-ledger/src/workers/consensus/white_flag.rs
```diff
@@ -0,0 +1,298 @@
...
+    Message, MessageId,
+};
+use bee_tangle::{ConflictReason, MsTangle};
+
+use crypto::hashes::blake2b::Blake2b256;
+
+use std::collections::{HashMap, HashSet};
+
+fn verify_signature(address: &Address, unlock_blocks: &UnlockBlocks, index: usize, essence_hash: &[u8; 32]) -> bool {
+    match unlock_blocks.get(index) {
```
zesterer: I think clippy will scream at you for this one

bee-ledger/src/workers/consensus/white_flag.rs
```diff
@@ -0,0 +1,298 @@
...
+                storage,
+                message_id,
+                &transaction.id(),
+                essence,
+                transaction.unlock_blocks(),
+                metadata,
+            )
+            .await
+        }
+        essence => Err(Error::UnsupportedTransactionEssenceKind(essence.kind())),
```
zesterer: Match on variants to inform future developers when new variants are added?

bee-ledger/src/workers/consensus/white_flag.rs
```diff
@@ -0,0 +1,298 @@
...
+    message: &Message,
+    metadata: &mut WhiteFlagMetadata,
+) -> Result<(), Error> {
+    metadata.referenced_messages += 1;
+
+    match message.payload() {
+        Some(Payload::Transaction(transaction)) => {
+            match apply_transaction(storage, message_id, transaction, metadata).await? {
+                ConflictReason::None => metadata.included_messages.push(*message_id),
+                conflict => metadata.excluded_conflicting_messages.push((*message_id, conflict)),
```
zesterer, line 203: Same here (or switch to `if let`)

bee-ledger/src/workers/consensus/white_flag.rs
```diff
@@ -0,0 +1,298 @@
...
+            Some((message, meta)) => {
+                if meta.flags().is_referenced() {
+                    visited.insert(*message_id);
+                    messages_ids.pop();
+                    continue;
+                }
+
+                let mut next = None;
+
+                for parent in message.parents().iter() {
```
zesterer:
```rust
if let Some(unvisited) = message.parents().iter()
    .skip_while(|p| visited.contains(p))
    .next()
{
    message_ids.push(*next);
} else {
    apply_message(storage, message_id, &message, metadata).await?;
    visited.insert(*message_id);
    messages_ids.pop();
}
```

Probably communicates intent a bit better and removes mutation + loop

bee-ledger/src/workers/consensus/white_flag.rs
```diff
@@ -0,0 +1,298 @@
...
+                    }
+                }
+            }
+            None => {
+                if !tangle.is_solid_entry_point(message_id).await {
+                    return Err(Error::MissingMessage(*message_id));
+                } else {
+                    visited.insert(*message_id);
+                    messages_ids.pop();
+                    continue;
```
zesterer: This `continue;` is executed at the end of the loop body and so does nothing.

bee-ledger/src/workers/consensus/white_flag.rs
```diff
@@ -0,0 +1,298 @@
...
+            metadata.referenced_messages,
+            metadata.excluded_no_transaction_messages.len(),
+            metadata.excluded_conflicting_messages.len(),
+            metadata.included_messages.len(),
+        ));
+    }
+
+    let diff_sum = metadata.balance_diffs.iter().map(|(_, diff)| diff.amount()).sum();
+
+    if diff_sum != 0 {
```
zesterer: Switch to `if { .. } else { .. }` instead of an early return to communicate intent more clearly and make incorrect refactoring harder

bee-ledger/src/workers/consensus/worker.rs
```diff
@@ -0,0 +1,328 @@
...
+            Payload::TreasuryTransaction(treasury) => match treasury.output() {
+                Output::Treasury(output) => output.clone(),
+                output => return Err(Error::UnsupportedOutputKind(output.kind())),
+            },
+            payload => return Err(Error::UnsupportedPayloadKind(payload.kind())),
+        },
+        milestone_id,
+    );
+
+    Ok(Migration::new(receipt, consumed_treasury, created_treasury))
```
zesterer:
```rust
match receipt.inner().transaction() {
    Payload::TreasuryTransaction(treasury) => match treasury.output() {
        Output::Treasury(output) => Ok(Migration::new(receipt, consumed_treasury, TreasuryOutput::new(output.clone()))),
        Output::Foo | Output::Bar => Err(Error::UnsupportedOutputKind(output.kind())),
    },
    payload => Err(Error::UnsupportedPayloadKind(payload.kind())),
}
```

Replacing `Foo`/`Bar` with unsupported payload kinds.

bee-ledger/src/workers/consensus/worker.rs
```diff
@@ -0,0 +1,328 @@
...
+            *milestone.essence().index(),
+            **ledger_index,
+        ));
+    }
+
+    let mut metadata = WhiteFlagMetadata::new(milestone.essence().index());
+
+    let parents = message.parents().iter().copied().collect();
+
+    drop(message);
```
zesterer: I usually add a comment with each explicit drop to explain why the drop is occurring, and what invariants must not be invalidated for it to be an okay place to drop it. Is there a lock guard being implicitly held? Good idea to prevent refactors incorrectly moving things across the border.

bee-ledger/src/types/balance_diff.rs
thibault-martinez: It's not, but we don't even need ownership in this function.

bee-ledger/src/workers/consensus/merkle_hasher.rs
thibault-martinez: Indeed

bee-ledger/src/workers/consensus/worker.rs
```diff
@@ -0,0 +1,328 @@
...
+                    // if let Err(e) = snapshot(snapshot_config.path(), event.index - depth) {
+                    //     error!("Failed to create snapshot: {:?}.", e);
+                    // }
+                }
+
+                if should_prune(&tangle, MilestoneIndex(*ledger_index), delay, &pruning_config) {
+                    // if let Err(e) = prune_database(&tangle, MilestoneIndex(*event.index - delay)) {
+                    //     error!("Failed to prune database: {:?}.", e);
+                    // }
+                }
```
zesterer: Unused checks?

bee-ledger/src/workers/consensus/worker.rs
```diff
@@ -0,0 +1,328 @@
...
+                )
+                .await
+                {
+                    error!("Confirmation error on {}: {}.", message_id, e);
+                    panic!("Aborting due to unexpected ledger error.");
+                }
+
+                if !tangle.is_confirmed() {
+                    continue;
+                }
```
zesterer, line 308: This is at the end of the loop body and so has no effect (assuming the checks below are not effectful).

bee-ledger/src/workers/pruning/condition.rs
```diff
@@ -0,0 +1,77 @@
...
+        config.interval_synced()
+    } else {
+        config.interval_unsynced()
+    };
+
+    if (solid_index < depth + snapshot_interval)
+        || (solid_index - depth) < pruning_index + 1 + SOLID_ENTRY_POINT_THRESHOLD_PAST
+    {
+        // Not enough history to calculate solid entry points.
+        return false;
```
zesterer, line 35: Unnecessary early return

bee-ledger/src/workers/pruning/condition.rs
```diff
@@ -0,0 +1,77 @@
...
+    if *tangle.get_snapshot_index() < SOLID_ENTRY_POINT_THRESHOLD_PAST + PRUNING_THRESHOLD + 1 {
+        return false;
+    }
+
+    let target_index_max =
+        MilestoneIndex(*tangle.get_snapshot_index() - SOLID_ENTRY_POINT_THRESHOLD_PAST - PRUNING_THRESHOLD - 1);
+
+    if index > target_index_max {
+        index = target_index_max;
+    }
```
zesterer, line 65: Rebinding like

```rust
let index = index.min(target_index_max);
```

is clearer and also removes the need for mutation.

(`min` is automatically implemented for all `Ord`: https://doc.rust-lang.org/std/cmp/trait.Ord.html#method.min )

bee-ledger/src/workers/consensus/metadata.rs
thibault-martinez: Removed

bee-ledger/src/workers/pruning/config.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl PruningConfigBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn enabled(mut self, enabled: bool) -> Self {
+        self.enabled.replace(enabled);
+        self
+    }
```
zesterer, line 25: It's not a huge issue, but something like:

```rust
Self {
    enabled: Some(enabled),
    ..self
}
```

is more conventional and easier to understand at a glance.

(Ditto for others in this file)

bee-ledger/src/workers/consensus/white_flag.rs
thibault-martinez: It doesn't, why should it ?

bee-ledger/src/workers/pruning/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+
+// TODO remove with pruning PR
+#![allow(missing_docs)]
+
+pub mod config;
+
+pub(crate) mod condition;
+pub(crate) mod constants;
+
+// use bee_protocol::{
```
zesterer, line 12: Unused code?

bee-ledger/src/workers/consensus/white_flag.rs
thibault-martinez: How would I get the conflict on the else branch with an if let ?

bee-ledger/src/workers/snapshot/config.rs
```diff
@@ -0,0 +1,128 @@
...
+const DEFAULT_FULL_PATH: &str = "./snapshots/mainnet/latest-full_snapshot.bin";
+const DEFAULT_DOWNLOAD_URLS: Vec<String> = Vec::new();
+const DEFAULT_DEPTH: u32 = 50;
+const DEFAULT_INTERVAL_SYNCED: u32 = 50;
+const DEFAULT_INTERVAL_UNSYNCED: u32 = 1000;
+
+/// Builder for a `SnapshotConfig`.
+#[derive(Default, Deserialize)]
+pub struct SnapshotConfigBuilder {
+    full_path: Option<String>,
```
zesterer: Perhaps consider just using `PathBuf` here? It's got tighter constraints than `String`.

bee-ledger/src/workers/consensus/white_flag.rs
thibault-martinez: Removed

bee-ledger/src/workers/snapshot/config.rs
thibault-martinez: Using String for serde I guess

bee-ledger/src/workers/consensus/worker.rs
thibault-martinez: This is the one I asked you the drop question for. Can you confirm I can safely remove it ?

bee-ledger/src/workers/snapshot/download.rs
```diff
@@ -0,0 +1,44 @@
...
+        let url = url.to_owned() + &file_name.to_string_lossy();
+
+        info!("Downloading snapshot file {}...", url);
+
+        match reqwest::get(&url).await.and_then(|res| res.error_for_status()) {
+            Ok(res) => {
+                copy(
+                    &mut res.bytes().await.map_err(|_| Error::DownloadingFailed)?.as_ref(),
+                    &mut File::create(file_path)?,
+                )?;
```
zesterer: Use `tokio` routines here, this is blocking IO in an async function. See:

https://docs.rs/tokio/1.5.0/tokio/io/fn.copy.html

https://docs.rs/tokio/1.5.0/tokio/fs/struct.File.html#method.open

etc.

bee-ledger/src/workers/snapshot/download.rs
```diff
@@ -0,0 +1,44 @@
...
+pub(crate) async fn download_snapshot_file(file_path: &Path, download_urls: &[String]) -> Result<(), Error> {
+    let file_name = file_path
+        .file_name()
+        .ok_or_else(|| Error::InvalidFilePath(file_path.to_string_lossy().to_string()))?;
+
+    std::fs::create_dir_all(
+        file_path
+            .parent()
+            .ok_or_else(|| Error::InvalidFilePath(file_path.to_string_lossy().to_string()))?,
+    )
```
zesterer: Use a non-blocking equivalent: https://docs.rs/tokio/1.5.0/tokio/fs/fn.create_dir_all.html

bee-ledger/src/workers/snapshot/download.rs
```diff
@@ -0,0 +1,44 @@
...
+                    &mut res.bytes().await.map_err(|_| Error::DownloadingFailed)?.as_ref(),
+                    &mut File::create(file_path)?,
+                )?;
+                break;
+            }
+            Err(e) => warn!("Downloading snapshot file failed with status code {:?}.", e.status()),
+        }
+    }
+
+    if !file_path.exists() {
```
zesterer, line 41: This is (subtly) blocking

bee-ledger/src/workers/snapshot/import.rs
```diff
@@ -0,0 +1,317 @@
...
+        let mut balance_diffs = BalanceDiffs::new();
+        let mut consumed = HashMap::new();
+
+        for (_, output) in diff.created().iter() {
+            balance_diffs.output_add(output.inner())?;
+        }
+
+        for (output_id, (created_output, consumed_output)) in diff.consumed().iter() {
+            balance_diffs.output_sub(created_output.inner())?;
+            consumed.insert(*output_id, (created_output.clone(), consumed_output.clone()));
```
zesterer: It doesn't look like `consumed` is mutated after this line. Consider a `map` + `collect` instead? It'll likely result in fewer allocations too.

bee-ledger/src/workers/snapshot/import.rs
```diff
@@ -0,0 +1,317 @@
...
+        match index {
+            index if index == MilestoneIndex(ledger_index + 1) => {
+                apply_milestone(&*storage, index, diff.created(), &consumed, &balance_diffs, &migration).await?;
+            }
+            index if index == MilestoneIndex(ledger_index) => {
+                rollback_milestone(&*storage, index, diff.created(), &consumed, &balance_diffs, &migration).await?;
+            }
+            _ => return Err(Error::Snapshot(SnapshotError::UnexpectedMilestoneDiffIndex(index))),
+        }
+    }
```
zesterer: This is better suited to an `if`/`else` chain than a `match`.

bee-ledger/src/workers/snapshot/import.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    if network_id != header.network_id() {
+        return Err(Error::Snapshot(SnapshotError::NetworkIdMismatch(
+            network_id,
+            header.network_id(),
+        )));
+    }
+
+    Ok(())
```
zesterer: Unnecessary early returns

bee-ledger/src/workers/snapshot/import.rs
```diff
@@ -0,0 +1,317 @@
...
+    if let Some(delta_path) = config.delta_path() {
+        if !delta_exists
+            && download_snapshot_file(delta_path, config.download_urls())
+                .await
+                .is_err()
+        {
+            warn!("Could not download the delta snapshot file and it will not be imported.");
+        }
+
+        if Path::exists(delta_path) {
```
zesterer: Why use the fully-qualified path here? (also, this is blocking)

bee-ledger/src/workers/snapshot/worker.rs
```diff
@@ -0,0 +1,94 @@
...
+                );
+            }
+        }
+
+        let mut solid_entry_points_stream = AsStream::<SolidEntryPoint, MilestoneIndex>::stream(&*storage)
+            .await
+            .map_err(|e| Error::Storage(Box::new(e)))?;
+        let mut solid_entry_points = HashMap::new();
+
+        while let Some((sep, index)) = solid_entry_points_stream.next().await {
```
zesterer: Use https://docs.rs/futures/0.3.13/futures/stream/trait.StreamExt.html#method.collect instead

bee-ledger/src/workers/storage.rs
```diff
@@ -0,0 +1,496 @@
...
+) -> Result<(), Error> {
+    match address {
+        Address::Ed25519(address) => {
+            Batch::<(Ed25519Address, OutputId), ()>::batch_insert(storage, batch, &(*address, *output_id), &())
+                .map_err(|e| Error::Storage(Box::new(e)))?;
+        }
+        address => return Err(Error::UnsupportedAddressKind(address.kind())),
+    }
+
+    Ok(())
```
zesterer: You can remove the `?` on line 95, the `return` on line 97, and the `Ok(())` on line 100.

bee-ledger/src/workers/storage.rs
```diff
@@ -0,0 +1,496 @@
...
+) -> Result<(), Error> {
+    match address {
+        Address::Ed25519(address) => {
+            Batch::<(Ed25519Address, OutputId), ()>::batch_delete(storage, batch, &(*address, *output_id))
+                .map_err(|e| Error::Storage(Box::new(e)))?;
+        }
+        address => return Err(Error::UnsupportedAddressKind(address.kind())),
+    }
+
+    Ok(())
```
zesterer: Same here

bee-ledger/src/workers/storage.rs
```diff
@@ -0,0 +1,496 @@
...
+        Output::SignatureLockedSingle(output) => {
+            insert_output_id_for_address_batch(storage, batch, output.address(), output_id)?
+        }
+        Output::SignatureLockedDustAllowance(output) => {
+            insert_output_id_for_address_batch(storage, batch, output.address(), output_id)?
+        }
+        output => return Err(Error::UnsupportedOutputKind(output.kind())),
+    }
+
+    Ok(())
```
zesterer: Remove the `?` from lines 133 and 136, the `return` from 138 and the `Ok(())` from 141.

bee-ledger/src/workers/storage.rs
```diff
@@ -0,0 +1,496 @@
...
+
+    for (output_id, output) in created_outputs.iter() {
+        insert_created_output_batch(storage, &mut batch, output_id, output)?;
+        created_output_ids.push(*output_id);
+    }
+
+    for (output_id, (_, consumed_output)) in consumed_outputs.iter() {
+        insert_consumed_output_batch(storage, &mut batch, output_id, consumed_output)?;
+        consumed_output_ids.push(*output_id);
+    }
```
zesterer:
```rust
let created_output_ids = created_outputs.iter()
    .map(|(output_id, output)| {
        insert_created_output_batch(storage, &mut batch, output_id, output)?;
        Ok(*output_id)
    })
    .collect::<Result<Vec<_>, _>>()?;
```

Ditto with `consumed_output_ids`.

Death to mutability! (also means you don't need to think about the `with_capacity`)

bee-ledger/src/workers/pruning/condition.rs
thibault-martinez: Will be dealt with in another PR.

bee-ledger/src/workers/pruning/condition.rs
thibault-martinez: Will be dealt with in another PR.

bee-ledger/src/workers/pruning/config.rs
thibault-martinez: Will be dealt with in another PR.

bee-ledger/src/workers/storage.rs
```diff
@@ -0,0 +1,496 @@
...
+        delete_created_output_batch(storage, &mut batch, output_id, created_output)?;
+    }
+
+    for (output_id, (created_output, _)) in consumed_outputs.iter() {
+        insert_created_output_batch(storage, &mut batch, output_id, created_output)?;
+        delete_consumed_output_batch(storage, &mut batch, output_id)?;
+    }
+
+    let mut balance_diffs = balance_diffs.clone();
+    balance_diffs.negate();
```
zesterer: Is it worth adding a `.negated() -> Self` method to `&BalanceDiffs`?

bee-ledger/src/workers/storage.rs
```diff
@@ -0,0 +1,496 @@
...
+    {
+        Some(outputs) => {
+            match outputs.len() {
+                // There has to be an unspent treasury output at all time.
+                0 => panic!("No unspent treasury output found"),
+                // Indexing is fine since length is known
+                1 => Ok(outputs[0].clone()),
+                // There should be one and only one unspent treasury output at all time.
+                _ => panic!("More than one unspent treasury output found"),
+            }
```
zesterer: You can use a slice pattern here to avoid the double check.

bee-ledger/src/workers/storage.rs
```diff
@@ -0,0 +1,496 @@
...
+    match Fetch::<bool, Vec<TreasuryOutput>>::fetch(storage, &false)
+        .await
+        .map_err(|e| Error::Storage(Box::new(e)))?
+    {
+        Some(outputs) => {
+            match outputs.len() {
+                // There has to be an unspent treasury output at all time.
+                0 => panic!("No unspent treasury output found"),
+                // Indexing is fine since length is known
+                1 => Ok(outputs[0].clone()),
```
zesterer: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove allows you to remove the entry you want without a `.clone()`.

bee-ledger/src/workers/snapshot/import.rs
thibault-martinez: How so ?

bee-ledger/src/workers/pruning/mod.rs
thibault-martinez: Will be dealt with in another PR.

bee-ledger/src/workers/consensus/worker.rs
thibault-martinez: Will be dealt with in another PR.

bee-ledger/src/workers/consensus/white_flag.rs
thibault-martinez: Pretty cool, thanks

bee-ledger/src/workers/snapshot/download.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::workers::snapshot::error::Error;
+
+use log::{info, warn};
+
+use std::{fs::File, io::copy, path::Path};
+
+pub(crate) async fn download_snapshot_file(file_path: &Path, download_urls: &[String]) -> Result<(), Error> {
```
samuel-rufi, line 10: Can we change the function signature to something like that:

`pub(crate) async fn download_snapshot_file(file_path: &Path, download_url: &String) -> Result<(), Error>`

The  `download_snapshot_file(...)` will be called 2 times, one for delta and one for full snapshot.

Where `download_url` already links to the exact file to download. The full url to the file.
And `file_path` will be used only to specify the location where to save the snapshot locally.

By this, `download_url` and `file_path` are decoupled. A user then can use whatever name for the snapshots in the config.toml.

bee-ledger/src/workers/storage.rs
thibault-martinez: The `?` is not working here

bee-ledger/src/workers/snapshot/download.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    std::fs::create_dir_all(
+        file_path
+            .parent()
+            .ok_or_else(|| Error::InvalidFilePath(file_path.to_string_lossy().to_string()))?,
+    )
+    .map_err(|_| Error::InvalidFilePath(file_path.to_string_lossy().to_string()))?;
+
+    for url in download_urls {
+        let url = url.to_owned() + &file_name.to_string_lossy();
```
samuel-rufi, line 24: When the url fully specify the file, we wouldn't need to do this here.

bee-ledger/src/workers/consensus/white_flag.rs
zesterer: You're matching on a single variant (i.e: it can be replaced with `if let`)

bee-ledger/src/workers/consensus/white_flag.rs
zesterer: Ah, good point. Didn't read it properly, I thought it was coming from a function parameter.

bee-ledger/src/workers/snapshot/download.rs
thibault-martinez: Can you open an issue for that please ? This is significant enough to be done outside of this PR.

samuel-rufi: So the config.toml could look like that:

```
[snapshot]
full_path  = "./snapshots/testnet/my_full_snapshot.bin"
delta_path  = "./snapshots/testnet/my_delta_snapshot.bin"
delta_url = "https://dbfiles.testnet.chrysalis2.com/delta_snapshot.bin"
full_url = "https://dbfiles.testnet.chrysalis2.com/full_snapshot.bin"
```

bee-ledger/src/workers/consensus/worker.rs
zesterer: Yes, I just checked. It can be removed.

bee-ledger/src/workers/snapshot/download.rs
thibault-martinez: Any suggestion ?

bee-ledger/src/workers/snapshot/import.rs
thibault-martinez: Why not ?

bee-ledger/src/workers/snapshot/download.rs
samuel-rufi: Okey yes will create a issue for that^

bee-ledger/src/workers/storage.rs
thibault-martinez: What double check ?

bee-ledger/src/workers/consensus/worker.rs
thibault-martinez: For the two blocks below, we need the tangle to be confirmed

zesterer: Those blocks are effectful? They're worded like they're not. If that's the case, `check_x` might be a better name than `is_x`.

thibault-martinez: What do you mean effectful ? What is commented will be uncommented at some point. It's doing pruning and snapshot so yes, it has an effect.

bee-ledger/src/workers/snapshot/config.rs
zesterer: image

There's an implementation for `PathBuf` too.

bee-ledger/src/workers/snapshot/download.rs
zesterer: I'm not sure `tokio` actually has an equivalent. I guess https://docs.rs/tokio/1.5.0/tokio/fs/fn.metadata.html will effectively do the same thing though.

zesterer: There's some interesting discussion in this issue that might be relevant to us too, actually: https://github.com/tokio-rs/tokio/pull/3375

bee-ledger/src/workers/snapshot/import.rs
zesterer:
```rust
if kind != header.kind() {
    Err(Error::Snapshot(SnapshotError::UnexpectedSnapshotKind(
        kind,
        header.kind(),
    )))
} else if network_id != header.network_id() {
    Err(Error::Snapshot(SnapshotError::NetworkIdMismatch(
        network_id,
        header.network_id(),
    )))
} else {
    Ok(())
}
```

bee-ledger/src/workers/snapshot/import.rs
zesterer: `delta_path` is `impl Deref<Target=Path>`, I think? You should be able to just do `delta_path.exists()`.

bee-ledger/src/workers/storage.rs
zesterer: What error did you get? It should do.

bee-ledger/src/workers/storage.rs
zesterer: > Indexing is fine since length is known

zesterer:
```rust
match outputs {
    [] => // Same as the 0 case
    [output] => Ok(output.clone()), // Same as the 1 case
    [_, ..] => // Same as the >1 case
}
```

zesterer: (You might need to take `outputs` by reference if it's an owned vector)

bee-ledger/src/workers/consensus/worker.rs
zesterer: Okay, so the intention is to uncomment them. That's fair enough.

bee-ledger/src/workers/snapshot/config.rs
thibault-martinez: Ah, I missed that, perfect then

bee-ledger/src/workers/snapshot/import.rs
thibault-martinez: Ahhhh sorry I thought you meant full actual delta path, yeah sure

bee-ledger/src/workers/storage.rs
thibault-martinez: Was lacking some time annotation

bee-ledger/src/lib.rs
Alex6323: I'ld keep `missing_docs` denied.

bee-ledger/src/workers/consensus/merkle_hasher.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Leaf domain separation prefix.
+const LEAF_HASH_PREFIX: u8 = 0x00;
+/// Node domain separation prefix.
+const NODE_HASH_PREFIX: u8 = 0x01;
+
+/// Computes the largest power of two less than or equal to `n`.
+fn largest_power_of_two(n: u32) -> usize {
+    1 << (32 - n.leading_zeros() - 1)
+}
```
Alex6323: `0` is not a valid input for this function. I think we should at least mention it in the docs of this function. Even better would be to make sure that this can not be exploited in any way to mess with the ledger or the node operation.

bee-ledger/src/workers/consensus/merkle_hasher.rs
```diff
@@ -0,0 +1,105 @@
...
+        hasher.finalize()
+    }
+
+    /// Returns the digest of a list of hashes as an `Output<D>`.
+    fn digest_inner(&mut self, message_ids: &[MessageId]) -> Output<D> {
+        match message_ids.len() {
+            0 => self.empty(),
+            1 => self.leaf(message_ids[0]),
+            _ => self.node(message_ids),
+        }
```
Alex6323: I guess Josh made a comment somewhere in a very similar situation about matching the slice directly. Would this not also apply here?
```rust
match message_ids {
    [] => self.empty(),
    [single] => self.leaf(single),
    [_, ..] => self.node(message_ids),
}
```

bee-ledger/src/workers/consensus/merkle_hasher.rs
```diff
@@ -0,0 +1,105 @@
...
+            "5fb90badb37c5821b6d95526a41a9504680b4e7c8b763a1b1d49d4955c848621",
+            "6325253fec738dd7a9e28bf921119c160f0702448615bbda08313f6a8eb668d2",
+            "0bf5059875921e668a5bdf2c7fc4844592d2572bcd0668d2d6c52f5054e2d083",
+            "6bf84c7174cb7476364cc3dbd968b0f7172ed85794bb358b0c3b525da1786f9f",
+        ]
+        .iter()
+        {
+            hashes.push(MessageId::from_str(hash).unwrap());
+        }
+
```
Alex6323: collect into `Vec`?

bee-ledger/src/workers/consensus/metadata.rs
```diff
@@ -0,0 +1,53 @@
...
+        WhiteFlagMetadata {
+            index,
+            referenced_messages: 0,
+            excluded_no_transaction_messages: Vec::new(),
+            excluded_conflicting_messages: Vec::new(),
+            included_messages: Vec::new(),
+            created_outputs: HashMap::new(),
+            consumed_outputs: HashMap::new(),
+            balance_diffs: BalanceDiffs::new(),
+            merkle_proof: Vec::new(),
```
Alex6323, line 45: Do we maybe know size of the proof upfront or at least a lower bound? Does it get filled in any case? If both questions can be answered with `yes` then maybe this is a scenario where pre-allocating a certain amount of bytes makes sense, so we can prevent unnecessary re-allocations.

bee-ledger/src/workers/consensus/state.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    if supply
+        .checked_add(treasury)
+        .ok_or_else(|| Error::LedgerStateOverflow(supply as u128 + treasury as u128))?
+        != IOTA_SUPPLY
+    {
+        return Err(Error::InvalidLedgerUnspentState(supply));
+    }
+
+    Ok(())
```
Alex6323: I have no problems with early returns tbh, but maybe you want to consider rewriting this to :
```rust
if .... != IOTA_SUPPLY {
    Err(Error::InvalidLedgerUnspentState(supply))
} else {
    Ok(())
}
```
because this is the essential check of this function. Same in the cases down below.

bee-ledger/src/workers/consensus/white_flag.rs
```diff
@@ -0,0 +1,288 @@
...
+                conflict => metadata.excluded_conflicting_messages.push((*message_id, conflict)),
+            }
+        }
+        _ => metadata.excluded_no_transaction_messages.push(*message_id),
+    }
+
+    Ok(())
+}
+
+async fn traversal<B: StorageBackend>(
```
Alex6323: Can we indicate which type of traversal this is to be a bit more self-documenting? Seems like it's traversing the past cone, so maybe `past_cone_traversal`, or just `past_traversal`, or in its imperative form: `traverse_past_cone`/`traverse_past`.

bee-ledger/src/workers/consensus/white_flag.rs
```diff
@@ -0,0 +1,288 @@
...
+                }
+            }
+        }
+    }
+
+    Ok(())
+}
+
+/// Computes the ledger state according to the White Flag method.
+pub async fn white_flag<B: StorageBackend>(
```
Alex6323, line 253: Maybe not now, but I think in the future we should be more descriptive here, what the `White Flag` method is actually. Would be nice to read it broken down into 1 or 2 single sentences what is being done in this function according to White Flag. For now a link to the White Flag RFC would be a nice addition to the doc comment.

bee-ledger/src/workers/snapshot/worker.rs
```diff
@@ -0,0 +1,89 @@
...
+        vec![TypeId::of::<TangleWorker>()].leak()
+    }
+
+    async fn start(node: &mut N, config: Self::Config) -> Result<Self, Self::Error> {
+        let (network_id, snapshot_config) = config;
+        let tangle = node.resource::<MsTangle<N::Backend>>();
+        let storage = node.storage();
+
+        match storage::fetch_snapshot_info(&*storage).await? {
+            None => {
```
Alex6323: maybe consider using `if let Some(..) = ... { ... } else { ... }` here as it reduces a bit of noise, and puts the happy path (we are happy, right?) first.

bee-ledger/src/workers/storage.rs
```diff
@@ -0,0 +1,483 @@
...
+) -> Result<(), Error> {
+    Batch::<(bool, TreasuryOutput), ()>::batch_insert(storage, batch, &(false, treasury_output.clone()), &())
+        .map_err(|e| Error::Storage(Box::new(e)))?;
+    Batch::<(bool, TreasuryOutput), ()>::batch_delete(storage, batch, &(true, treasury_output.clone()))
+        .map_err(|e| Error::Storage(Box::new(e)))
+}
+
+/// Fetches the unspent treasury output from the storage.
+pub async fn fetch_unspent_treasury_output<B: StorageBackend>(storage: &B) -> Result<TreasuryOutput, Error> {
+    match Fetch::<bool, Vec<TreasuryOutput>>::fetch(storage, &false)
```
Alex6323: `if let` ?

bee-ledger/src/workers/consensus/metadata.rs
thibault-martinez: We can't know for sure, the proof size depends on the chosen hasher.

bee-ledger/src/workers/consensus/merkle_hasher.rs
thibault-martinez: It can't be abused as it's private and only used in contexts where we know it's not 0 but I'll add a comment

Alex6323 (dismissed):

zesterer (dismissed):

zesterer (approved):

samuel-rufi (approved):

Adam-Gleave (approved):

PR #588, opened 2021-05-18
Add endpoints module

samuel-rufi:


samuel-rufi:
Will be replaced with a future PR.

PR #589, opened 2021-05-18
Add "console" feature for task diagnostics with tokio-console

Adam-Gleave:
# Description of change

Adds the necessary instrumentation to expose Bee worker tasks (spawned in the node `spawn` method) to [`tokio-console`](https://github.com/tokio-rs/console), for improved `async` diagnostics. This can be used with the `console` feature.

## Type of change

I have run this locally (with the feature flag both enabled and disabled) to ensure that the new functionality works, and that the node operates as expected with and without the new instrumentation:

Capture

## Change checklist

# How does it work?
This change adds instrumentation to Bee tasks in the form of `span`s, provided by the `tracing` crate. These are helpful for producing diagnostic information for a period of time, rather than a one-off event. They live for the scope of the task that they instrument:
```rust
let span = tracing::info_span!(
    target: "tokio::task", 
    "task", 
    file = caller.file(), 
    line = caller.line(),
);

tokio::spawn(future.instrument(span))
```
In order to collect information and log these spans, `tracing_subscriber` can be used and customised to handle span events (on entry, on exit, etc). The `console_subscriber` crate in the `tokio-console` project provides a way to customise a `tracing_subscriber` to broadcast structured gRPC messages on these events. The following snippet creates a subscriber that does this, constraining the captured spans to `INFO` level:
```rust
let (layer, server) = console_subscriber::TasksLayer::new();

let filter = tracing_subscriber::EnvFilter::from_default_env()
    .add_directive(tracing::Level::INFO.into())
    .add_directive("tokio=info".parse().unwrap());

tracing_subscriber::registry()
    .with(tracing_subscriber::fmt::layer())
    .with(filter)
    .with(layer)
    .init();
```
The `console` project then interprets the messages to display them in real-time in a console UI.

# How do I use it?
To build/run with the instrumentation, make sure the `console` feature is enabled:
```
> cargo run --release --features console
```
Additionally, your `RUSTFLAGS` environment variable must contain the following, in order for `tokio` to be built with the `tracing` feature enabled:
```
--cfg tokio_unstable
```
Once the node is up and running, clone a local copy of the [`tokio-console`](https://github.com/tokio-rs/console) project, and run the `console` crate from its directory:
```
.../tokio-console/console/ > cargo run
```


bee-node/src/main.rs
```diff
@@ -19,9 +47,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
     };
 
-    if let Err(e) = logger_init(config.logger.clone()) {
-        panic!("Failed to initialise the logger: {}", e);
-    }
+    #[cfg(feature = "console")]
+    let console_handle = logger_init();
+
+    #[cfg(not(feature = "console"))]
+    logger_init(config.logger.clone());
```
thibault-martinez, line 54: I'm confused, why would we not have logging if we have the console ?

Adam-Gleave: The logger is replaced by the `tracing_subscriber`, so if we have already initialised one beforehand, initialising the subscriber panics. The subscriber logs everything it captures to stdout as well as whatever the additional layers direct it to do.

thibault-martinez: Alright, makes sense 👍

bee-node/src/main.rs
```diff
@@ -10,6 +9,35 @@ use log::error;
 
 const CONFIG_PATH: &str = "./config.toml";
 
+#[cfg(feature = "console")]
+fn logger_init() -> tokio::task::JoinHandle<Result<(), Box<dyn std::error::Error + Send + Sync>>> {
```
zesterer, line 13: Perhaps `impl Error` is fine here? It's not like `Box<dyn Error>` is any more useful from a type perspective.

zesterer (approved):

bee-node/src/main.rs
Adam-Gleave: The `serve()` future that's spawned here returns `Result<(), Box<dyn Error + Send + Sync>>`, so that's just being forwarded here. I think there's an implicit `Sized` bound here too, so `impl Error` fails to compile.

thibault-martinez (approved):

PR #605, opened 2021-05-31
Atomic ledger operations

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-ledger/src/workers/consensus/worker.rs
+++ b/bee-ledger/src/workers/consensus/worker.rs
@@ -37,6 +37,7 @@ use tokio_stream::wrappers::UnboundedReceiverStream;
 use std::{any::TypeId, convert::TryInto};
 
 /// Event of the consensus worker.
+#[allow(clippy::type_complexity)]
 pub enum ConsensusWorkerEvent {
     /// Event to confirm a milestone.
     Confirm(MessageId),
--- a/bee-ledger/src/workers/consensus/worker.rs
+++ b/bee-ledger/src/workers/consensus/worker.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    types::{CreatedOutput, LedgerIndex, Migration, Receipt, TreasuryOutput},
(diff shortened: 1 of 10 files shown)


samuel-rufi (dismissed):
To me this looks like a nice solution; this is great! I didn't find any issue with this.

Adam-Gleave (dismissed):

bee-ledger/src/workers/consensus/worker.rs
```diff
@@ -28,15 +29,31 @@ use bee_tangle::{ConflictReason, MsTangle, TangleWorker};
...
+    Output(
+        OutputId,
+        oneshot::Sender<(Result<Option<CreatedOutput>, Error>, LedgerIndex)>,
+    ),
+    /// Event to fetch the outputs of an address.
+    Outputs(
+        Address,
+        oneshot::Sender<(Result<Option<Vec<OutputId>>, Error>, LedgerIndex)>,
+    ),
+}
```
Alex6323: Do we want to put this type into its own module? At some point we may want to add an additional `impl` block to this event type containing some convenience methods, and also impl certain traits for it. Wdyt?

Also: probably typo in doc comment: Event**s**.

Alex6323: Two additional suggestions:
From my understanding of the code those events have a `command`-like notion to them. Events usually don't prescribe what the listener has to do upon receiving them. A command however has this notion. So what I would have done here (just mentioning my viewpoint here!) is to call it `ConsensusWorkerCommand` to indicate what the receiver is supposed to do, and also include the instruction as an imperative to the enum variants, so:
`ConfirmMilestone`, `FetchBalance`, `FetchOutput`, `FetchOutputs`.

thibault-martinez: I changed to Command and the new names. Regarding the 1st comment, I think this is consistent with the rest of the codebase like that, I think it's easier to follow if the Event/Command type is together with the worker that consumes them.

Alex6323: alright 👌

Alex6323 (dismissed):

samuel-rufi (dismissed):

Adam-Gleave (dismissed):

PR #668, opened 2021-07-15
Add the plugin system

pvdrz:
- [x] Figure out how to choose each plugin's address/port.
    - [x] Implement a handshake mechanism using `stdio`.
    - [x] Ensure that the handshake is only done after the gRPC server for the plugin is up.
    - [x] Let the `PluginHandler` spawn the plugin and do the handshake directly.
    - [x] Clean the implementation.
    - [x] Implement proper error handling.
- [x] Use logs instead of println.
- [x] Remove sleeps.
- [x] Implement proper error handling.
- [x] Add plugin hotloading.
- [x] Use gRPC calls with streams instead of single event gRPC calls.
- [x] Document public types.
- [ ] Write spec.

cc @thibault-martinez


bee-node/src/main.rs
```diff
@@ -27,5 +36,30 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         return Ok(());
     }
```
samuel-rufi: Would it make sense to move that code to the `/examples` dir?

bee-plugin/CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
```
samuel-rufi: Should there be an actual description of changes?

bee-plugin/examples/counter.rs
```diff
@@ -0,0 +1,61 @@
...
+    async fn process_dummy_event(&self, _event: DummyEvent) {
+        self.increase();
+    }
+}
+
+#[tokio::main(flavor = "multi_thread")]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let counter = Counter::new();
+
+    serve_plugin(counter).await?;
```
samuel-rufi: wow cool stuff :D seems quite simple to use

bee-plugin/src/event.rs
```diff
@@ -0,0 +1,34 @@
...
+//! Types for events.
+
+pub use crate::grpc::{DummyEvent, SillyEvent};
+
+use std::convert::TryFrom;
+
+/// Identifier for each event type/
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+#[repr(u8)]
+pub enum EventId {
```
samuel-rufi: These are the events from the examples, we can't remove them here as the examples then won't work anymore right? Also for better understanding: all events need to be hard-coded in this enum right?

bee-plugin/src/handshake.rs
```diff
@@ -0,0 +1,86 @@
...
+        let address_chunk = chunks.next().ok_or(InvalidHandshake::MissingAddress)?;
+
+        let address: SocketAddr = address_chunk.parse()?;
+
+        let name = chunks.next().ok_or(InvalidHandshake::MissingName)?.to_string();
+
+        let mut event_ids = vec![];
+
+        for chunk in chunks {
+            let raw_id: u8 = chunk.parse().unwrap();
```
samuel-rufi: uncommented/unhandled `unwrap()`

bee-plugin/src/handler.rs
```diff
@@ -0,0 +1,215 @@
...
+        command.kill_on_drop(true).stdout(Stdio::piped()).stderr(Stdio::piped());
+
+        log::info!(
+            "spawning command `{:?}` for the plugin with ID `{:?}`",
+            command,
+            plugin_id
+        );
+        let mut process = command.spawn()?;
+
+        let stderr = BufReader::new(process.stderr.take().unwrap());
```
samuel-rufi: uncommented/unhandled `unwrap()`

bee-plugin/src/handler.rs
```diff
@@ -0,0 +1,215 @@
...
+        let mut stdout = BufReader::new(process.stdout.take().unwrap());
```
samuel-rufi: uncommented/unhandled `unwrap()`

bee-plugin/src/handler.rs
```diff
@@ -0,0 +1,215 @@
...
+
+        let name = format!("{}-{}", handshake_info.name, plugin_id.0);
+        let target = format!("plugins::{}", name);
+        let stdio_task = tokio::spawn(async move {
+            let mut stdout_lines = stdout.lines();
+            let mut stderr_lines = stderr.lines();
+
+            loop {
+                tokio::select! {
+                    res = stdout_lines.next_line() => match res.unwrap() {
```
samuel-rufi: uncommented/unhandled `unwrap()`

bee-plugin/src/handler.rs
```diff
@@ -0,0 +1,215 @@
...
+                        Some(line) => {
+                            log::info!(target: &target, "{}", line);
+                        },
+                        None => break,
+                    },
+                    res = stderr_lines.next_line() => match res.unwrap() {
```
samuel-rufi: uncommented/unhandled `unwrap()`

samuel-rufi (changes requested):

bee-plugin/src/event.rs
pvdrz: Yes, this enum must be kept in sync with the `.proto` file.

bee-plugin/src/handshake.rs
pvdrz: unhandled! will fix

bee-plugin/src/handler.rs
pvdrz: undocumented! will comment

bee-plugin/CHANGELOG.md
pvdrz: Hmmm I'm not sure given that the crate didn't exist before.

bee-plugin/examples/counter.rs
pvdrz: I know right? :D

bee-plugin/src/event.rs
pvdrz: Actually I'm not sure if I want to remove them or not because otherwise the examples won't work.

Adam-Gleave: It feels a bit odd to me that examples like `DummyEvent` and `SillyEvent` are hardcoded here. If a user wants to write a new plugin, do they also need to rebuild this source too? (If that's a possible use-case).

bee-plugin/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+edition = "2018"
+description = "The bee node API for plugins"
+readme = "README.md"
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "tangle", "bee", "framework", "plugin" ]
+homepage = "https://www.iota.org"
+
+[build-dependencies]
+tonic-build = { version = "0.5", default-features = false, features = ["transport", "prost"]}
```
Adam-Gleave:
```suggestion
tonic-build = { version = "0.5", default-features = false, features = [ "transport", "prost" ] }
```

bee-plugin/src/event.rs
pvdrz: Just to be clear. Those aren't events that I made up for the examples themselves. I made them because there is no events at the moment :trollface:

For me the idea is that this `EventId` and anything in the `.proto` file can only change between versions. So once we have real events we'd have to update the examples.

bee-plugin/src/hotloading.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Holdloading related utilities.
```
Alex6323: "Holdloading" Tbh, one of the funniest typos I've read in a while ;)

bee-plugin/src/hotloading.rs
```diff
@@ -0,0 +1,103 @@
...
+        Self {
+            directory: directory.as_ref().to_owned(),
+            plugins_info: HashMap::new(),
+            manager: PluginManager::new(event_bus),
+        }
+    }
+
+    /// Watches for changes in a directory and loads new plugins.
+    ///
+    /// This is done following these rules:
```
Alex6323: typo: This is done **by** following these rules:

bee-plugin/src/hotloading.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// - If a file is created, it will be executed and loaded as a plugin.
```
Alex6323: suggestion: consider swapping terms "executed" and "loaded" (which seems to be the more logical order of things happening)

bee-plugin/src/hotloading.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// - If a file is removed, the process for it will be terminated and the plugin will be
+    /// unloaded.
+    /// - If a file is modified, it will behave as if the file was removed and created in
+    /// succession.
+    pub async fn run(mut self) -> Result<(), PluginError> {
+        loop {
+            let mut dir: _ = tokio::fs::read_dir(&self.directory).await?;
```
Alex6323: can it not infer the type without the underscore?

bee-plugin/src/hotloading.rs
```diff
@@ -0,0 +1,103 @@
...
+                        to_delete.push(path.clone());
+                        self.manager.unload_plugin(info.id).await?;
+                    }
+                }
+            }
+
+            // remove info for files that no longer exist
+            for path in to_delete {
+                self.plugins_info.remove(&path);
+            }
```
Alex6323: I feel like the synchronization step (between the actual directory and the plugins_info state) could be factored out into its own helper function (to make the code a bit more self-describing).

bee-plugin/src/hotloading.rs
```diff
@@ -0,0 +1,103 @@
...
+
+            // Load plugins whose files did not exist before.
+            for (path, last_write) in last_writes {
+                let command = Command::new(&path);
+                let id = self.manager.load_plugin(command).await?;
+                self.plugins_info.insert(path, PluginInfo { id, last_write });
+            }
+
+            sleep(Duration::from_secs(1)).await;
```
Alex6323: I suggest we could use a constant here  - in milliseconds even - for adjustability and readability, maybe something like PLUGIN_CHECK_INTERVAL_MILLIS or something better you might come up with :)

bee-plugin/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! The bee node plugin system.
+
+#![warn(missing_docs)]
```
Alex6323: Can we not just deny missing docs?

bee-plugin/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+mod error;
+pub mod event;
+pub mod plugin;
+
+pub use error::PluginError;
+pub use manager::PluginManager;
+
+use std::any::TypeId;
+
+/// An unique identifier for `EventBus` callbacks.
```
Alex6323: typo: "A unique ..." (actually had to look up for the exact reason, and found this explanation: https://wordwhirled.blogspot.com/2005/12/unique-or-unique.html)

bee-plugin/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+impl From<PluginId> for UniqueId {
+    fn from(id: PluginId) -> Self {
+        Self::Plugin(id)
+    }
+}
+
+/// An unique identifier for each plugin.
```
Alex6323: same typo as above ;) (don't judge me for my german-ness, haha)

bee-plugin/src/manager.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        self.count += 1;
+
+        Ok(plugin_id)
+    }
+
+    /// Unloads a plugin with the specified identifier.
+    pub async fn unload_plugin(&mut self, id: PluginId) -> Result<(), PluginError> {
+        if let Some(handler) = self.handlers.remove(&id) {
+            log::info!("shutting down the plugin with ID `{:?}`", id);
```
Alex6323: can we also print the plugin name in these logs?

bee-plugin/src/error.rs
```diff
@@ -0,0 +1,23 @@
...
+    #[error("IO error for children process: {0}")]
+    Io(#[from] std::io::Error),
+    /// Transport errors between the node and the plugin.
+    #[error("gRPC transport error: {0}")]
+    TransportError(#[from] tonic::transport::Error),
+    /// Status errors from the gRPC plugin server.
+    #[error("gRPC status error: {0}")]
+    StatusError(#[from] tonic::Status),
+    /// Invalid handshake info errors.
+    #[error("invalid handshake info: {0}")]
```
Alex6323: nitpick-factor 1000: can we drop the word 'info' (as this makes me (personally) expect a type `InvalidHandshakeInfo` which may represent something different than an invalid handshake?

bee-plugin/README.md
```diff
@@ -0,0 +1 @@
+# bee-plugin
```
Alex6323: I suggest simply c&p'ing the `lib` or the `Cargo.toml` description here again.

bee-plugin/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+authors = [ "IOTA Stiftung" ]
+edition = "2018"
+description = "The bee node API for plugins"
+readme = "README.md"
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "tangle", "bee", "framework", "plugin" ]
+homepage = "https://www.iota.org"
+
+[build-dependencies]
```
Alex6323: maybe also too nitpicky (I honestly apologise if so), but should we stick to certain Cargo.toml dependency order? Currently I favor:
(1) deps > (2) dev-deps > (3) build-deps
but I am open to basically everything else as long as we all stick to it :)

bee-plugin/proto/plugin.proto
```diff
@@ -0,0 +1,23 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
```
Alex6323: Could we add a line between `syntax` and `package`? (just a personal preference, though ... ignore if you want)

bee-plugin/proto/plugin.proto
```diff
@@ -0,0 +1,23 @@
...
+package plugin;
+
+service Plugin {
+    rpc Shutdown(ShutdownRequest) returns (ShutdownReply);
+    rpc ProcessDummyEvent(stream DummyEvent) returns (ProcessReply);
+    rpc ProcessSillyEvent(stream SillyEvent) returns (ProcessReply);
```
Alex6323: what's the deal with `DummyEvent` and `SillyEvent`. I don't get it. Feels like something like we should not have in a production setting, so I guess it's gonna be changed?

bee-plugin/src/event.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Types for events.
+
+pub use crate::grpc::{DummyEvent, SillyEvent};
+
+use std::convert::TryFrom;
+
+/// Identifier for each event type/
```
Alex6323: typo at the end

bee-plugin/src/event.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+#[repr(u8)]
+pub enum EventId {
+    /// Identifier for `DummyEvent`.
+    Dummy = 0,
+    /// Identifier for `SillyEvent`.
+    Silly = 1,
+}
+
+impl TryFrom<u8> for EventId {
```
Alex6323: Depending on how long this enum will get eventually it might make sense to derive `num::FromPrimitive` for this enum.

bee-plugin/src/handler.rs
```diff
@@ -0,0 +1,218 @@
...
+use tonic::{transport::Channel, Request};
+
+use std::{
+    any::type_name,
+    collections::{hash_map::Entry, HashMap},
+    process::Stdio,
+    time::Duration,
+};
+
+macro_rules! spawn_streamers {
```
Alex6323: I wonder if that macro would better live in the `streamer` module? (I think from a user pov one would expect it to be there idk)

bee-plugin/src/handler.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// Creates a new plugin handler from a process running the plugin logic.
+    pub(crate) async fn new(
+        plugin_id: PluginId,
+        mut command: Command,
+        bus: &EventBus<'static, UniqueId>,
+    ) -> Result<Self, PluginError> {
+        command.kill_on_drop(true).stdout(Stdio::piped()).stderr(Stdio::piped());
+
+        log::info!(
+            "spawning command `{:?}` for the plugin with ID `{:?}`",
```
Alex6323: I suggest printing also the plugin name whenever we can, because the ID is not fixed for a particular plugin (if I add a new plugin to the directory ids potentially change which could make debugging harder)

bee-plugin/src/handler.rs
```diff
@@ -0,0 +1,218 @@
...
+            command,
+            plugin_id
+        );
+        let mut process = command.spawn()?;
```
Alex6323: Do you think we need a security model for plugins? I was thinking about running plugins in a sandbox environment only. We could probably use Docker for this?

bee-plugin/src/handler.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        let address = format!("http://{}/", handshake_info.address);
+        log::info!("connecting to the `{}` plugin at `{}`", name, address);
+        let client = async {
+            let mut count = 0;
+            loop {
+                match PluginClient::connect(address.clone()).await {
+                    Ok(client) => break Ok(client),
+                    Err(err) => {
+                        log::warn!("connection to the `{}` plugin failed: {}", name, err);
```
Alex6323: What's the scenario in which a plugin cannot be reached on the first attempt, but after some time becomes reachable again? (I should just copy this back-off logic for the network where it's still a TODO, haha)

bee-plugin/src/streamer.rs
```diff
@@ -0,0 +1,49 @@
...
+use tokio::{
+    select,
+    sync::{mpsc::UnboundedReceiver, oneshot::Receiver},
+};
+
+use tokio_stream::wrappers::UnboundedReceiverStream;
+
+use tonic::transport::Channel;
+
+pub(crate) struct PluginStreamer<T> {
```
Alex6323: I could be helpful to explain the `PluginStreamer` abstraction a bit in a few lines of doc comments. It's not obvious immediatedly from the type name (to me at least) what its exact job description is.

bee-plugin/src/plugin.rs
```diff
@@ -0,0 +1,77 @@
...
+                    }
+
+                    Ok(Response::new(ProcessReply {}))
+                }
+            )*
+        }
+    };
+}
+
+plugin_trait! {
```
Alex6323: Can we not use dynamic dispatch or even `Any` for the events (as we already do in the event bus)?

bee-plugin/proto/plugin.proto
pvdrz: Well we don't have any events to the date so I added them to at least be able to test something.

bee-plugin/src/event.rs
pvdrz: I think it will be pretty long.

bee-plugin/src/handler.rs
pvdrz: Well sort of because the macro is actually called in this module and this module and all its types are private, so the user doesn't even see the `PluginHandler` values.

bee-plugin/src/handler.rs
pvdrz: here we cannot do it because the name is obtained after the handshake.

bee-plugin/src/handler.rs
pvdrz: i think the user is allowed to use docker somehow if they want by running a process that starts the container and takes it down when killed.

bee-plugin/src/handler.rs
pvdrz: This happens if the plugin sends a valid handshake before starting its gRPC server. Because the node will try to connect to the server stated in the handshake immediately.

bee-plugin/src/plugin.rs
pvdrz: maybe we can but I don't know if that's more newbie friendly because then you have to do a bunch of downcasting and so on to get a concrete event type (there's the performance argument too).

bee-plugin/src/lib.rs
pvdrz: What? english doesn't make any sense. Let's go back to latin... in quadrigis habet commodum alicuius generis quisque discere ea

bee-plugin/src/event.rs
samuel-rufi: So this means, anyone who want to write a node plugin and run it with the node just need to edit the `.proto` file? And then it works without any changes to the code of the node?

bee-plugin/src/handler.rs
samuel-rufi: I thought about the same, i think running plugins in a safe environment makes sense. Similarly as smart contracts are run in a safe environment.

bee-plugin/proto/plugin.proto
samuel-rufi: Related discussion: https://github.com/iotaledger/bee/pull/668#discussion_r679228046

thibault-martinez (changes requested):
Since plugins only make sense for the node, can we do something like this ?
```
bee-node/bee-node
bee-node/bee-plugin
```

pvdrz:
> Since plugins only make sense for the node, can we do something like this ?
> 
> ```
> bee-node/bee-node
> bee-node/bee-plugin
> ```

Sure but plugin authors also use this package, is that OK?

thibault-martinez:
> > Since plugins only make sense for the node, can we do something like this ?
> > ```
> > bee-node/bee-node
> > bee-node/bee-plugin
> > ```
> 
> Sure but plugin authors also use this package, is that OK?

Sure

bee-plugin/src/event.rs
pvdrz: not exactily. It means that anyone who wants to write a plugin can use the `.proto` file as it is (once we have actual events to put in it).

My point is that we could leave the `.proto` file without events but then there would be no examples at all until we add events to the `.proto` file.

pvdrz: OK I am using real events now that we are closer to have a network able to receive messages and we have a way to parse them.

bee-node/bee-plugin/src/streamer.rs
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+}
+
+macro_rules! impl_streamer {
+    ($($event_ty:ty => $method_name:ident),*) => {
+        $(
+            impl PluginStreamer<$event_ty> {
+                pub(crate) async fn run(mut self) {
+                    select! {
```
thibault-martinez, line 35: Should we bring back ShutdownStream ?

bee-node/bee-plugin/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+
+pub use error::PluginError;
+pub use manager::PluginManager;
+pub use plugin::{serve_plugin, Plugin};
+
+use std::any::TypeId;
+
+/// A unique identifier for `EventBus` callbacks.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum UniqueId {
```
thibault-martinez: I still think that this type could be useful in the bee-event-bus crate. With variants `Type` and `Object` maybe.

bee-node/bee-plugin/src/event.rs
```diff
@@ -0,0 +1,37 @@
...
+pub enum EventId {
+    /// Identifier for [`MessageParsedEvent`].
+    MessageParsed = 0,
+    /// Identifier for [`ParsingFailedEvent`].
+    ParsingFailed = 1,
+    /// Identifier for [`MessageRejectedEvent`].
+    MessageRejected = 2,
+}
+
+impl TryFrom<u8> for EventId {
```
thibault-martinez: Is there a crate to derive this maybe ? Can get quite error prone when the enum gets big

bee-node/bee-plugin/src/streamer.rs
pvdrz: Most likely

bee-node/bee-plugin/src/event.rs
pvdrz: Alex mentioned something about the num crate. I'll check

samuel-rufi (dismissed):

bee-node/bee-plugin/src/event.rs
Adam-Gleave: I think this might be what you're looking for:
https://docs.rs/num_enum/0.5.3/num_enum/

Adam-Gleave (dismissed):

bee-node/bee-plugin/src/event.rs
pvdrz: I don't know why github is not letting me mark this as solved, but it is. I added a macro to do this

bee-node/bee-plugin/src/handshake.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    pub(crate) fn parse(buf: &str) -> Result<Self, InvalidHandshake> {
+        let mut chunks = buf.trim().split('|');
+        let address_chunk = chunks.next().ok_or(InvalidHandshake::MissingAddress)?;
+        let address: SocketAddr = address_chunk.parse()?;
+        let name = chunks.next().ok_or(InvalidHandshake::MissingName)?.to_string();
+        let mut event_ids = vec![];
+
+        for chunk in chunks {
```
thibault-martinez: Can this be reworked in an iter/collect ?

pvdrz: yes

bee-node/bee-plugin/src/hotloader.rs
```diff
@@ -0,0 +1,111 @@
...
+        loop {
+            let mut dir = tokio::fs::read_dir(&self.directory).await?;
+
+            let mut last_writes = HashMap::new();
+
+            while let Some(entry) = dir.next_entry().await? {
+                let path = entry.path();
+                let metadata = entry.metadata().await?;
+
+                if metadata.is_file() {
```
thibault-martinez, line 64: This doesn't seem sufficient to me, it will try to execute regular files and crash

thibault-martinez (dismissed):

bee-node/bee-plugin/src/hotloader.rs
pvdrz: yeah but it will be a child process crashing so we will get a warning in the logs and that's it.

Edit: It actually crashed but I solved it with the last commit.

bee-node/bee-plugin/src/hotloader.rs
```diff
@@ -0,0 +1,128 @@
...
+            self.plugins_info.remove(&path);
+        }
+        for (path, last_write) in to_load {
+            let command = Command::new(&path);
+            match self.manager.load(command).await {
+                Ok(id) => {
+                    self.plugins_info.insert(path, PluginInfo { id, last_write });
+                }
+                Err(err) => {
+                    log::warn!("could not load plugin from path `{}`: {}", path.display(), err);
```
thibault-martinez:
```suggestion
                    log::warn!("could not load plugin from path \"{}\": {}", path.display(), err);
```

thibault-martinez (approved):

PR #674, opened 2021-07-23
Add `bee-network` crate

Alex6323:
Network layer (on top of `iotaledger/backstage`) for the GoShimmer testnet.


bee-network/src/config.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.infos.get(ip_addr)
+    }
+
+    /// Adds a new static peer.
+    pub fn add(&mut self, _info: ManualPeerInfo) -> bool {
+        todo!("add manual peers")
+    }
+
+    /// Iterates over all manual peers.
+    pub fn iter(&self) -> collections::hash_map::Iter<IpAddr, ManualPeerInfo> {
```
pvdrz: do you need this concrete type somewhere? Otherwise you can use `impl Iterator<Item=(&IpAddr, &ManualPeerInfo)>`

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,374 @@
...
+pub fn handshake(
+    stream: TcpStream,
+    socket_addr: SocketAddr,
+    local_id: &LocalIdentity,
+    direction: Direction,
+    peer_info: ManualPeerInfo,
+) -> Result<(BufReader<TcpStream>, BufWriter<TcpStream>, Identity, Alias), HandshakeError> {
+    println!("Handshaking with {}...", socket_addr);
+
+    let mut writer = BufWriter::new(stream.try_clone().map_err(|e| HandshakeError::Io(e))?);
```
pvdrz: you can do `.map_err(HandshakeError:Io)` directly

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,374 @@
...
+    to: IpAddr,
+    local_id: &LocalIdentity,
+) -> Result<BytesMut, HandshakeError> {
+    let ty = PacketType::Handshake;
+
+    let data = HandshakeRequest::new(to).protobuf()?;
+    let signature = local_id.sign(&data).to_bytes();
+
+    let packet = Packet::new(ty, &data, local_id.public_key().as_ref(), &signature);
+    let packet_bytes = packet.protobuf().map_err(|e| HandshakeError::Encode(e))?;
```
pvdrz: Same

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    writer.write_all(&packet_bytes).map_err(|e| HandshakeError::Io(e))?;
```
pvdrz: Same

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,374 @@
...
+    Ok(())
+}
+
+fn await_request(
+    reader: &mut BufReader<TcpStream>,
+    writer: &mut BufWriter<TcpStream>,
+    local_id: &LocalIdentity,
+    _peer_info: &ManualPeerInfo,
+) -> Result<Identity, HandshakeError> {
+    let mut buf = vec![0; MAX_HANDSHAKE_PACKET_SIZE];
```
pvdrz, line 111: Can't this be an array or a boxed slice?

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,374 @@
...
+    let mut pk = [0u8; 32];
+    pk.copy_from_slice(&peer_public_key[..32]);
+    let peer_public_key = ed25519::PublicKey::from_compressed_bytes(pk).map_err(|e| HandshakeError::PublicKey(e))?;
+    let peer_identity = Identity::from_public_key(peer_public_key);
+
+    Ok(peer_identity)
+}
+
+/// Errors, that may occur during the handshake process.
+#[derive(Debug)]
```
pvdrz, line 212: maybe you can use `thiserror` and avoid all the `map_err(HandshakeErrror::Variant)`

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,374 @@
...
+    inner: proto::HandshakeRequest,
+}
+
+impl HandshakeRequest {
+    /// Creates a new handshake request.
+    pub fn new(to: IpAddr) -> Self {
+        let inner = proto::HandshakeRequest {
+            version: VERSION,
+            to: to.to_string(),
+            timestamp: SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs() as i64,
```
pvdrz: oh yes... the ol' "time went backwards" unwrap. :trollface:

pvdrz (dismissed):

bee-network/src/config.rs
Alex6323: you're right. I probably don't :smiley:

bee-network/src/handshake.rs
Alex6323: You're right. I keep forgetting about this convenience stuff :joy:

bee-network/src/handshake.rs
Alex6323: yup, definitely switching to `thiserror` eventually :thumbsup:

bee-network/src/handshake.rs
Alex6323: indeed :joy:

pvdrz (dismissed):

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 members = [
   "bee-common/bee-event-bus",
   "bee-common/bee-logger",
+  "bee-network",
```
Adam-Gleave: Order! :trollface:

bee-network/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+/// Stores connection and other information about a manual peer.
+#[derive(Debug, Clone)]
+pub struct ManualPeerInfo {
+    /// The identity of the peer.
+    pub identity: Identity,
+    /// The address of the peer.
+    pub address: SocketAddr,
+    /// A human friendly identifier of the peer.
+    pub alias: String,
+    dialer: bool,
```
Adam-Gleave: I know there doesn't need to be a comment here, but this is irrationally annoying :joy:

Adam-Gleave: Perhaps `is_dialer` is a better name.

bee-network/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+
+            let peer_info = ManualPeerInfo {
+                identity,
+                address,
+                alias,
+                dialer,
+            };
+
+            if infos.contains_key(&ip) {
+                unimplemented!("multiple instances with same ip address");
```
Adam-Gleave, line 153: Does `unimplemented!` make sense here? It feels like it should just be a regular `panic!`.

bee-network/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /// Returns a [`ManualPeerInfo`] associated with a particular [`IpAddr`].
+    pub fn get(&self, ip_addr: &IpAddr) -> Option<&ManualPeerInfo> {
+        self.infos.get(ip_addr)
+    }
+
+    /// Adds a new static peer.
+    pub fn add(&mut self, _info: ManualPeerInfo) -> bool {
+        todo!("add manual peers")
```
Adam-Gleave, line 169: I didn't know this was a thing, nice!

bee-network/src/conn.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub fn new() -> Self {
+        Self(Arc::new(RwLock::new(HashSet::new())))
+    }
+
+    pub fn add(&self, item: T) {
+        self.0.write().expect("error getting the lock").insert(item);
+    }
+
+    // TODO: update this list on connection loss
+    pub fn _remove(&self, item: T) -> bool {
```
Adam-Gleave: Why is this prefixed with an underscore?

bee-network/src/identity.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl LocalIdentity {
+    /// Creates a new local identity.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Creates a local identity from a 'base58' encoded ED25519 private key.
+    pub fn from_bs58_secret_key_str(sk: &str) -> Self {
+        let sk = bs58::decode(sk).into_vec().expect("error restoring secret key");
```
Adam-Gleave, line 30: The names `sk` and `secret_key` are a little confusing here.

bee-network/src/config.rs
Alex6323: added docs

Alex6323: sure, why not

bee-network/src/config.rs
Alex6323: I think the notion of `unimplemented` is that we want to make the "user" aware that we currently do not intend to implement this (unless this becomes a requested feature in the future). If something is just not implemented due to "priorities" (but is considered to be in the needs-to-be-done-before-we-release), then I'ld use the `todo!` macro. Because I wanted to transport the context (that we currently do not intend to implement this "feature"), I think it makes sense how it is atm, but if others lean more towards your pov I am happy to change.

bee-network/src/config.rs
Alex6323: wdym here, Adam? :smile:

bee-network/src/conn.rs
Alex6323: because it's dead code atm, but will be used as soon as we also want to remove some peers. That's not the case atm, because we have no autopeering. I removed the underscore and allowed dead code instead.

bee-network/src/identity.rs
Alex6323: Do you want me to use `private_key` instead? (I think the only reason why the term `secret_key` is also used very commonly is because it shortens to `sk` which allows to unambiguously assign `pk` to "public key").

thibault-martinez (changes requested):
*Don't mind me*

bee-network/src/identity.rs
Adam-Gleave: I think I would probably use `private_key` -- it's also what's in the the  doc-comment, so I think it makes the most sense.

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,395 @@
...
+    pk.copy_from_slice(&peer_public_key[..32]);
+    let peer_public_key = ed25519::PublicKey::try_from_bytes(pk).map_err(HandshakeError::PublicKey)?;
+    let peer_identity = Identity::from_public_key(peer_public_key);
+
+    Ok(peer_identity)
+}
+
+/// Errors, that may occur during the handshake process.
+#[derive(Debug)]
+pub enum HandshakeError {
```
Adam-Gleave, line 213: What happened to `thiserror`? :trollface: (I'm just being annoying, I'm not bothered)

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,395 @@
...
+
+impl HandshakeRequest {
+    /// Creates a new handshake request.
+    pub fn new(to: IpAddr) -> Self {
+        let inner = proto::HandshakeRequest {
+            version: VERSION,
+            to: to.to_string(),
+            timestamp: SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("Wake up, NEO. Follow the White Rabbit.")
```
Adam-Gleave, line 261: 🤔

bee-network/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Utilities to establish and maintain network connections with peers.
+
+#![deny(missing_docs)]
```
Adam-Gleave, line 6:
```suggestion
#![deny(warnings, missing_docs)]
```

bee-network/src/peer.rs
```diff
@@ -0,0 +1,294 @@
...
+
+                let mut position = 0;
+                let mut messages = Vec::with_capacity(16);
+
+                while position < num_received {
+                    // Determine the length of the next message within this batch
+                    let mut msg_len_buf = [0u8; 4];
+                    msg_len_buf.copy_from_slice(&self.buffer[position..position + 4]);
+                    let msg_len = Buf::get_u32(&mut &msg_len_buf[..]) as usize - 1;
+                    // println!("Message length (excl. type specifier byte): {}.", msg_len);
```
Adam-Gleave, line 98: Does this need removing?

bee-network/src/peer.rs
```diff
@@ -0,0 +1,294 @@
...
+
+                    // Determine the message type
+                    let msg_type: MessageType =
+                        num::FromPrimitive::from_u8(self.buffer[position + 4]).ok_or(Error::UnknownMessageType)?;
+                    // println!("Message type: {:?}.", msg_type);
```
Adam-Gleave, line 103: There's a few commented `println!`s in this file.

bee-node/bee-node/src/main.rs
```diff
@@ -1,31 +1,112 @@
...
+        let config = Config::new(port, local_id, "./peers.yaml").map_err(ActorError::exit)?;
+
+        rt.add_resource(config).await;
+
+        rt.start(Some("network".into()), NetworkWorker::new()).await.unwrap();
+
+        while let Some(event) = rt.inbox_mut().next().await {
+            match event {
+                BeeSupervisorEvent::Eol | BeeSupervisorEvent::Report => {
+                    // FIXME: handle network status report events.
```
Adam-Gleave: Does this need implementing?

bee-network/src/handshake.rs
Alex6323: Have you seen the Matrix 4 - Resurrection Trailer?

Adam-Gleave: Yeah, it actually looks like it could be good!

bee-network/src/lib.rs
Alex6323: Why is that a good idea exactly? This way builds will possibly fail with a new compiler version (where things get slowly deprecated and become warnings first)

bee-network/src/peer.rs
Alex6323: definitely :+1:

bee-node/bee-node/src/main.rs
Alex6323: @pvdrz Wdyt?

bee-network/src/handshake.rs
Alex6323: jokes aside....the typical message there is "Clocks went backwards" or similar. 

The chance of that happening is probably comparable to the next iteration of the Chosen One emerging :trollface: .

bee-network/src/peer.rs
Alex6323: that's right...I'll remove or replace them with proper logging.

bee-network/src/lib.rs
thibault-martinez: I am fairly sure you were the one proposing that in the first place Alex, a couple of months ago 😄 
Our CI already denies warnings anyway so it's about having a local consistency.

bee-node/bee-node/src/main.rs
```diff
@@ -1,31 +1,112 @@
...
+            LocalIdentity::new()
+        };
+
+        log::info!("node local id: {:?}", local_id);
+
+        let config = Config::new(port, local_id, "./peers.yaml").map_err(ActorError::exit)?;
+
+        rt.add_resource(config).await;
+
+        rt.start(Some("network".into()), NetworkWorker::new()).await.unwrap();
```
samuel-rufi: unhandled/uncommented unwrap()

bee-network/src/conn.rs
```diff
@@ -0,0 +1,39 @@
...
+impl<T: Eq + Hash> FilterList<T> {
+    pub fn new() -> Self {
+        Self(Arc::new(RwLock::new(HashSet::new())))
+    }
+
+    pub fn add(&self, item: T) {
+        self.0.write().expect("error getting the lock").insert(item);
+    }
+
+    // TODO: update this list on connection loss
```
samuel-rufi, line 30: will this be handled with a future PR?

bee-network/src/lib.rs
Alex6323: I was once in favor of doing this, true, but then I learned about why that's a bad idea in general and the argument is pretty sound. Is that only a few months ago?? I feel like I am against it much longer now, but I might be wrong :smile: 

I don't understand what the CI has anything to do with it. If it would only be us and our CI compiling the code I wouldn't mind having the flag set in the code at all. I am more worried about people, that just updated to a new compiler version, cannot compile their version of the repo anymore, unless they either downgrade the compiler again, fix the warning themselves or remove the `deny` attribute, or wait until we fixed the warning for the new compiler, so that they can pull the fix.

bee-network/src/backstage/network.rs
```diff
@@ -0,0 +1,76 @@
...
+                    let id = info.id();
+
+                    let reader = PeerReaderWorker::new(peer.reader, info.clone());
+                    let writer = PeerWriterWorker::new(peer.writer, info);
+
+                    rt.start(Some(format!("{}_reader", id)), reader).await?;
+                    rt.start(Some(format!("{}_writer", id)), writer).await?;
+                }
+                NetworkEvent::PeerWorkerEol | NetworkEvent::PeerWorkerReport => {
+                    // FIXME: handle status report for peers
```
samuel-rufi: can this be empty for now? maybe add "TODO" so some IDEs can highlight it

bee-network/src/handshake.rs
Alex6323: on it ... finally :trollface:

bee-network/src/config.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Whether the peer is supposed to be the initiator of a connection.
+    pub fn is_dialer(&self) -> bool {
+        self.is_dialer
+    }
+}
+
+/// Error returned when the peer configuration file could not be loaded.
+#[derive(Debug, Error)]
+pub enum ManualPeerConfigError {
+    /// The YAMl file has an invalid layout.
```
samuel-rufi: "YAML"

bee-network/src/config.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        if docs.len() != 1 {
+            return Err(ManualPeerConfigError::InvalidConfig("expected a single document"));
+        }
+
+        let doc = docs.remove(0);
+
+        let peers_config = doc
+            .into_vec()
+            .ok_or(ManualPeerConfigError::InvalidConfig("peers yaml is not an array"))?;
```
samuel-rufi: uppercase "YAML" ?

bee-network/src/config.rs
```diff
@@ -0,0 +1,176 @@
...
+        let mut infos = HashMap::with_capacity(peers_config.len());
+
+        for m in peers_config {
+            let hm = m
+                .into_hash()
+                .ok_or(ManualPeerConfigError::InvalidConfig("invalid yaml file"))?;
```
samuel-rufi: uppercase "YAML" ?

bee-network/src/event.rs
```diff
@@ -0,0 +1,37 @@
...
+
+//! A module that deals with events published by the network (layer).
+
+use crate::peer::ConnectedPeer;
+
+/// Represents a network event.
+pub enum NetworkEvent {
+    /// Fired when a peer has been successfully connected and handshaked.
+    PeerConnected(ConnectedPeer),
+    /// This is event
```
samuel-rufi: punctuation
also maybe better describe the event?

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,395 @@
...
+    /// Wrong packet type.
+    PacketType(io::Error),
+    /// The peer send a packet which is too large.
+    PacketSizeMismatch {
+        /// The received packet size.
+        received: usize,
+        /// The maximum allowed packet size.
+        max_allowed: usize,
+    },
+    /// TODO
```
samuel-rufi, line 244: what does this TODO relate to?

bee-network/src/message.rs
```diff
@@ -0,0 +1,101 @@
...
+use std::fmt;
+
+pub(crate) struct Message(proto::Message);
+
+/// The type of a message.
+#[derive(Clone, Copy, Debug, FromPrimitive)]
+#[repr(u8)]
+#[non_exhaustive]
+pub enum MessageType {
+    /// Send a message
```
samuel-rufi, line 21: punctuation

bee-network/src/network.rs
```diff
@@ -0,0 +1,181 @@
...
+    on_event: impl Fn(NetworkEvent),
+    local_id: LocalIdentity,
+    manual_peer_config: ManualPeerConfig,
+    connected_list: ConnectedList,
+) {
+    loop {
+        let result = server.accept().await;
+        match result {
+            Ok((tcp_stream, socket_addr)) => {
+                // let conn_id = NUM_CONNECTIONS.fetch_add(1, Ordering::Relaxed);
```
samuel-rufi, line 86: unused code?

bee-network/src/network.rs
```diff
@@ -0,0 +1,181 @@
...
+            // (b) the peer is supposed to dial *us*.
+            if connected_list.contains(peer_info.address) || peer_info.is_dialer() {
+                continue;
+            } else {
+                let socket_addr = peer_info.address;
+                let result = TcpStream::connect(socket_addr).await;
+
+                match result {
+                    Ok(tcp_stream) => {
+                        // let conn_id = NUM_CONNECTIONS.fetch_add(1, Ordering::Relaxed);
```
samuel-rufi, line 147: unused code?

bee-network/src/handshake.rs
Alex6323: @Adam-Gleave Unfortunately I think Matrix 4 will be pretty bad, but I hope I am wrong.

samuel-rufi (changes requested):

bee-node/bee-node/src/main.rs
pvdrz: yes but is not critical. Those events are sent if the `NetworkWorker` dies or changes of status. We can use them to launch the `NetworkWorker` again if required.

bee-network/src/backstage/network.rs
pvdrz: My editor highlights them :trollface: but sure `TODO` is fine too

bee-network/src/event.rs
pvdrz: This is what happens when you add random docs to suppress the warnings hahaha

bee-node/bee-node/src/main.rs
louaykamel: Ideally, both of them should be merged into one variant.

bee-network/src/handshake.rs
Adam-Gleave: I'm almost certain it will be 😢

bee-node/config.toml
```diff
@@ -0,0 +1,16 @@
+# Duplicate this file to config.toml to use it
```
thibault-martinez, line 1: This file shouldn't be here, remove please

thibault-martinez (changes requested):

bee-network/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+edition = "2018"
+description = "Utilities to establish and maintain network connections with peers"
+readme = "README.md"
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "tangle", "bee", "framework", "node" ]
+homepage = "https://www.iota.org"
+
+[lib]
+path = "src/lib.rs"
```
thibault-martinez, line 14: Can we remove this please ? It's a bit inconsistent as we never specify it.

bee-network/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+
+[dependencies]
+async-trait = { version = "0.1", default-features = false }
+backstage = { git = "https://github.com/iotaledger/backstage", default-features = false, branch = "stable", optional = true }
+base64 = { version = "0.13", default-features = false }
+bs58 = { version = "0.4", default-features = false, features = [ "alloc" ] }
+bytes = { version = "1.0", default-features = false }
+iota-crypto = { version = "0.7.0", default-features = false, features = ["ed25519", "random", "sha" ] }
```
thibault-martinez, line 22:
```suggestion
iota-crypto = { version = "0.7.0", default-features = false, features = [ "ed25519", "random", "sha" ] }
```

bee-network/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+log = { version = "0.4", default-features = false }
+num = { version = "0.4", default-features = false }
+num-derive = { version = "0.3", default-features = false }
+num-traits = { version = "0.2", default-features = false }
+prost = { version = "0.8", default-features = false }
+prost-build = { version = "0.8", default-features = false }
+thiserror = { version = "1.0", default-features = false }
+tokio = { version = "1.5", default-features = false, features = ["sync", "time", "rt"] }
```
thibault-martinez, line 30:
```suggestion
tokio = { version = "1.5", default-features = false, features = [ "sync", "time", "rt" ] }
```

bee-network/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+yaml-rust = { version = "0.4", default-features = false }
+
+[dev-dependencies]
+anyhow = { version = "1.0", default-features = false }
+env_logger = { version = "0.9", default-features = false }
+tokio = { version = "1.5", default-features = false, features = ["macros", "rt-multi-thread", "signal"] }
```
thibault-martinez, line 36:
```suggestion
tokio = { version = "1.5", default-features = false, features = [ "macros", "rt-multi-thread", "signal" ] }
```

bee-network/build.rs
```diff
@@ -0,0 +1,16 @@
...
+use std::io::Result;
+
+fn main() -> Result<()> {
+    prost_build::compile_protos(
+        &[
+            "src/proto/handshake.proto",
+            "src/proto/packet.proto",
+            "src/proto/message.proto",
+        ],
+        &["src/"],
```
thibault-martinez, line 13: What does this mean exactly ? Why src ?

bee-network/build.rs
```diff
@@ -0,0 +1,16 @@
...
+    )?;
+    Ok(())
```
thibault-martinez, line 15:
```suggestion

    Ok(())
```

bee-node/bee-node/Cargo.toml
```diff
@@ -12,7 +12,10 @@ homepage = "https://www.iota.org"
 
 [dependencies]
 bee-logger = { path = "../../bee-common/bee-logger", default-features = false, features = [ "serde" ] }
+bee-network = { path = "../../bee-network", default-features = false, features = ["backstage"] }
```
thibault-martinez:
```suggestion
bee-network = { path = "../../bee-network", default-features = false, features = [ "backstage" ] }
```

bee-node/bee-node/src/cli.rs
```diff
@@ -9,6 +9,10 @@ use structopt::StructOpt;
 /// Defines the node CLI arguments.
 #[derive(Clone, Debug, StructOpt)]
 pub struct NodeCliArgs {
+    #[structopt(short, long, help = "The listening port")]
+    port: u16,
```
thibault-martinez: Not precise enough, we'll end up with many listening ports. Maybe use `gossip_port` instead ?

thibault-martinez (changes requested):

bee-network/build.rs
pvdrz: @Alex6323

Alex6323: from the `prost-build` docs:
```rust
pub fn compile_protos(
    protos: &[impl AsRef<Path>], 
    includes: &[impl AsRef<Path>]
)
```
**includes** - Paths to directories in which to search for imports. Directories are searched in order. The .proto files passed in protos must be found in one of the provided include directories.

thibault-martinez: But does it looks recursively then ? Because the protos are not in `src`. Why don't we pass `src/proto/` ?

bee-network/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "bee-network"
+version = "0.1.0"
+authors = [ "IOTA Stiftung" ]
+edition = "2018"
+description = "Utilities to establish and maintain network connections with peers"
+readme = "README.md"
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "tangle", "bee", "framework", "node" ]
```
thibault-martinez, line 10:
```suggestion
keywords = [ "iota", "tangle", "bee", "framework", "network" ]
```

bee-network/examples/actor.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    startup().await.unwrap();
+}
+
+async fn startup() -> anyhow::Result<()> {
+    std::panic::set_hook(Box::new(|info| {
+        log::error!("{}", info);
+    }));
+
+    // RuntimeScope::<ActorRegistry>::launch(|scope| {
```
thibault-martinez, line 17: What's the status of this example ? If it does nothing let's just remove it.

bee-network/src/backstage/mod.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A bee network layer API using backstage.
```
thibault-martinez, line 4: Maybe a link to backstage ?

bee-network/src/handshake.rs
```diff
@@ -0,0 +1,395 @@
...
+use crypto::signatures::ed25519;
+use prost::{bytes::BytesMut, Message as _};
+
+use std::{
+    fmt,
+    net::{IpAddr, SocketAddr},
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+use tokio::{
```
thibault-martinez, line 22: tokio before std

bee-network/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+#[cfg(feature = "backstage")]
+pub mod backstage;
+pub mod config;
+pub mod event;
+pub mod identity;
+pub mod message;
+pub mod network;
+pub mod peer;
+
+pub use consts::MAX_PACKET_SIZE;
```
thibault-martinez, line 24: Does this actually need to be exported ? Sounds like implementation detail to me.

bee-network/src/network.rs
```diff
@@ -0,0 +1,178 @@
...
+    time::{sleep, Duration},
+};
+
+use std::sync::atomic::AtomicUsize;
+
+static _NUM_CONNECTIONS: AtomicUsize = AtomicUsize::new(0);
+
+const RECONNECT_INTERVAL_SECS: u64 = 30;
+
+use std::io;
```
thibault-martinez, line 27: Merge with the other std import

bee-network/src/peer.rs
```diff
@@ -0,0 +1,294 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A module that deals with peers.
+
+use std::sync::atomic::{AtomicBool, Ordering};
```
thibault-martinez, line 6: After 3rd party imports

bee-node/bee-node/src/main.rs
```diff
@@ -1,31 +1,111 @@
 // Copyright 2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use std::error::Error;
```
thibault-martinez: After bee imports

thibault-martinez (changes requested):

bee-network/build.rs
Alex6323: we could do that and be more specific, but:
1) we don't know if we want to keep the directory structure like that (`src` however will probably forever be correct)
2) we don't know if being more concrete here does help with anything (compilation performance maybe?)
3) I've never seen anything else put there yet (but that doesn't mean anything)

bee-network/Cargo.toml
Alex6323: :+1:

bee-network/src/lib.rs
Alex6323: it's not needed anymore to export it. It was only required in the first impl.

bee-network/src/backstage/peer.rs
```diff
@@ -0,0 +1,105 @@
...
+impl PeerReaderWorker {
+    pub(crate) fn new(reader: PeerReader, info: Arc<PeerInfo>) -> Self {
+        Self { reader, info }
+    }
+}
+
+#[async_trait::async_trait]
+impl<S: SupHandle<Self>> Actor<S> for PeerReaderWorker {
+    type Data = ();
+
```
thibault-martinez, line 28: Do you mind not putting new lines between associated types ? This is inconsistent with the rest of the codebase.

bee-network/src/backstage/peer.rs
```diff
@@ -0,0 +1,105 @@
...
+impl PeerWriterWorker {
+    pub(crate) fn new(writer: PeerWriter, info: Arc<PeerInfo>) -> Self {
+        Self { writer, info }
+    }
+}
+
+#[async_trait::async_trait]
+impl<S: SupHandle<Self>> Actor<S> for PeerWriterWorker {
+    type Data = ();
+
```
thibault-martinez, line 85: Same

thibault-martinez (changes requested):

bee-node/bee-node/src/config.rs
```diff
@@ -30,13 +31,15 @@ pub enum Error {
...
+    pub(crate) network: Option<NetworkConfigBuilder>,
+    pub(crate) manual_peering: Option<ManualPeeringConfigBuilder>,
 }
 
 impl NodeConfigBuilder {
     /// Creates a [`NodeConfigBuilder`] from a config file.
     pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
         match fs::read_to_string(path) {
-            Ok(json) => serde_json::from_str::<Self>(&json).map_err(Error::ConfigFileDeserializationFailed),
+            Ok(json) => { println!("{:?}", json); serde_json::from_str::<Self>(&json).map_err(Error::ConfigFileDeserializationFailed)},
```
thibault-martinez: Remove println

bee-node/bee-node/src/config.rs
```diff
@@ -55,8 +58,17 @@ impl NodeConfigBuilder {
 
     /// Finished the [`NodeConfigBuilder`] into a [`NodeConfig`].
     pub fn finish(self) -> NodeConfig {
+        let network = self.network.expect("missing network configuration").finish();
+        let manual_peering = self.manual_peering.unwrap_or_default().finish(&network.local_id);
+
+        println!("{:?}", network.bind_addr);
+        println!("{:?}", network.local_id);
+        println!("{:?}", manual_peering.len());
```
thibault-martinez: Remove

bee-node/bee-node/src/cli.rs
```diff
@@ -47,4 +51,14 @@ impl NodeCliArgs {
     pub fn version(&self) -> bool {
         self.version
     }
+
+    /// Returns the listening port for gossip.
+    pub fn gossip_port(&self) -> Option<&u16> {
+        self.gossip_port.as_ref()
+    }
+
+    /// Returns the identity of the node if specified.
```
thibault-martinez: Why do we say "if specified" here but not on the port ? I wouldn't mention it as it's an Option tbh

thibault-martinez (changes requested):

bee-network/src/backstage/mod.rs
Alex6323: Done

bee-network/examples/actor.rs
Alex6323: Removed.

bee-network/src/backstage/mod.rs
Alex6323: Done

bee-network/src/handshake.rs
Alex6323: Done

bee-network/src/lib.rs
Alex6323: Removed.

bee-network/src/handshake.rs
Alex6323: Fixed

bee-network/src/backstage/peer.rs
Alex6323: I think Christian doesn't. I'll remove the empty line.

bee-network/src/backstage/peer.rs
Alex6323: Done

bee-node/bee-node/src/config.rs
Alex6323: :facepalm:  thx

bee-network/bee-autopeering/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+readme = "README.md"
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "tangle", "bee", "framework", "autopeering" ]
+homepage = "https://www.iota.org"
+
+[dependencies]
+bee-identity = { path = "../bee-identity" }
+
+serde = "1.0.130"
```
thibault-martinez: Remove patch and add default-features false

bee-network/bee-autopeering/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+
+[dev-dependencies]
```
thibault-martinez: Remove

bee-network/bee-autopeering/build.rs
```diff
@@ -0,0 +1,15 @@
...
+
+use std::io::Result;
+
+fn main() -> Result<()> {
+    // prost_build::compile_protos(
+    //     &[
+    //         "src/proto/handshake.proto",
+    //     ],
+    //     &["src/"],
+    // )?;
```
thibault-martinez: Remove ?

bee-network/bee-autopeering/src/config.rs
```diff
@@ -0,0 +1,33 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Autopeering configuration.
+
+use serde::{Deserialize, Serialize};
+
+/// Manual peer configuration.
```
thibault-martinez: Not manual

bee-network/bee-gossip/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "tangle", "bee", "framework", "gossip" ]
+homepage = "https://www.iota.org"
+
+[dependencies]
+bee-identity = { path = "../bee-identity" }
+bee-task = { path = "../../bee-common/bee-task" }
+bee-manualpeering = { path = "../bee-manualpeering" }
+bee-autopeering = { path = "../bee-autopeering" }
```
thibault-martinez: Order + default features false

bee-network/bee-gossip/src/backstage/network.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::{
+    backstage::peer::{GossipReaderActor, GossipWriterActor},
+    network::Gossip,
+};
+
+use backstage::core::{AbortableUnboundedChannel, Actor, ActorError, ActorResult, Rt, StreamExt, SupHandle};
+
+use std::sync::Arc;
+
+/// A network actor.
```
thibault-martinez: Gossip

bee-network/bee-gossip/src/config.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A module containing types necessary for network (layer) configuration.
+
+use serde::{Deserialize, Serialize};
+
+use std::net::{SocketAddr, ToSocketAddrs};
+
+/// Network configuration.
```
thibault-martinez: Gossip

bee-network/bee-gossip/src/conn.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub fn new() -> Self {
+        Self(Arc::new(RwLock::new(HashSet::new())))
+    }
+
+    pub fn add(&self, item: T) {
+        self.0.write().expect("error getting the lock").insert(item);
+    }
+
+    // TODO: update this list on connection loss
+    #[allow(dead_code)]
```
thibault-martinez, line 31: Why dead code if it's pub ?

bee-network/bee-gossip/src/event.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! A module that deals with events published by the network (layer).
+
+use crate::peer::ConnectedPeer;
+
+/// Represents a network event.
```
thibault-martinez: Gossip

bee-network/bee-gossip/src/handshake.rs
```diff
@@ -0,0 +1,379 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    conn::Direction,
+    consts::{HANDSHAKE_TIMEOUT_SECS, MAX_HANDSHAKE_PACKET_SIZE, VERSION},
+    packet::{packet_hash, Packet, PacketType},
+    proto,
+};
+use bee_identity::identity::{LocalId, PeerId};
```
thibault-martinez, line 10:
```suggestion

use bee_identity::identity::{LocalId, PeerId};
```

bee-network/bee-gossip/src/network.rs
```diff
@@ -0,0 +1,180 @@
...
+    config::GossipConfig,
+    conn::{ConnectedList, Direction},
+    event::GossipEvent,
+    handshake::handshake,
+    peer::ConnectedPeer,
+};
+
+use bee_identity::{config::NetworkIdentityConfig, identity::LocalId};
+use bee_manualpeering::config::ManualPeeringConfig;
+
```
thibault-martinez:
```suggestion
```

bee-node/bee-node/Cargo.toml
```diff
@@ -11,8 +11,14 @@ keywords = [ "iota", "tangle", "bee", "framework", "node" ]
 homepage = "https://www.iota.org"
 
 [dependencies]
+bee-autopeering = { path = "../../bee-network/bee-autopeering" }
```
thibault-martinez: default features false

thibault-martinez (changes requested):

bee-network/bee-gossip/src/event.rs
Alex6323: Changed

bee-network/bee-gossip/src/handshake.rs
Alex6323: added

bee-network/bee-gossip/src/network.rs
Alex6323: I guess that's fixed.

bee-node/bee-node/Cargo.toml
Alex6323: added

bee-network/bee-gossip/src/conn.rs
Alex6323: bc only if it's exported via `pub mod conn` or `pub use my_func` the warning goes away (here it's `pub` but still not part of the public API of the crate).

bee-network/bee-gossip/src/backstage/network.rs
Alex6323: thx

bee-network/bee-autopeering/build.rs
Alex6323: updated... actually

bee-network/bee-autopeering/Cargo.toml
Alex6323: done

bee-network/bee-autopeering/Cargo.toml
Alex6323: done

bee-network/bee-autopeering/src/config.rs
Alex6323: Fixed, thx

bee-network/bee-gossip/Cargo.toml
Alex6323: done

bee-network/bee-gossip/src/config.rs
Alex6323: changed

bee-network/bee-gossip/src/handshake.rs
```diff
@@ -0,0 +1,380 @@
...
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+type Alias = String;
+
+pub async fn handshake(
+    stream: TcpStream,
+    socket_addr: SocketAddr,
+    local_id: &LocalId,
+    direction: Direction,
```
pvdrz, line 36: I wonder if this argument could be a `const OUTBOUND: bool` generic parameter. What do you think @Alex6323?

bee-network/bee-gossip/src/handshake.rs
```diff
@@ -0,0 +1,380 @@
...
+
+                break packet;
+            }
+        }
+    };
+
+    let peer_public_key = packet.public_key();
+    let mut pk = [0u8; 32];
+    pk.copy_from_slice(&peer_public_key[..32]);
+    let peer_public_key = ed25519::PublicKey::try_from_bytes(pk).map_err(HandshakeError::PublicKey)?;
```
pvdrz, line 142: is this copy necessary?

bee-network/bee-gossip/src/handshake.rs
```diff
@@ -0,0 +1,380 @@
...
+
+impl HandshakeRequest {
+    /// Creates a new handshake request.
+    pub fn new(to: IpAddr) -> Self {
+        let inner = proto::HandshakeRequest {
+            version: VERSION,
+            to: to.to_string(),
+            timestamp: SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("system clock error")
```
pvdrz, line 246: I wonder if we should normalize ` timestamp: SystemTime::now().duration_since(UNIX_EPOCH).expect("here be dragons")` in a crate like `bee-common` or `bee-utils`.

bee-network/bee-gossip/src/peer.rs
```diff
@@ -0,0 +1,291 @@
...
+
+                while position < num_received {
+                    // Determine the length of the next message within this batch
+                    let mut msg_len_buf = [0u8; 4];
+                    msg_len_buf.copy_from_slice(&self.buffer[position..position + 4]);
+                    let msg_len = Buf::get_u32(&mut &msg_len_buf[..]) as usize - 1;
+
+                    // Determine the message type
+                    let msg_type: MessageType =
+                        num::FromPrimitive::from_u8(self.buffer[position + 4]).ok_or(Error::UnknownMessageType)?;
```
pvdrz, line 102: can we `self.buffer.split` instead to avoid extra bound checks?

bee-network/bee-gossip/src/handshake.rs
pvdrz: I realized it is necessary because all those methods receive `[u8; 32]` arrays and not references to them. 

However, I thought that maybe we are double checking bounds when calling `copy_from_slice` as the signature of that function uses a slice and not an array referece, meaning that it doesn't have the size of the slice during compilation.

So I decided to write an unsafe version of this function that calls `clone` instead and check the generated assembly in my good ol' friend [godbolt](https://rust.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:2,endLineNumber:9,positionColumn:2,positionLineNumber:9,selectionStartColumn:2,selectionStartLineNumber:9,startColumn:2,startLineNumber:9),source:'pub+fn+foo(bytes:+%26%5Bu8%5D)+-%3E+%5Bu8%3B+32%5D+%7B%0A++++let+mut+arr+%3D+%5B0%3B+32%5D%3B%0A++++arr.copy_from_slice(%26bytes%5B..32%5D)%3B%0A++++arr%0A%7D%0A%0Apub+fn+foo2(bytes:+%26%5Bu8%5D)+-%3E+%5Bu8%3B+32%5D+%7B%0A++++unsafe%7B%26*(%26bytes%5B..32%5D+as+*const+%5Bu8%5D+as+*const%5Bu8%3B+32%5D)%7D.clone()%0A%7D'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1520,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-O',selection:(endColumn:12,endLineNumber:40,positionColumn:1,positionLineNumber:23,selectionStartColumn:12,selectionStartLineNumber:40,startColumn:1,startLineNumber:23),source:1,tree:'1'),l:'5',n:'0',o:'rustc+1.52.0+(Rust,+Editor+%231,+Compiler+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,fontUsePx:'0',tree:'1',wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.52.0+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

The diff between the two functions is:
```diff
1c1
< example::foo:
---
> example::foo2:
3,5d2
<         xorps   xmm0, xmm0
<         movups  xmmword ptr [rdi + 16], xmm0
<         movups  xmmword ptr [rdi], xmm0
7c4
<         jbe     .LBB0_1
---
>         jbe     .LBB1_1
15,16c12,13
< .LBB0_1:
<         lea     rax, [rip + .L__unnamed_1]
---
> .LBB1_1:
>         lea     rax, [rip + .L__unnamed_2]
```
which means we are saving one `xor` and two `mov` with the unsafe version.

EDIT: this change is implemented in https://github.com/iotaledger/bee/pull/674/commits/3e2999221111558ffc37df381b985d30a8c3bffb

bee-network/bee-gossip/src/handshake.rs
pvdrz: Change done in https://github.com/iotaledger/bee/pull/674/commits/5d3e917b8ce87a105483f3937dd14833470f6c13

bee-network/bee-gossip/src/peer.rs
pvdrz: Done in https://github.com/iotaledger/bee/pull/674/commits/3b5f562211ba7d8de2c217b5c2a7fd86cf8e01dd 

This  is the change that I'm least sure it actually makes anything faster. But the logic here is that any operation with the shape `array[constant]` is optimized away by the compiler and removes the bound check.

bee-network/bee-gossip/src/handshake.rs
Alex6323: I watched a few of Matt's talks, because he's also super entertaining :smile: I wonder how I haven't seen anything from him yet. I also wonder though whether we're falling into the pre-optimization trap here. :thinking:  IMHO benchmarks should first show, whether introducing `unsafe` here is really worth it. As Matt said in the talk you linked to, removing some ops in the assempbly doesn't necessarily mean, that this results in a noticable performance improvement. Especially in this situation here: Handshaking is only done once per gossip connection, maybe roughly done at a rate 1(handshake):10000(gossip messages), for manual peers this ratio is probably even smaller.

bee-network/bee-gossip/src/handshake.rs
Alex6323: We can do that @pvdrz , but it comes (I think) with two downsides (that we need to decide it's worth having them):
(1) `match handshake::<false>(....)` is a bit less readable:
The reader has to first look up `handshake`s signature to understand what the consequences of `false` are. (Not really a problem for most IDE users which usually will pop up the functions documentation.)
(2) Increased binary size. Monomorphization will duplicate this function as they are both used.

What are your thoughts on it? @pvdrz

bee-network/bee-gossip/src/handshake.rs
Alex6323: sounds good to me :+1:

bee-network/bee-gossip/src/peer.rs
Alex6323: Understood, but why not just use `get_unchecked`?

bee-network/bee-gossip/src/handshake.rs
Alex6323: I think (2) is negligible in this case as the function is not that big, but (1) disturbs me quite a bit. If it would be `handshake::<Direction::Inbound>(...)` things were different.

bee-network/bee-gossip/src/handshake.rs
pvdrz: I agree that this code doesn't run a lot but this same idea of copying into an array is used in several places in the network. Maybe we should abstract this to a new safe function of type `fn<const N: usize, T>(&[T]) -> Option<&[T; N]>` in some utils crate or something.

EDIT: I measured the difference in valgrind and there is no change in the number of executed instructions :trollface: . So most likely these instructions never run or something.

bee-network/bee-gossip/src/peer.rs
pvdrz: are we actually sure that this code won't panic if the message is invalid? otherwise `get_unchecked` would cause UB.

bee-network/bee-gossip/src/handshake.rs
pvdrz: Yeah the cost of (2) doesn't worry me that much because we end up with two shorter functions instead of a longer one (even though the two functions are longer together).

About (1) it is a bit hard because const generics only allows bools and ints for now as parameters. We could have two constants called `OUTBOUND` and `INBOUND` instead. What do you think?

bee-network/bee-gossip/src/peer.rs
Alex6323: Right. I think we have no guarantees. Before indexing into it, we could always make sure, that enough bytes (at least 5) are left in the buffer, and thus, prevent any panics and/or UB.

PR #682, opened 2021-07-30
Make `Wrap::wrap` a fallible operation

pvdrz:
This makes `Wrap::wrap` a fallible operation which can help avoid extra `clone()` calls while packing.


bee-common/bee-packable/src/wrap.rs
```diff
@@ -2,20 +2,25 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! A module to wrap [`Packable`](crate::Packable) values.
+use core::convert::Infallible;
```
thibault-martinez, line 5:
```suggestion

use core::convert::Infallible;
```

PR #690, opened 2021-08-05
Derive `Packable` for `IndexationPayload`

pvdrz:
what the title says :)

### DIFF
--- a/bee-message/src/payload/indexation/mod.rs
+++ b/bee-message/src/payload/indexation/mod.rs
@@ -9,10 +9,7 @@ use crate::{payload::MessagePayload, MessageUnpackError, ValidationError, MESSAG
 
 pub use padded::PaddedIndex;
 
-use bee_packable::{
-    coerce::PackCoerceInfallible, error::UnpackPrefixError, BoundedU32, InvalidBoundedU32, PackError, Packable, Packer,
-    UnpackError, Unpacker, VecPrefix,
-};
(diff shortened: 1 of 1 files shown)


bee-message/src/payload/indexation/mod.rs
```diff
@@ -27,13 +24,22 @@ const PREFIXED_INDEX_LENGTH_MIN: u32 = *INDEXATION_INDEX_LENGTH_RANGE.start() as
 const PREFIXED_INDEX_LENGTH_MAX: u32 = *INDEXATION_INDEX_LENGTH_RANGE.end() as u32;
 const PREFIXED_DATA_LENGTH_MAX: u32 = *MESSAGE_LENGTH_RANGE.end() as u32;
 
+fn unpack_prefix_to_validation_error(err: UnpackPrefixError<Infallible>) -> ValidationError {
+    match err {
+        UnpackPrefixError::InvalidPrefixLength(len) => ValidationError::InvalidIndexationIndexLength(len),
+        UnpackPrefixError::Packable(e) => match e {},
+    }
+}
```
Adam-Gleave, line 32: Doesn't this return `InvalidIndexationIndexLength` even if it's the data field that's incorrect?

pvdrz: Oh yes... that's not good at all... But it also means there's a hole in the tests :trollface:. I think we can transform the `Packable` derive into super saiyan 5 and do something that let's you do `with` on a field basis.

pvdrz:
Closed in favor of #692

PR #698, opened 2021-08-12
Add MQTT plugin

samuel-rufi:
Relevant MQTT specification: https://playground.asyncapi.io/?load=https://raw.githubusercontent.com/luca-moser/protocol-rfcs/rfc/node-event-api/text/0033-node-event-api/0033-node-event-api.yml


bee-node/src/plugins/mqtt/broker.rs
```diff
@@ -0,0 +1,84 @@
...
+use bee_runtime::{node::Node, worker::Worker};
+
+use async_trait::async_trait;
+
+use librumqttd::{LinkRx, LinkTx};
+use log::*;
+
+use std::convert::Infallible;
+
+pub struct MqttBrokerConfig {
```
Alex6323: maybe "Config" is not the best name for this type as it doesn't contain any configuration data, but only channel halfs. Might still be from what I introduced back then. (Feel free to change if so :smile: ).

Alex6323: I see why it's done like this, but it's still confusing. Can we not move what is done when creating this struct somewhere else into `start` down below, and then return a tuple `(Self, ...channel_halfs...)`? That would mean that the `LinkTx, LinkRx` are created within `start`, and also we would have`type Config = ()` until we are able to properly configure the broker (with some of those constants from the other module)

bee-node/Cargo.toml
```diff
@@ -36,9 +36,10 @@ iota-crypto = { version = "0.7.0", features = [ "ed25519", "random", "blake2b" ]
 log = "0.4"
 mime_guess = { version = "2.0", optional = true }
 multiaddr = "0.13"
-paho-mqtt = { version = "0.9", default-features = false, features = [ "bundled" ]}
 rand = "0.8"
 rpassword = "5.0"
+rumqttd = "0.7.0"
+rumqttlog = "0.7.0"
```
thibault-martinez: Remove patch

bee-node/Cargo.toml
```diff
@@ -36,9 +36,10 @@ iota-crypto = { version = "0.7.0", features = [ "ed25519", "random", "blake2b" ]
 log = "0.4"
 mime_guess = { version = "2.0", optional = true }
 multiaddr = "0.13"
-paho-mqtt = { version = "0.9", default-features = false, features = [ "bundled" ]}
 rand = "0.8"
 rpassword = "5.0"
+rumqttd = "0.7.0"
```
thibault-martinez: Remove patch

bee-node/src/plugins/mqtt/broker.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{plugins::mqtt::handlers::*, storage::StorageBackend};
+
+use bee_runtime::{node::Node, worker::Worker};
+
+use async_trait::async_trait;
+
```
thibault-martinez:
```suggestion
```

bee-node/src/plugins/mqtt/error.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright 2020 IOTA Stiftung
```
thibault-martinez: Change all headers to 2021

bee-node/src/plugins/mqtt/handlers/messages_solidified.rs
```diff
@@ -0,0 +1,98 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{plugins::mqtt::broker::MqttBroker, storage::StorageBackend};
+
+use bee_message::{MessageId};
+use bee_rest_api::types::{responses::MessageMetadataResponse};
+use bee_runtime::{node::Node, shutdown_stream::ShutdownStream};
+use bee_tangle::{MsTangle};
+
+use bee_protocol::workers::event::MessageSolidified;
```
thibault-martinez: With other bee imports

bee-node/src/plugins/mqtt/handlers/outputs.rs
```diff
@@ -0,0 +1,106 @@
...
+        consensus::{ConsensusWorker, ConsensusWorkerCommand},
+        error::Error,
+    },
+};
+use bee_message::output::OutputId;
+use bee_rest_api::types::responses::OutputResponse;
+use bee_runtime::node::Node;
+use bee_storage::access::Fetch;
+
+use crate::plugins::mqtt::handlers::HandlerError;
```
thibault-martinez: With other crate import

bee-node/src/plugins/mqtt/handlers/transactions_included_message.rs
```diff
@@ -0,0 +1,87 @@
...
+};
+
+use bee_ledger::types::CreatedOutput;
+use bee_message::output::OutputId;
+use bee_runtime::node::Node;
+use bee_storage::access::Fetch;
+
+use bee_common::packable::Packable;
+use bee_message::payload::transaction::TransactionId;
+use bee_tangle::MsTangle;
```
thibault-martinez: With other bee imports

bee-protocol/src/workers/message/payload/indexation.rs
```diff
@@ -5,12 +5,14 @@ use crate::{
     types::metrics::NodeMetrics,
     workers::{storage::StorageBackend, MetricsWorker},
 };
+use crate::workers::event::IndexationMessage;
```
thibault-martinez: Merge them

bee-protocol/src/workers/message/payload/indexation.rs
```diff
@@ -5,12 +5,14 @@ use crate::{
...
 
 use bee_message::{
     payload::{indexation::PaddedIndex, transaction::Essence, Payload},
     MessageId,
 };
 use bee_runtime::{node::Node, shutdown_stream::ShutdownStream, worker::Worker};
+use bee_runtime::event::Bus;
```
thibault-martinez: Merge them

thibault-martinez (changes requested):

bee-node/src/plugins/mqtt/broker.rs
samuel-rufi: Agree with you that "Config" is not the best name for this type as it doesn't contain any configuration data, but only channel halfs. So I would rename this too.

Regarding your second comment:
If I understood correctly, you would like to move [following code](https://github.com/rufsam/bee/blob/chrysalis-pt-2/bee-node/src/plugins/mqtt/mod.rs#L95-L188) to the `start()` function right?
Also, instead of passing [here ](https://github.com/rufsam/bee/blob/chrysalis-pt-2/bee-node/src/plugins/mqtt/mod.rs#L208) the channel halfs, we could pass [that config](https://github.com/rufsam/bee/blob/chrysalis-pt-2/bee-node/src/plugins/mqtt/mod.rs#L84).

Alex6323: something like that, yea. I think the code is better understandable and consistent with our other crates if a `Config` actually contains data, that changes/tweaks the behavior of the code to the user's likings. Contrary to that those channel halfs are essential to the correct working of the system, hence is not configurable in any way. You can basically do whatever you think is appropriate to achieve this kind of architecture :smiley:

samuel-rufi: absolutely! i have updated the PR, it should be cleaner now :D

bee-node/src/plugins/mqtt/broker.rs
```diff
@@ -0,0 +1,133 @@
...
+where
+    N::Backend: StorageBackend,
+{
+    type Config = mqtt::Config;
+    type Error = Infallible;
+
+    async fn start(node: &mut N, config: Self::Config) -> Result<Self, Self::Error> {
+        let mut broker = mqtt::Broker::new(config);
+
+        let mut milestones_latest_tx = broker.link("milestones-latest").expect("linking mqtt sender failed");
```
Adam-Gleave, line 31: There's lots of `expect`s here -- is this something that makes sense to describe with an error result or is the panicking fine?

bee-node/src/plugins/mqtt/handlers/messages_referenced.rs
```diff
@@ -0,0 +1,101 @@
...
+    let tangle = node.resource::<Tangle<N::Backend>>();
+    let (tx, rx) = mpsc::unbounded_channel::<MessageReferenced>();
+
+    node.spawn::<MqttBroker, _, _>(|shutdown| async move {
+        debug!("MQTT 'messages/referenced' topic handler running.");
+
+        let mut receiver = ShutdownStream::new(shutdown, UnboundedReceiverStream::new(rx));
+
+        while let Some(event) = receiver.next().await {
+            // the message is newly referenced and therefore it's safe to unwrap
```
Adam-Gleave, line 32: Really picky but starting comments with a capital and ending with a full-stop makes everything consistent 😅

bee-node/src/plugins/mqtt/mod.rs
```diff
@@ -1,101 +1,93 @@
...
+    } = config;
+
+    let connection_settings = mqtt::ConnectionSettings {
+        connection_timeout_ms: DEFAULT_CONNECTION_TIMEOUT_MS,
+        max_client_id_len: DEFAULT_MAX_CLIENT_ID_LEN,
+        throttle_delay_ms: DEFAULT_THROTTLE_DELAY_MS,
+        max_payload_size: DEFAULT_MAX_PAYLOAD_SIZE,
+        max_inflight_count: DEFAULT_MAX_INFLIGHT_COUNT,
+        max_inflight_size: DEFAULT_MAX_INFLIGHT_SIZE,
+        username: None, // Option<String>,
```
Adam-Gleave, line 57: Do we need these comments here?

Adam-Gleave:
Are we still requiring error/panic messages to be all lowercase with no period? I think that was what we decided on a little while back.

bee-node/src/plugins/mqtt/mod.rs
samuel-rufi: these constants are basically copied from the mqtt library examples... however there are no comments too. so not fully sure how they are used. i can try to comment them, but not fully sure on them

samuel-rufi: what do you suggest? also @thibault-martinez

bee-node/src/plugins/mqtt/handlers/messages_referenced.rs
samuel-rufi: haha im picky on comments too, will change

bee-node/src/plugins/mqtt/broker.rs
samuel-rufi: hmm right i guess we should return the error here instead

bee-node/src/plugins/mqtt/mod.rs
Adam-Gleave: Oh, I meant the inline comments with the types. I'm fine with the constants (although if we want to be really picky they're declared in what looks like a random order :trollface:)

PR #701, opened 2021-08-13
Add protobuf version of `Message`

pvdrz:
It seems we need to be able to expose something akin to `bee-message::Message` from `bee-plugin` so we can expose messages to plugins.

Next steps:
- [x] Finish the `From<bee_message::Message>` implementation.
- [x] Decide how to handle the inverse of the previous conversion.
- [ ] Figure out if there is a way to avoid the extra `Option` types used because of the way `prost` serializes `oneof` fields.
- [x] Document all the new public types.
- [x] Test that conversions between message representations are correct.

cc: @thibault-martinez @Adam-Gleave

### DIFF
--- a/bee-node/bee-plugin/proto/plugin.proto
+++ b/bee-node/bee-plugin/proto/plugin.proto
@@ -62,7 +62,7 @@ message MessageId {
 }
 
 message Payload {
-    oneof payload_kind {
+    oneof kind {
         DataPayload data = 1;
         TransactionPayload transaction = 2;
         FpcPayload fpc = 3;
@@ -85,7 +85,7 @@ message TransactionPayload {
 }
 
 message UnlockBlock {
-    oneof unlock_block_kind {
+    oneof kind {
         SignatureUnlock signature = 1;
         ReferenceUnlock reference = 2;
     }
@@ -96,7 +96,7 @@ message SignatureUnlock {
 }
 
 message Signature {
-    oneof signature_kind {
+    oneof kind {
         Ed25519Signature ed25519 = 1;
         BlsSignature bls = 2;
     }
@@ -125,7 +125,7 @@ message TransactionEssence {
 }
 
 message Input {
-    oneof input_kind {
+    oneof kind {
         UtxoInput utxo = 1;
     }
 }
@@ -144,7 +144,7 @@ message TransactionId {
 }
 
 message Output {
-    oneof output_kind {
+    oneof kind {
         SignatureLockedSingleOutput signature_locked_single = 1;
         SignatureLockedAssetOutput signature_locked_asset = 2;
     }
@@ -156,7 +156,7 @@ message SignatureLockedSingleOutput {
 }
 
 message Address {
-    oneof address_kind {
+    oneof kind {
         Ed25519Address ed25519 = 1;
         BlsAddress bls = 2;
     }
--- a/bee-node/bee-plugin/proto/plugin.proto
+++ b/bee-node/bee-plugin/proto/plugin.proto
@@ -17,8 +17,225 @@ message ShutdownReply {}
 message ProcessReply {}
 
 // Event triggered when a message succesfully passes the parsing stage.
-message MessageParsedEvent {}
+message MessageParsedEvent {
+    // The parsed message.
+    Message message = 1;
+}
+
 // Event triggered when parsing a message fails.
 message ParsingFailedEvent {}
 // Event triggered when the syntactical validation of a message fails.
 message MessageRejectedEvent {}
+
+// A message from the IOTA network.
+message Message {
+    repeated ParentsBlock parents_block = 1;
+    bytes issuer_public_key = 2;
+    uint64 issue_timestamp = 3;
+    uint32 sequence_number = 4;
+    optional Payload payload = 5;
+    uint64 nonce = 6;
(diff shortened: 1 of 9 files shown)


bee-node/bee-plugin/proto/plugin.proto
```diff
@@ -17,8 +17,237 @@ message ShutdownReply {}
...
+    // Blocks of message parents. Each block contains a list of parent message IDs grouped by type.
+    repeated ParentsBlock parents_blocks = 1;
+    // The public key of the issuing node.
+    bytes issuer_public_key = 2;
+    // The Unix timestamp at the moment of issue.
+    uint64 issue_timestamp = 3;
+    // The sequence number of the message, indicating the marker sequence it belongs to.
+    uint32 sequence_number = 4;
+    // The optional payload of the message.
+    Payload payload = 5;
```
thibault-martinez, line 41: Is there no way to do optionals with pb3 ?

bee-node/bee-plugin/src/message.rs
```diff
@@ -0,0 +1,783 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Types for messages.
+
+pub use crate::grpc::Message;
+
+use std::convert::TryInto;
```
thibault-martinez: After crate::grpc

bee-node/bee-plugin/src/message.rs
```diff
@@ -0,0 +1,783 @@
...
+
+use crate::grpc::{
+    address::Kind as AddressKind, input::Kind as InputKind, output::Kind as OutputKind, payload::Kind as PayloadKind,
+    signature::Kind as SignatureKind, unlock_block::Kind as UnlockBlockKind, Address, ApplicationMessagePayload,
+    AssetBalance, AssetId, BeaconPayload, BlsAddress, BlsSignature, CollectiveBeaconPayload, Conflict, DataPayload,
+    DkgPayload, Ed25519Address, Ed25519Signature, EncryptedDeal, FpcPayload, IndexationPayload, Input, MessageId,
+    Output, OutputId, ParentsBlock, ParentsKind, Payload, ReferenceUnlock, Salt, SaltDeclarationPayload, Signature,
+    SignatureLockedAssetOutput, SignatureLockedSingleOutput, SignatureUnlock, Timestamp, TransactionEssence,
+    TransactionId, TransactionPayload, UnlockBlock, UtxoInput,
+};
```
Adam-Gleave, line 27: Can we make an exception and use `::*` here? This makes me all kinds of sad...

bee-node/bee-plugin/src/message.rs
```diff
@@ -0,0 +1,783 @@
...
+}
+
+#[allow(clippy::from_over_into)]
+impl Into<bee_message::payload::transaction::TransactionPayload> for TransactionPayload {
+    fn into(self) -> bee_message::payload::transaction::TransactionPayload {
+        bee_message::payload::transaction::TransactionPayloadBuilder::default()
+            .with_essence((*self.essence.unwrap()).into())
+            .with_unlock_blocks(
+                bee_message::unlock::UnlockBlocks::new(self.unlock_blocks.into_iter().map(UnlockBlock::into).collect())
+                    .unwrap(),
```
Adam-Gleave, line 200: Maybe some explanation somewhere as to why all these unwraps are okay (just once somewhere in the file?)

bee-node/bee-plugin/src/message.rs
Adam-Gleave: Oh no I just realised the `as`... 🪦

bee-node/bee-plugin/proto/plugin.proto
pvdrz: sadly/luckily, every non-scalar field is represented as an `Option<T>` anyway so we're good here.

bee-node/bee-plugin/src/message.rs
pvdrz: yeah you're right :)

bee-node/bee-plugin/src/message.rs
```diff
@@ -0,0 +1,793 @@
...
+            issue_timestamp: message.issue_timestamp(),
+            sequence_number: message.sequence_number(),
+            payload: message.payload().as_ref().map(Payload::from),
+            nonce: message.nonce(),
+            signature: message.signature().to_vec(),
+        }
+    }
+}
+
+#[allow(clippy::from_over_into)]
```
thibault-martinez: Maybe do a module level one instead ?

bee-node/bee-plugin/src/message.rs
```diff
@@ -0,0 +1,793 @@
...
+impl Into<bee_message::parents::ParentsBlock> for ParentsBlock {
+    fn into(self) -> bee_message::parents::ParentsBlock {
+        // This and every other `unwrap` in an `Into::into` implementation is ok because we assume
+        // that `self` was created from a valid `bee_message::Message` and it has no missing
+        // fields.
```
thibault-martinez: Is this necessary since you already have a note about it at module level ? This isn't even the first unwrap of the file so this is weird

thibault-martinez (approved):

PR #722, opened 2021-09-12
Add possibility to add password over env var

Dr-Electron:
# Description of change

Add possibility to pass the password for the password tool via environment variable. 

## Type of change

Tested locally with and without env var.

## Change checklist

### DIFF
--- a/bee-node/src/tools/password.rs
+++ b/bee-node/src/tools/password.rs
@@ -29,8 +29,8 @@ pub fn exec(_tool: &PasswordTool) -> Result<(), PasswordError> {
         env_password
     } else {
         let password = read_password_from_tty(Some("Password: "))?;
-        let password_reenter = read_password_from_tty(Some("Re-enter password: "))?;
-        if password != password_reenter {
+
+        if password != read_password_from_tty(Some("Re-enter password: "))? {
             return Err(PasswordError::NonMatching);
         }
         password
--- a/bee-node/src/tools/password.rs
+++ b/bee-node/src/tools/password.rs
@@ -25,16 +25,16 @@ pub enum PasswordError {
 pub struct PasswordTool {}
 
 pub fn exec(_tool: &PasswordTool) -> Result<(), PasswordError> {
-    let password;
-    if let Ok(env_password) = env::var("BEE_TOOL_PASSWORD") {
-        password = env_password;
+    let password = if let Ok(env_password) = env::var("BEE_TOOL_PASSWORD") {
+        env_password
     } else {
-        password = read_password_from_tty(Some("Password: "))?;
+        let password = read_password_from_tty(Some("Password: "))?;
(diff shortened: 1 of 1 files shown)


bee-node/src/tools/password.rs
```diff
@@ -23,11 +25,17 @@ pub enum PasswordError {
 pub struct PasswordTool {}
 
 pub fn exec(_tool: &PasswordTool) -> Result<(), PasswordError> {
-    let password = read_password_from_tty(Some("Password: "))?;
-    let password_reenter = read_password_from_tty(Some("Re-enter password: "))?;
-    if password != password_reenter {
-        return Err(PasswordError::NonMatching);
+    let password;
+    if let Ok(env_password) = env::var("BEE_TOOL_PASSWORD") {
+        password = env_password;
```
thibault-martinez: Just make the if/else return the password instead

Alex6323: I also wanted to comment on that. I'ld suggest the following (Thibault probably means the same):
```rust
let password = if let Ok(env_password) = .... {
    env_password
} else {
    ...
    password
};
```

bee-node/src/tools/password.rs
```diff
@@ -23,11 +25,17 @@ pub enum PasswordError {
...
-    let password_reenter = read_password_from_tty(Some("Re-enter password: "))?;
-    if password != password_reenter {
-        return Err(PasswordError::NonMatching);
-    }
+    let password = if let Ok(env_password) = env::var("BEE_TOOL_PASSWORD") {
+        env_password
+    } else {
+        let password = read_password_from_tty(Some("Password: "))?;
+        let password_reenter = read_password_from_tty(Some("Re-enter password: "))?;
+        if password != password_reenter {
```
thibault-martinez: Can you compare to `read_password_from_tty` directly ? Since we don't actually use the result apart from the comparison

thibault-martinez, line 33:
```suggestion

                if password != password_reenter {
```

thibault-martinez:
```suggestion

        if password != password_reenter {
```

Dr-Electron: Of course. I just thought it maybe would be more user readable this way :)

bee-node/src/tools/password.rs
```diff
@@ -23,11 +25,17 @@ pub enum PasswordError {
...
-    }
+    let password = if let Ok(env_password) = env::var("BEE_TOOL_PASSWORD") {
+        env_password
+    } else {
+        let password = read_password_from_tty(Some("Password: "))?;
+
+        if password != read_password_from_tty(Some("Re-enter password: "))? {
+            return Err(PasswordError::NonMatching);
+        }
+        password
```
thibault-martinez, line 37:
```suggestion

        password
```

thibault-martinez (approved):
Thanks :)

PR #756, opened 2021-09-30
Improve console feature + remove spawner traits

Adam-Gleave:
# Description of change

Improves the way that the `console` feature works, since `tokio` have made a number of recent updates that make it easier to use.

Removes the `TaskSpawner` trait, since we no longer need any instrumentation to display task locations. Unfortunately these still do not show up in `console`s "Location" column, since that relies on a change in `tokio` that has not yet been released. Instead, we just use a `spawn` function with the `track_caller` attribute, so that `tokio` creates spans containing the parent location, that shows up in the "Fields" column on the console.

Also improves the logging when using the `console` feature. The subscriber implementation has been improved, so we can now use the config file in the same way as we do with our standard logging, with the exception that it (currently) only supports logging to `stdout`, and not to files (I might take a look into this though). Because the logger initialisation is now more complex, I have moved it to the `bee-common` crate.

`bee-runtime` could probably use a change in version, since these changes are breaking. 

## Type of change

Tested different logging configurations for `stdout`, and everything works as expected. Task spawn location shows up in the `console` view.


bee-common/bee-common/Cargo.toml
```diff
@@ -12,16 +12,19 @@ homepage = "https://www.iota.org"
...
 chrono = "0.4"
+console-subscriber = { git = "https://github.com/tokio-rs/console.git", branch = "main", optional = true }
 fern = { version = "0.6", features = [ "colored" ] }
 jsonwebtoken = { version = "7.2", optional = true }
 log = { version = "0.4", features = [ "serde" ] }
 rand = { version = "0.8", optional = true }
 rust-argon2 = { version = "0.8", optional = true }
 serde = { version = "1.0", features = [ "derive" ] }
 thiserror = "1.0"
+tracing-subscriber = { version = "0.2.21", optional = true }
```
Alex6323: Should be safe to go without the patch version here as well, right?

thibault-martinez: Yes please, remove it

thibault-martinez (changes requested):
CI failing

bee-runtime/Cargo.toml
```diff
@@ -24,4 +24,4 @@ tracing = { version = "0.1", optional = true }
 tokio = { version = "1.11", features = [ "rt", "macros", "time" ] }
 
 [features]
-console = [ "tracing" ]
+tokio-console = [ "tracing" ]
```
thibault-martinez: Not needed anymore ? Same for the dep ?

thibault-martinez (changes requested):

bee-runtime/Cargo.toml
Adam-Gleave: Yep, will remove it 👍

bee-common/bee-common/src/logger/config.rs
```diff
@@ -14,10 +14,10 @@ const DEFAULT_COLOR_ENABLED: bool = true;
 const DEFAULT_TARGET_WIDTH: usize = 42;
 /// Default value for the level width.
 const DEFAULT_LEVEL_WIDTH: usize = 5;
-/// Default name for an output.
-const DEFAULT_OUTPUT_NAME: &str = LOGGER_STDOUT_NAME;
 /// Default log level for an output.
 const DEFAULT_OUTPUT_LEVEL: LevelFilter = LevelFilter::Info;
+/// Default name for an output.
+const DEFAULT_OUTPUT_NAME: &str = LOGGER_STDOUT_NAME;
```
thibault-martinez: It was because name is before level in the config :(

Adam-Gleave: Ah, I moved it up a level earlier and then moved it back to the wrong place :trollface:

bee-common/bee-common/CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security -->
 
+## 0.5.0 - 2021-09-30
+
+### Added
+
+- `tokio-console` feature, enabling improved logging when using tokio's `console` project.
```
thibault-martinez:
```suggestion
- `tokio-console` feature, enabling improved logging when using tokio's `console` project;
```

thibault-martinez (dismissed):

pvdrz (dismissed):
That was a short run for `TaskSpawner` :trollface: 
image

thibault-martinez (dismissed):

thibault-martinez (approved):

pvdrz (approved):

Alex6323 (approved):

PR #760, opened 2021-10-07
Add MessageMetadata storage

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-storage/bee-storage-rocksdb/src/access/batch.rs
+++ b/bee-storage/bee-storage-rocksdb/src/access/batch.rs
@@ -5,7 +5,7 @@
 
 use crate::{column_families::*, Storage};
 
-use bee_message::{Message, MessageId};
+use bee_message::{Message, MessageId, MessageMetadata};
 use bee_packable::Packable;
 use bee_storage::{
     access::{Batch, BatchBuilder},
@@ -62,6 +62,39 @@ impl Batch<MessageId, Message> for Storage {
         batch
             .inner
             .delete_cf(self.cf_handle(CF_MESSAGE_ID_TO_MESSAGE)?, message_id);
+
+        Ok(())
+    }
+}
+
+impl Batch<MessageId, MessageMetadata> for Storage {
+    fn batch_insert(
+        &self,
+        batch: &mut Self::Batch,
+        message_id: &MessageId,
+        message_metadata: &MessageMetadata,
+    ) -> Result<(), <Self as StorageBackend>::Error> {
+        batch.value_buf.clear();
+        // Packing to bytes can't fail.
+        message_metadata.pack(&mut batch.value_buf).unwrap();
+
+        batch.inner.put_cf(
+            self.cf_handle(CF_MESSAGE_ID_TO_MESSAGE_METADATA)?,
+            message_id,
+            &batch.value_buf,
+        );
+
+        Ok(())
+    }
+
+    fn batch_delete(
+        &self,
+        batch: &mut Self::Batch,
+        message_id: &MessageId,
+    ) -> Result<(), <Self as StorageBackend>::Error> {
+        batch
+            .inner
+            .delete_cf(self.cf_handle(CF_MESSAGE_ID_TO_MESSAGE_METADATA)?, message_id);
 
         Ok(())
     }
--- a/bee-storage/bee-storage-sled/src/access/batch.rs
+++ b/bee-storage/bee-storage-sled/src/access/batch.rs
@@ -5,7 +5,7 @@
 
 use crate::{trees::*, Storage};
 
-use bee_message::{Message, MessageId};
+use bee_message::{Message, MessageId, MessageMetadata};
 use bee_packable::packable::Packable;
 use bee_storage::{
     access::{Batch, BatchBuilder},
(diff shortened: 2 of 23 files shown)


bee-storage/bee-storage-sled/src/access/batch.rs
```diff
@@ -69,3 +69,38 @@ impl Batch<MessageId, Message> for Storage {
         Ok(())
     }
 }
+
+impl Batch<MessageId, MessageMetadata> for Storage {
+    fn batch_insert(
+        &self,
+        batch: &mut Self::Batch,
+        message_id: &MessageId,
+        message_metada: &MessageMetadata,
```
Alex6323: typo (and a funny one :smile: )

thibault-martinez: Ugh

bee-storage/bee-storage-sled/src/access/fetch.rs
```diff
@@ -29,3 +29,14 @@ impl Fetch<MessageId, Message> for Storage {
...
 }
+
+impl Fetch<MessageId, MessageMetadata> for Storage {
+    fn fetch(&self, message_id: &MessageId) -> Result<Option<MessageMetadata>, <Self as StorageBackend>::Error> {
+        Ok(self
+            .inner
+            .open_tree(TREE_MESSAGE_ID_TO_MESSAGE_METADATA)?
+            .get(message_id)?
+            // Unpacking from storage slice can't fail.
+            .map(|v| MessageMetadata::unpack(&mut v.as_ref()).unwrap()))
```
Alex6323, line 40: it can't fail as long as there's no hardware failure, right? I understand that we can use `unwrap` if we for example checked before that `option.is_some() == true`, but aren't we relying here on the storage being in a healthy state? I wonder if we should return an error here - something like a `InvalidStorageData`.

thibault-martinez: At this point, `.get(message_id)?` already succeeded so there is no hardware failure. What this unwrap means is that the bytes we actually fetched are not the ones we inserted, handling this would basically mean handling cosmic rays, I am not entirely sure it's necessary.

Adam-Gleave: This could happen if the storage got corrupted somehow, right? But then again, I guess most of these unpacks could fail in that way. I'm not sure we need to handle it, it's obvious that if this ever fails there is a problem with the storage data that shouldn't be the case.

Adam-Gleave (approved):

bee-storage/bee-storage-sled/src/access/fetch.rs
thibault-martinez: If such a thing was to happen, we would need to propagate the error all the way down and stop the program anyway, a panic seems easier.

Adam-Gleave: Agreed, no point in recovering and carrying on running in that state anyway :+1:

Alex6323: The node is doing a lot of other stuff, which should be shut down in a safe manner. Instead of just crashing and burning it could end the socket connections with its peers in a friendly way, and do all kinds of other shutdown logic, before telling the user that the storage got corrupted (due to cosmic rays as usual or a broken disk). Ending the socket connections nicely gives our peers the information, that they need to look for other ones (because on our side went something wrong). If we just crash, there is still overhead on our peers trying to rescue the connection.

thibault-martinez: I am not 100% sur we should invest time and logic in something that may happen once in a lifetime but I'm not opposed to it. Let's see what all the others think.

bee-storage/bee-storage-rocksdb/src/access/batch.rs
```diff
@@ -66,3 +66,36 @@ impl Batch<MessageId, Message> for Storage {
...
+impl Batch<MessageId, MessageMetadata> for Storage {
+    fn batch_insert(
+        &self,
+        batch: &mut Self::Batch,
+        message_id: &MessageId,
+        message_metadata: &MessageMetadata,
+    ) -> Result<(), <Self as StorageBackend>::Error> {
+        batch.value_buf.clear();
+        // Packing to bytes can't fail.
+        message_metadata.pack(&mut batch.value_buf).unwrap();
```
pvdrz, line 79: I'm thinking that we should measure how many of these unwraps we do and maybe do something like this:
```rust
match res {
    Ok(val) => val,
    // SAFETY: Packing to bytes can't fail.
    Err(_) => unsafe { std::hint::unreachable_unchecked() },
}
```

pvdrz (approved):

bee-storage/bee-storage-rocksdb/src/access/batch.rs
thibault-martinez: Can you explain a bit more ?

bee-storage/bee-storage-sled/src/access/fetch.rs
Alex6323: I think for now I am fine it, because we have things more important, I guess. Maybe we should keep this as an issue though (which will then be there for years probably). I dont think hardware failure will happen so rarely, though, because many people run nodes on the cheapest possible VPS for understandably good reasons. I know I did back in the day.😊

PR #762, opened 2021-10-07
Feat/add local snapshot source

no8ody:
This PR adds an additional download source for a local snapshot.
Also the delta path was added again in the comnet config.


bee-node/config.chrysalis-comnet.toml
```diff
@@ -76,6 +76,7 @@ white_flag_solidification_timeout = 2
 
 [snapshot]
 full_path         = "./snapshots/comnet/latest-full_snapshot.bin"
+delta_path        = "./snapshots/mainnet/latest-delta_snapshot.bin"
```
thibault-martinez: comnet

thibault-martinez (approved):

PR #771, opened 2021-10-12
Remove tangle LRU cache

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-tangle/src/tangle.rs
+++ b/bee-tangle/src/tangle.rs
@@ -618,7 +618,7 @@ impl<B: StorageBackend> Tangle<B> {
 
                 if let Some(key) = vertices.iter().next().map(|(k, _)| k).copied() {
                     if vertices.remove(&key).is_some() {
-                        len = self.len.load(Ordering::Relaxed);
+                        len = self.len.fetch_sub(1, Ordering::SeqCst) - 1;
                     }
                 }
             }
--- a/bee-tangle/src/tangle.rs
+++ b/bee-tangle/src/tangle.rs
@@ -17,13 +17,14 @@ use bee_message::{
 };
 use bee_runtime::resource::ResourceHandle;
 
-use hashbrown::{hash_map::DefaultHashBuilder, HashMap};
+use hashbrown::HashMap;
 use log::{info, trace};
-use lru::LruCache;
+use rand::Rng;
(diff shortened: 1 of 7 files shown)


bee-tangle/src/vertices.rs
```diff
@@ -0,0 +1,122 @@
...
+        RwLockWriteGuard::try_map(table, |table| {
+            table.get_mut(hash, equivalent_id(message_id)).map(|(_, v)| v)
+        })
+        .ok()
+    }
+
+    pub(crate) async fn pop_random(&self) -> Option<Vertex> {
+        let index = rand::thread_rng().gen_range(0..self.tables.len());
+        // SAFETY: `index < self.tables.len()` by construction.
+        let mut table = unsafe { self.tables.get_unchecked(index) }.write().await;
```
Adam-Gleave, line 82: Would this work here to get rid of the `unsafe`?

```suggestion
        let mut table = self.tables.choose(&mut rand::thread_rng()).write().await;
```

bee-tangle/src/tangle.rs
```diff
@@ -644,62 +607,38 @@ impl<B: StorageBackend> Tangle<B> {
...
+                let mut retries = 1;
+                loop {
+                    if self.vertices.pop_random().await.is_none() {
+                        log::info!("retrying cache eviction (attempt #{})", retries);
+                        retries += 1;
+                    } else {
+                        break;
+                    }
+
+                    if retries > 10 {
```
Adam-Gleave: Maybe make this a constant somewhere.

bee-tangle/src/vertices.rs
thibault-martinez: I believe the unchecked get was here for performance reason as we know the index is valid. @pvdrz ?

Adam-Gleave: Ah that makes sense, I think performance would definitely be better that way.

pvdrz: yes, but I wouldn't be surprised if rust can optimize it away, who knows

pvdrz (dismissed):

Adam-Gleave (dismissed):

pvdrz (approved):

Adam-Gleave (approved):

PR #785, opened 2021-10-20
Ignore `time` related vulnerabilities

pvdrz:
Angry sec noises


.cargo/audit.toml
```diff
@@ -0,0 +1,2 @@
+[advisories]
```
thibault-martinez: Any way to have this file in `.github/workflows/` ?

pvdrz: i suppose that we could modify the workflows to move whatever is in `.github/workflows/.cargo` to `~/.cargo` but I'd have to investigate.

thibault-martinez: For reference https://github.com/iotaledger/bee/blob/dev/.github/workflows/licenses.yml#L29

Adam-Gleave:
Adding to #786.

thibault-martinez (approved):

PR #787, opened 2021-10-20
Make eviction less "contentious"

pvdrz:
This PR changes the `pop_random` method so it uses `try_write` instead of `write`. Meaning that it should not `await` if the randomly chosen partition is being used.

This should cause less contention while evicting at the cost of increasing the number of retries.

### DIFF
--- a/bee-tangle/src/tangle.rs
+++ b/bee-tangle/src/tangle.rs
@@ -616,6 +616,8 @@ impl<B: StorageBackend> Tangle<B> {
                         "could not perform cache eviction after {} attempts",
                         MAX_EVICTION_RETRIES
                     );
+
+                    break;
                 }
             }
         }
--- a/bee-tangle/src/tangle.rs
+++ b/bee-tangle/src/tangle.rs
@@ -611,19 +611,11 @@ impl<B: StorageBackend> Tangle<B> {
 
         if self.vertices.len() > max_len {
             while self.vertices.len() > ((1.0 - CACHE_THRESHOLD_FACTOR) * max_len as f64) as usize {
-                let mut retries = 1;
-                loop {
-                    if self.vertices.pop_random().await.is_none() {
-                        log::debug!("retrying cache eviction (attempt #{})", retries);
-                        retries += 1;
-                    } else {
-                        break;
-                    }
-
-                    if retries > MAX_EVICTION_RETRIES {
-                        log::warn!("could not perform cache eviction after {} attempts", retries);
-                        return;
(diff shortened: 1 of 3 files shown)


bee-tangle/src/tangle.rs
```diff
@@ -611,19 +611,11 @@ impl<B: StorageBackend> Tangle<B> {
...
-
-                    if retries > MAX_EVICTION_RETRIES {
-                        log::warn!("could not perform cache eviction after {} attempts", retries);
-                        return;
-                    }
+                if self.vertices.pop_random::<MAX_EVICTION_RETRIES>().await.is_none() {
+                    log::warn!(
+                        "could not perform cache eviction after {} attempts",
+                        MAX_EVICTION_RETRIES
+                    );
```
thibault-martinez, line 618: There used to be a return but now it's gone ?

pvdrz: yeah, the retry logic is inside `pop_random` now. If it returns `None` is because the maximum number of retries was exceeded.

thibault-martinez: Sure but then we do nothing and keep looping, where before we stopped even trying

pvdrz: Oh I get it yeah.

bee-tangle/src/vertices.rs
```diff
@@ -76,24 +76,32 @@ impl Vertices {
         .ok()
     }
 
-    pub(crate) async fn pop_random(&self) -> Option<Vertex> {
-        let index = rand::thread_rng().gen_range(0..self.tables.len());
-        // SAFETY: `index < self.tables.len()` by construction.
-        let mut table = unsafe { self.tables.get_unchecked(index) }.write().await;
+    pub(crate) async fn pop_random<const RETRIES: usize>(&self) -> Option<Vertex> {
```
thibault-martinez: What's the advantage of having RETRIES a generic param ?

pvdrz: to be honest it could be a regular function parameter. I just wanted to be able to control the number of retries from the `Tangle` instead.

thibault-martinez: Was thinking that maybe this is the kind of things we may want to have configurable to be able to tweak without recompiling

pvdrz: Ok I'll add it to the configuration then

thibault-martinez (approved):

PR #795, opened 2021-10-25
Add `into_packer` and `into_unpacker` methods for packable errors.

pvdrz:


thibault-martinez (dismissed):

bee-common/bee-packable/src/packable/bounded.rs
```diff
@@ -63,6 +63,12 @@ macro_rules! bounded {
             const MAX: $ty = MAX;
         }
 
+        impl<const MIN: $ty, const MAX: $ty> Default for $wrapper<MIN, MAX> {
+            fn default() -> Self {
+                Self(MIN)
```
thibault-martinez: Is MIN a sensible default ? Why ?

pvdrz: we have this idea that `BoundedU32<u32::MIN, u32::MAX>` is the same as `u32` and `u32::default() == u32::MIN`.

thibault-martinez: If the assumption is that we only impl `Bounded` for unsigned types it makes sense, otherwise it would be a bit weird ?

pvdrz: i could do `0.max(MIN)` if that helps.

thibault-martinez: What about `Self($ty::default())` ?

pvdrz: hahaha, there is an interest case. What if we do `UnboundedI32<-100, -50>`. Which one is the good default? `-100` or `-50`?

pvdrz: > What about Self($ty::default()) ?

This will break the invariant most of the time. Check `UnboundedU32<1, U32::MAX>` for example.

thibault-martinez: Ah right, trying to think about this during a call...

pvdrz: I think the "best" default would be something like this:
- `MIN` for unsigned
- `(MIN + MAX)/2` (i.e. the middle point of the range) for signed.

So something like
```rust
if $ty::MIN < 0 {
    (MIN  + MAX) / 2
} else {
    MIN
}
```
what say you?

thibault-martinez: Is `if $ty::MIN < 0` even legal for unsigned ?

pvdrz: Yes, it produces a warning, but yes: https://play.integer32.com/?version=stable&mode=debug&edition=2021&gist=0745beed373a9e99832a3c51e324ff2d

pvdrz: OK i think my last proposal is not the best idea because it doesn't produce an "obvious" value for the default (because of rounding after dividing and so on).

So I just did another commit where we take the closest bounded integer to zero as the default. This means that the default will be zero if zero is in-bounds, the maximum of the range if the range is completely negative and the minimum of the range if the range is completely positive.

bee-common/bee-packable/src/packable/bounded.rs
```diff
@@ -63,6 +63,27 @@ macro_rules! bounded {
             const MAX: $ty = MAX;
         }
 
+        impl<const MIN: $ty, const MAX: $ty> $wrapper<MIN, MAX> {
+            /// Creates a bounded integer without checking wether the value is in-bounds.
```
thibault-martinez:
```suggestion
            /// Creates a bounded integer without checking whether the value is in-bounds.
```

thibault-martinez (dismissed):

Adam-Gleave (dismissed):
Looks good to me, and useful 👍

bee-common/bee-packable/src/packable/bounded.rs
```diff
@@ -63,6 +63,27 @@ macro_rules! bounded {
...
+                debug_assert!((value >= MIN) && (value <= MAX));
+
+                Self(value)
+            }
+        }
+
+        /// This implementation returns the closest bounded integer to zero.
+        impl<const MIN: $ty, const MAX: $ty> Default for $wrapper<MIN, MAX> {
+            fn default() -> Self {
+                // SAFETY: this value is larger or equal than `MIN` and smaller or equal then
```
Adam-Gleave:
```suggestion
                // SAFETY: this value is larger or equal than `MIN` and smaller or equal than
```

Adam-Gleave (dismissed):

thibault-martinez (approved):

Adam-Gleave (approved):

PR #803, opened 2021-10-27
Fix `VecPrefix` constructor

pvdrz:


bee-message/src/payload/indexation/mod.rs
```diff
@@ -20,20 +23,31 @@ use core::{
...
+pub(crate) const PREFIXED_INDEXATION_INDEX_LENGTH_MIN: u32 = *INDEXATION_INDEX_LENGTH_RANGE.start() as u32;
+pub(crate) const PREFIXED_INDEXATION_INDEX_LENGTH_MAX: u32 = *INDEXATION_INDEX_LENGTH_RANGE.end() as u32;
+pub(crate) const PREFIXED_INDEXATION_DATA_LENGTH_MAX: u32 = *MESSAGE_LENGTH_RANGE.end() as u32;
 
-fn unpack_prefix_to_invalid_index_length(err: UnpackPrefixError<Infallible>) -> ValidationError {
+fn unpack_prefix_to_invalid_index_length(
+    err: UnpackPrefixError<
+        Infallible,
+        InvalidBoundedU32<PREFIXED_INDEXATION_INDEX_LENGTH_MIN, PREFIXED_INDEXATION_INDEX_LENGTH_MAX>,
+    >,
```
Adam-Gleave, line 33: I wish this looked nicer 😆

thibault-martinez (dismissed):

Adam-Gleave (dismissed):

bee-message/src/payload/indexation/mod.rs
pvdrz: Me too but someone wanted all the error information :trollface:

pvdrz: I mean, I'd be tempted to not include the value in the error at all just like [`TryFromIntError`](https://doc.rust-lang.org/std/num/struct.TryFromIntError.html) because it is not like we do something with the value or the `MIN..MAX` range.

I think just having one type like `struct InvalidBoundedInteger;` for all the bounded integer conversion errors is enough.

thibault-martinez (approved):

PR #814, opened 2021-11-03
Add `Packable::unpack` variants

pvdrz:
# Description of change

This change updates `Packable::unpack` to be generic over the constant `VERIFY: bool` allowing each implementation to specify which checks can be skipped when `CHECK == false`. It also adds two new default methods `unpack_verified` and `unpack_unverified` which are equivalent to `unpack::<_, true>` and `unpack::<_, false>` respectively, but using `SliceUnpacker` instead of a generic `Unpacker`.

## Type of change

All the `bee` crates using `bee-packable` were updated and their tests are passing.

## Change checklist

### DIFF
--- a/bee-common/bee-packable/src/packable/mod.rs
+++ b/bee-common/bee-packable/src/packable/mod.rs
@@ -50,7 +50,7 @@ pub trait Packable: Sized {
     ) -> Result<Self, UnpackError<Self::UnpackError, U::Error>>;
 }
 
-/// Extension trait for types that implement `Packable`.
+/// Extension trait for types that implement [`Packable`].
 pub trait PackableExt: Packable {
     /// Convenience method that packs this value into a [`Vec<u8>`].
     fn pack_to_vec(&self) -> Vec<u8>;
--- a/bee-common/bee-packable/src/packable/mod.rs
+++ b/bee-common/bee-packable/src/packable/mod.rs
@@ -55,18 +55,13 @@ pub trait PackableExt: Packable {
     /// Convenience method that packs this value into a [`Vec<u8>`].
     fn pack_to_vec(&self) -> Vec<u8>;
 
-    /// Unpacks this value from the given [`Unpacker`] doing syntactical checks.
-    fn unpack_checked<U: Unpacker>(
-        unpacker: &mut U,
-    ) -> Result<Self, UnpackError<<Self as Packable>::UnpackError, U::Error>>;
+    /// Unpacks this value from a sequence of bytes doing syntactical checks.
+    fn unpack_verified<T: AsRef<[u8]>>(
+        bytes: T,
+    ) -> Result<Self, UnpackError<<Self as Packable>::UnpackError, UnexpectedEOF>>;
 
-    /// Unpacks this value from the given [`Unpacker`] without doing syntactical checks.
-    fn unpack_unchecked<U: Unpacker>(
-        unpacker: &mut U,
-    ) -> Result<Self, UnpackError<<Self as Packable>::UnpackError, U::Error>>;
-
-    /// Unpacks this value from a type that implements [`AsRef<[u8]>`].
-    fn unpack_from_slice<T: AsRef<[u8]>>(
+    /// Unpacks this value from a sequence of bytes without doing syntactical checks.
+    fn unpack_unverified<T: AsRef<[u8]>>(
         bytes: T,
     ) -> Result<Self, UnpackError<<Self as Packable>::UnpackError, UnexpectedEOF>>;
 }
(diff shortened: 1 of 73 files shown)


bee-common/bee-packable/src/packable/mod.rs
```diff
@@ -61,11 +61,23 @@ pub trait Packable: Sized {
     }
 
     /// Unpacks this value from the given [`Unpacker`].
-    fn unpack<U: Unpacker>(unpacker: &mut U) -> Result<Self, UnpackError<Self::UnpackError, U::Error>>;
+    fn unpack<U: Unpacker, const CHECK: bool>(
```
grtlr: I'm not sure if we should expose `CHECK: bool` in a public trait interface, because it feels like an implementation detail. Wouldn't `unpack_checked` and `unpack_unchecked` suffice, while keeping `unpack<..,const CHECK: bool>` internally for the more complex unpacking cases?

pvdrz: Hmmm... I agree that having `CHECK` exposed at the same time as `unpack_(un)?checked` is redundant but those last are supposed to be just helper methods like `pack_to_vec`. 

The other thing is that I'm not sure how to keep `CHECK` hidden but at the same time allow the user to implement it.

grtlr: Would most users use `unpack_(un)?checked` or `unpack<U: Unpacker, const CHECK: bool>` in their code? As far as I understand, the purpose of the latter is to only have one method that the users of the trait have to implement?

Adam-Gleave: The older API is `unpack<R: Read, const CHECK: bool>` (in cpt2/dev), so that may be more familiar to existing users.

I guess another thing to consider is because `unpack_unchecked` and `unpack_checked` aren't hidden, a user could effectively implement all three, with three different behaviours, which doesn't make sense. While that situation isn't likely, and I'm not sure whether it actually needs considering, it might make more sense to either offer `unpack`, or `unpack_unchecked` and `unpack_checked`, rather than all three.

pvdrz: Maybe we can do something similar to what we did with the `coerce` module and create a `PackableExt` trait that is only implemented for `P: Packable` and seal it so the user cannot override its behaviour. What say you?

EDIT: check 6ae0c31e7238e1410d02f5164298cb0a32825453

bee-common/bee-packable/src/packable/mod.rs
```diff
@@ -44,7 +44,29 @@ pub trait Packable: Sized {
     /// Packs this value into the given [`Packer`].
     fn pack<P: Packer>(&self, packer: &mut P) -> Result<(), P::Error>;
 
+    /// Unpacks this value from the given [`Unpacker`].
+    fn unpack<U: Unpacker, const VERIFY: bool>(
+        unpacker: &mut U,
+    ) -> Result<Self, UnpackError<Self::UnpackError, U::Error>>;
+}
+
+/// Extension trait for types that implement `Packable`.
```
thibault-martinez: Put the brackets for documentation

bee-common/bee-packable/tests/bounded.rs
```diff
@@ -51,9 +51,13 @@ macro_rules! impl_packable_test_for_bounded_integer {
         #[test]
         fn $packable_invalid_name() {
             let bytes = vec![0u8; core::mem::size_of::<$wrapped>()];
-            let unpacked = <$wrapper>::unpack_from_slice(&bytes);
+            let unpacked = <$wrapper>::unpack_verified(&bytes);
 
-            assert!(matches!(unpacked, Err(UnpackError::Packable($error(0)))))
+            assert!(
+                matches!(unpacked, Err(UnpackError::Packable($error(0)))),
+                "found {:?}",
```
thibault-martinez: Was this debug or ... ?

pvdrz: it was debug but maybe we should do it in every `assert!(matches!(...))`.

bee-node/bee-plugin/src/hotloader.rs
```diff
@@ -98,7 +98,7 @@ impl PluginHotloader {
             // Removes info for files that no longer exist and loads plugins whose files did not exist before.
             self.sync_plugin_info(to_remove, last_writes).await?;
 
-            sleep(Duration::from_millis(PLUGIN_CHECK_INTERVAL_MILLIS)).await;
+            sleep(Duration::from_millis(PLUGIN_VERIFY_INTERVAL_MILLIS)).await;
```
thibault-martinez: Oh no

bee-common/bee-packable/tests/bounded.rs
thibault-martinez: Alright, I wonder to what extent it's useful if we're expecting an error

pvdrz: I did it because I changed the errors and couldn't figure out the right variants. But future PR maybe.

bee-node/bee-plugin/src/hotloader.rs
pvdrz: Oh lord. I've been betrayed by my own devices

thibault-martinez (approved):

PR #820, opened 2021-11-04
Add check script

pvdrz:
# Description of change

This PR adds a `check.sh` script that could be run by contributors before opening a pull request to be (almost) sure that their changes will pass our CI process.

## Type of change

The script was tested over the `coordicide` branch and ran successfully.

## Change checklist


check.sh
```diff
@@ -0,0 +1,31 @@
...
+    status_code=$?
+    if [[ $status_code -eq 0 ]] ; then
+        echo $cmd was succesful!
+    else
+        echo $cmd failed!
+    fi
+    return $status_code
+}
+
+#
```
Adam-Gleave: Should this be a comment? ("If no arguments", etc)

Adam-Gleave (commented):
Do we definitely want this in the root? (It doesn't bother me, but might be worth discussing).

It also may be worth renaming to something like `checks.sh` or `all_checks.sh`, to differentiate it from `cargo check` or the check workflow specifically.

pvdrz:
> Do we definitely want this in the root? (It doesn't bother me, but might be worth discussing).
> 
> It also may be worth renaming to something like `checks.sh` or `all_checks.sh`, to differentiate it from `cargo check` or the check workflow specifically.

I like `all_checks.sh` :)

check.sh
pvdrz: Indeed

thibault-martinez:
> Do we definitely want this in the root? (It doesn't bother me, but might be worth discussing).

No we don't :trollface:

pvdrz:
@thibault-martinez where would you put it?

thibault-martinez:
> @thibault-martinez where would you put it?

Maybe simply a `scripts` folder at the root ?

pvdrz:
SGTM

thibault-martinez (approved):

PR #837, opened 2021-11-16
Add default implementation for `Packable::packed_len`

### LINKED ISSUE #836: `Packable::packed_len` is a particular case of `Packable::pack` (almost?) always.
## Description

Introduce a new type `LengthPacker`:
```rust
pub(crate) struct LengthPacker(pub(crate) usize);

impl Packer for LengthPacker {
    type Error = Infallible;

    fn pack_bytes<B: AsRef<[u8]>>(&mut self, bytes: B) -> Result<(), Self::Error> {
        self.0 += bytes.as_ref().len();
        Ok(())
    }
}
```
And then add a default implementation for `Packable::packed_len`:
```rust
pub trait Packable: Sized {
    ...
    fn packed_len(&self) -> usize {
        let mut packer = LengthPacker(0);

        match self.pack(&mut packer) {
            Ok(()) => packer.0,
            Err(err) => match err {},
        }
    }
    ...
}
```

## Motivation

Given that `Packable::pack` and `Packable::packed_len` must be consistent between each other. It makes sense to let `Packable::packed_len` have a default implementation that just calls `Packable::pack` using `LengthPacker`.

This should reduce the amount of code written while implementing `Packable` manually and remove some code from `bee-packable-derive`.

### Is this default implementation slower than the existing implementations?

We would have to check if the generated code is optimized well enough to be equivalent to the current implementations.

### Should we move `packed_len` to `PackableExt`?

I think that this implementation of `packed_len` is semantically correct all the time because it must be consistent with `pack`. In that order of ideas, this implementation should live inside `PackableExt` to forbid anyone from overriding it. On the other hand, it would be nice to be allowed to return `core::mem::size_of<T>()` for the integral types anyway.
 
Yes, but I'll be happy to mentor anyone from the community to implement this.

pvdrz:
fixes #836.

### DIFF
--- a/bee-common/bee-packable/src/packer/mod.rs
+++ b/bee-common/bee-packable/src/packer/mod.rs
@@ -5,9 +5,11 @@
 
 #[cfg(feature = "io")]
 mod io;
+mod len;
 mod slice;
 mod vec;
 
+pub(crate) use len::LenPacker;
 pub use slice::SlicePacker;
 pub use vec::VecPacker;
 
--- a/bee-common/bee-packable/src/packer/mod.rs
+++ b/bee-common/bee-packable/src/packer/mod.rs
@@ -10,6 +10,7 @@ mod slice;
 mod vec;
 
 pub(crate) use len::LenPacker;
+
 pub use slice::SlicePacker;
 pub use vec::VecPacker;
 
--- a/bee-message/src/unlock/unlock_blocks.rs
+++ b/bee-message/src/unlock/unlock_blocks.rs
@@ -73,6 +73,7 @@ impl Deref for UnlockBlocks {
 impl Packable for UnlockBlocks {
     type UnpackError = MessageUnpackError;
 
+    #[inline(always)]
     fn pack<P: Packer>(&self, packer: &mut P) -> Result<(), P::Error> {
         self.0.pack(packer)
     }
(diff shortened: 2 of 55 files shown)


bee-message/src/unlock/unlock_blocks.rs
```diff
@@ -73,10 +73,7 @@ impl Deref for UnlockBlocks {
 impl Packable for UnlockBlocks {
     type UnpackError = MessageUnpackError;
 
-    fn packed_len(&self) -> usize {
-        self.0.packed_len()
-    }
-
+    #[inline(always)]
```
thibault-martinez, line 76: Why is this the only inline ?

pvdrz: wdym by only? I decided to inline all the `pack` impls that only call `pack` once.

thibault-martinez: Alright 👍

bee-common/bee-packable/src/packer/mod.rs
```diff
@@ -5,9 +5,11 @@
 
 #[cfg(feature = "io")]
 mod io;
+mod len;
 mod slice;
 mod vec;
 
+pub(crate) use len::LenPacker;
```
thibault-martinez, line 12:
```suggestion
pub(crate) use len::LenPacker;

```

thibault-martinez (approved):

Adam-Gleave (approved):

PR #845, opened 2021-11-19
Use shorter validation errors inside `bee-message`

pvdrz:
Introduce new alias types for each bounded integer type used inside `bee-message` and rewrite the validation error types using the `<BoundedInt as TryFrom<usize>>::Error` syntax.


thibault-martinez:
Rebase to fix udeps ?

pvdrz:
We need to wait for a new version of `cargo-udeps` that includes https://github.com/est31/cargo-udeps/commit/8f9851044f3f11200193b7d5886ec7b4d8851ba1

bee-common/bee-packable/tests/boxed_slice_prefix.rs
```diff
@@ -18,18 +18,15 @@ fn boxed_slice_prefix_from_boxed_slice_invalid_error() {
...
+    assert!(matches!(prefixed, Err(TryIntoBoundedU32Error::Invalid(16))));
 }
 
 #[test]
 fn boxed_slice_prefix_from_boxed_slice_truncated_error() {
     let boxed_slice = vec![0u8; 257].into_boxed_slice();
     let prefixed = BoxedSlicePrefix::<u8, u8>::try_from(boxed_slice);
 
-    assert!(matches!(prefixed, Err(TryIntoPrefixError::Truncated(257))));
+    assert!(prefixed.is_err());
```
thibault-martinez, line 29: Why are we being less specific now ?

bee-message/src/payload/transaction/essence.rs
```diff
@@ -24,10 +24,8 @@ use core::{convert::Infallible, fmt};
 /// Length (in bytes) of Transaction Essence pledge IDs (node IDs relating to pledge mana).
 pub const PLEDGE_ID_LENGTH: usize = 32;
 
-pub(crate) const PREFIXED_INPUTS_LENGTH_MAX: u32 = *INPUT_COUNT_RANGE.end() as u32;
-pub(crate) const PREFIXED_OUTPUTS_LENGTH_MAX: u32 = *OUTPUT_COUNT_RANGE.end() as u32;
-pub(crate) const PREFIXED_INPUTS_LENGTH_MIN: u32 = *INPUT_COUNT_RANGE.start() as u32;
-pub(crate) const PREFIXED_OUTPUTS_LENGTH_MIN: u32 = *OUTPUT_COUNT_RANGE.start() as u32;
+pub(crate) type InputCount = BoundedU32<{ *INPUT_COUNT_RANGE.start() as u32 }, { *INPUT_COUNT_RANGE.end() as u32 }>;
```
thibault-martinez, line 27: You sometime put plural and sometime singular before *Count, I suggest using plural. Can you have a look and harmonise them all please ?

pvdrz: I kept the name consistent with the constants that were defined before my changes. If the constant said `COOKIE(S?)`, the type is called `Cookie(s?)`. I can change it if you want nevertheless

bee-common/bee-packable/tests/boxed_slice_prefix.rs
pvdrz: because`<u8 as TryFrom<usize>>::Error` is  [`TryFromIntError`](https://doc.rust-lang.org/std/num/struct.TryFromIntError.html) and its only field is private, so we cannot use `matches!` with it.

thibault-martinez: Ok

bee-message/src/payload/transaction/essence.rs
thibault-martinez: Would be nice if we could yes. What do you think make the most sense ? Singular or plural ?

pvdrz: I'll check it and let you know because I'm not sure.

Edit: Singular seems to be the grammatically correct option.

bee-common/bee-packable/src/packable/bounded.rs
```diff
@@ -132,35 +171,55 @@ macro_rules! bounded {
...
 impl Bounded for u8 {
     type Bounds = Self;
+    type Range = Range<Self>;
 
     const MIN: Self::Bounds = u8::MIN;
     const MAX: Self::Bounds = u8::MAX;
+
+    fn range(start: Self::Bounds, end: Self::Bounds) -> Self::Range {
+        start..end
+    }
```
thibault-martinez: Shouldn't this be a default implementation ?

bee-common/bee-packable/src/packable/prefix.rs
```diff
@@ -104,102 +73,80 @@ impl<T, B: Bounded> Default for VecPrefix<T, B> {
...
+
+    fn unpack<U: Unpacker, const VERIFY: bool>(
+        unpacker: &mut U,
+    ) -> Result<Self, UnpackError<Self::UnpackError, U::Error>> {
+        // The length of any dynamically-sized sequence must be prefixed.
+        let len = B::unpack::<_, VERIFY>(unpacker)
+            .map_packable_err(UnpackPrefixError::Prefix)?
+            .into();
+
+        let mut inner = Vec::with_capacity(len.try_into().unwrap_or(0));
```
thibault-martinez, line 151: Maybe i'm just tired but what exactly would it mean if that was failing ?

pvdrz: It's a bit of a corner case that ends up failing almost all the time anyway:  If `B = u64` and we were in a 32-bit platform then it might for any  `len > u32::MAX`. If that happens you create a `0` capacity vector and then you start pushing elements of type `T` to it until you run out of memory.

The only way OOM doesn't happen is if `T` has size zero. So, in principle you can decode `len` values of type `T` without reaching memory exhaustion. However, I'm not sure what ends up happening there. I'll test.

Update: It panics anyway due to a capacity overflow https://play.integer32.com/?version=stable&mode=release&edition=2021&gist=3dab0bcbd20cfedf3dcedd0b303c1825

To be honest it is not clear for me what should be the "correct" behavior here. Technically you can pack a `VecPrefix<u8, u64>` of length `u32::MAX + 1` in a 64-bit platform but then you cannot unpack that value in a 32-bit platform. However, that doesn't happen due to the bytes being invalid, it is just a platform restriction.

bee-common/bee-packable/src/packable/bounded.rs
pvdrz: I don't see any reason why it shouldn't. I'll try to change it and see what happens.

Edit: It is because `Range<T>: Iterator<Item = T>` if `T: Step`. But that trait is behind a [feature gate](https://doc.rust-lang.org/std/iter/trait.Step.html) so we cannot use it in our bounds. 

But you made me realize that we can remove the range from the trait and just put those bounds somewhere else by putting `Range<T>: Iterator<Item = T>` directly in a `where` clause. I'll add a commit fixing this now.

thibault-martinez (approved):

PR #856, opened 2021-11-25
Improve CI by combining workflows

### LINKED ISSUE #854: Improvements to the CI
- [ ] Only triggers when Rust or Cargo.toml/lock files are changed
- [ ] Investigate why crates are being recompiled even when cache has been recovered
- [ ] Merge check and test workflows ?

grtlr:
# Description of change

The goal of this PR is to fix some of the problems with our CI and hopefully also improve performance. It adds the following changes

* Only run CI on changes to the code (exclude `documentation/` and `**.md` files);
* Job-based system;
  * Combine `build` and `test` workflows into a single job;
  * Run other checks only if those have been successful;
 * More generous caching of dependencies and build artifacts;

* closes #854 

## Type of change

The caching is tricky to test, but I have seen significant speed ups. Also, the GitHub Actions UI is a bit cleaner now.

## Change checklist


.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ on:
       - stable
 
 jobs:
-  test:
+  setup:
```
thibault-martinez: This ends up being the name shown in the interface so `setup` is not that great. You can either rename this or add a `name` property I think

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Test
+name: CI
```
thibault-martinez, line 1: Sorry I meant this one, `CI` is a too broad name for this IMO, all workflows are CI, maybe this one should be `Build/Test` or something ?

grtlr: I think now that almost all jobs run in this workflow the name is appropriate, but I can also change it if you would prefer that.

pvdrz (dismissed):

grtlr:
Another nice effect is that we get better reporting within GitHub Action. Here is an example: https://github.com/iotaledger/bee/actions/runs/1503959223. We now can for example cancel almost the complete CI for a PR with a single click. Before we had to do this manually for each workflow.

Alex6323 (dismissed):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,265 @@
...
+            ${{ matrix.os }}-${{ matrix.rust }}-
+            ${{ matrix.os }}-
+
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-targets --all-features -- -D warnings
+
+  format:
+    needs: build-and-test
```
thibault-martinez: Is this really necessary ? I'd rather have this information as soon as possible.

grtlr: Right now it first does a build. If that fails it gets reported immediately. To run the tests, you would first have to build  the individual crates anyway. A syntactic validation of testcases might take a littler longer, but in all other cases I think it shouldn't make a big difference time-wise. One downside is that the reporting is not as straight forward as it used to be (I think you have to click on the check to see which command went wrong).

But I could also be missing something. Why do you think the combined workflow would be slower?

thibault-martinez: Just to be sure, I was speaking about the `format` check. It doesn't technically need build and test to perform. And I'm wondering if having to wait for them both just to discover that there is a formatting issue is wise or not.

grtlr: Oh, sorry that was my mistake, I should get a coffee. I guess it is always a tradeoff: Spinning up machines takes time and prevents other jobs to run in the meantime. But I guess format runs fast enough to put it towards the front 👍 .

thibault-martinez: All good, it runs in seconds when all the rest run in more than 15mins, I think it's fine to have it run first

.github/workflows/ci.yml
```diff
@@ -0,0 +1,265 @@
...
+      - dev
+      - stable
+    paths-ignore:
+      - 'documentation/**'
+      - '**.md'
+
+jobs:
+  build-and-test:
+    name: Build and Test
+    runs-on: ${{ matrix.os }}
```
thibault-martinez: It triggers me a tiny bit that we use matrix before defining it :trollface:

grtlr: Ah the beauty of declarative languages 😄

.github/workflows/ci.yml
```diff
@@ -0,0 +1,265 @@
...
+
+    - name: Cache cargo state
+      uses: actions/cache@v2
+      env:
+        cache-name: cache-cargo
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+        key: ${{ matrix.os }}-${{ matrix.rust }}-build-${{ env.cache-name }}-${{ hashFiles('**/Cargo.toml') }}
```
thibault-martinez: I think that `build` part is a relic from the previous version ? It should be covered by the cache name now.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,265 @@
...
+          ${{ matrix.os }}-${{ matrix.rust }}-build-${{ env.cache-name }}-
+          ${{ matrix.os }}-${{ matrix.rust }}-build-
+          ${{ matrix.os }}-${{ matrix.rust }}-
+          ${{ matrix.os }}-
+
+    - name: Build
+      uses: actions-rs/cargo@v1
+      with:
+        command: build
+        args: --all --release
```
thibault-martinez: Wonder if there should be --all--features as well 🤔

.github/workflows/ci.yml
```diff
@@ -0,0 +1,265 @@
...
+            ubuntu-latest-nightly-
+            ubuntu-latest-
+
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-targets --all-features -- -D warnings
+
+  udeps:
+    needs: [build-and-test,clippy]
```
thibault-martinez:
```suggestion
    needs: [build-and-test, clippy]
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,265 @@
...
+          command: install
+          args: cargo-udeps
+
+      - uses: actions-rs/cargo@v1
+        with:
+          command: udeps
+          args: --all-targets --all-features --backend=depinfo
+
+  licenses:
+    needs: [build-and-test,clippy]
```
thibault-martinez:
```suggestion
    needs: [build-and-test, clippy]
```

.github/workflows/ci.yml
grtlr: We need it for more granularity in the caching. But it was in the wrong position here, so thanks for pointing that out!

Alex6323 (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,226 @@
...
+      - name: Cache cargo state
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-cargo
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            ~/.cargo/bin
+          key: ${{ matrix.os }}-${{ matrix.rust }}-${{ env.cache-name }}-${{ hashFiles('**/Cargo.toml') }}-build
```
thibault-martinez, line 57: Why is this `-build` when it's later `-other` ?

grtlr: We don't want the `clippy` job to overwrite our `build` cache (which was more expensive to create).

thibault-martinez (approved):

pvdrz (approved):

PR #857, opened 2021-11-30
Avoid wrappers for `no_std` packers and unpackers

pvdrz:
# Description of change

This PR removes the `VecPacker` and `SliceUnpacker` types and implements `Packer` and `Unpacker` for `Vec<u8>` and `&[u8]` directly.

In contrast, it also adds two new wrapper types `IoPacker` and `IoUnpacker` that implement `Packer` or `Unpacker` if the wrapped type implements `Read` or `Write` respectively.

This is done because using the crate without `std` should be the sensible default and it should be as simple as possible. In contrast, using `std` requires an extra step to make it work now.

## Type of change
- Breaking change (fix or feature that would cause existing functionality to not work 

The whole branch compiles and passes tests without issue.

## Change checklist

### DIFF
--- a/bee-common/bee-packable/tests/common/mod.rs
+++ b/bee-common/bee-packable/tests/common/mod.rs
@@ -1,7 +1,11 @@
 // Copyright 2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use bee_packable::{packer::SlicePacker, Packable, PackableExt};
+use bee_packable::{
+    packer::{IoPacker, SlicePacker},
+    unpacker::IoUnpacker,
+    Packable, PackableExt,
+};
 
 use core::fmt::Debug;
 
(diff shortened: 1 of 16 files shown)


bee-common/bee-packable/tests/common/mod.rs
```diff
@@ -49,16 +46,16 @@ where
...
-    packable.pack(&mut vec_packer).unwrap();
-    let unpacked = P::unpack_verified(&mut vec_packer.as_slice()).unwrap();
+    let mut vec = Vec::new();
+    packable.pack(&mut vec).unwrap();
+    let unpacked = P::unpack_verified(&mut vec.as_slice()).unwrap();
     assert_eq!(packable, &unpacked);
 
     // Tests for Read and Write
 
     let mut vec = Vec::new();
```
thibault-martinez: This should use `IoPacker` and `IoUnpacker`

thibault-martinez (approved):

PR #873, opened 2021-12-10
Byte cost for message outputs

grtlr:
# Description of change

Byte cost is a central concept in the upcoming changes to the dust protection mechanism ([corresponding RFC](https://github.com/iotaledger/protocol-rfcs/pull/39)). This PR implements a procedural macro that aims at making it easy to derive the byte cost for the new UTXO output types of future IOTA versions.

## Type of change

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist

## Process

The following is an overview of the types that have been implemented and the challenges that remain to be solved.

### Challenges

* [x] `DustDepositFeatureBlock` validation does not make sense in Stardust;
* [x] `MetadataFeatureBlock` and `IndexationFeatureBlock` contain raw `Box<[u8]>`s for which `Packable` is not defined;
* [x] Can't find `TokenId` in TIP (it's from https://github.com/lzpap/protocol-rfcs/blob/master/tips/TIP-0018/tip-0018.md)
* [x] We use the `impl_id` macro for ids that have a byte cost and ones that don't (`MessageId`), making annotation harder.

### Implementation

* [x] Feature Blocks
  * [x] `SenderFeatureBlock`
  * [x] `TimelockMilestoneIndexFeatureBlock`
  * [x] `TimelockUnixFeatureBlock`
  * [x] `ExpirationMilestoneFeatureBlock`
  * [x] `ExpirationUnixBlock`
  * [x] `MetadataFeatureBlock`
  * [x] `IndexationFeatureBlock`
  * [x] `IssuerFeatureBlock`
  * [x] `DustDepositFeatureBlock`
* [x] Outputs
  * [x] `AliasOutput`
  * [x] `SimpleOutput`
  * [x] `FoundryOutput`
  * [x] `TreasuryOutput`
  * [x] `ExtendedOutput`
  * [x] `NftOutput`

### DIFF
--- a/bee-message/bee-byte-cost-derive/Cargo.toml
+++ b/bee-message/bee-byte-cost-derive/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "bee-byte-cost-derive"
+version = "0.1.0"
+authors = [ "IOTA Stiftung" ]
+edition = "2021"
+description = "Internal crate that adds derive macros for computing the byte cost of outputs."
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "tangle", "bee", "framework", "bytecost", "dust" ]
+homepage = "https://www.iota.org"
+publish = false
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro-error = { version = "1.0.4", default-features = false, features = [ "syn-error" ] }
+proc-macro2 = { version = "1.0.30", default-features = false }
+quote = { version = "1.0.10", default-features = false }
+syn = { version = "1.0.80", default-features = false, features = [ "clone-impls", "derive", "parsing", "printing", "proc-macro" ] }
+
+[dev-dependencies]
+bee-byte-cost = { path = "../bee-byte-cost", version = "0.1.0", default-features = false }
+packable = { version = "0.1.0", default-features = false }
+
+trybuild = { version = "1.0.50", default-features = false, features = [ "diff" ] }
--- a/bee-message/bee-byte-cost-derive/src/builder.rs
+++ b/bee-message/bee-byte-cost-derive/src/builder.rs
@@ -1,0 +1,45 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use proc_macro2::TokenStream;
+use quote::{quote, ToTokens};
+use syn::Ident;
+
+pub(crate) struct ByteCostBuilder {
+    ident: Ident,
+    pub key: TokenStream,
+    pub data: TokenStream,
+    pub derived: TokenStream,
+}
+
+impl ByteCostBuilder {
+    pub(crate) fn new(ident: Ident) -> Self {
+        Self {
+            ident,
+            key: quote!{ 0 },
+            data: quote!{ 0 },
+            derived: quote!{ 0 },
+        }
+    }
+}
+
+impl ToTokens for ByteCostBuilder {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let ident = &self.ident;
+        let key = &self.key;
+        let data = &self.data;
+        let derived = &self.derived;
+
+        let output = quote! {
+            impl ByteCost for #ident {
+                fn weighted_bytes(&self, config: &bee_byte_cost::ByteCostConfig) -> u64 {
+                    (#key) * config.weight_for_key +
+                    (#data) * config.weight_for_data +
+                    #derived // they have already been weighted
+                }
+            }
+        };
+
+        output.to_tokens(tokens);
+    }
+}
--- a/bee-message/bee-byte-cost-derive/src/builder.rs
+++ b/bee-message/bee-byte-cost-derive/src/builder.rs
@@ -1,4 +1,4 @@
-// Copyright 2021 IOTA Stiftung
+// Copyright 2021-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
 use proc_macro2::TokenStream;
--- a/bee-message/bee-byte-cost-derive/src/builder.rs
+++ b/bee-message/bee-byte-cost-derive/src/builder.rs
@@ -16,9 +16,9 @@ impl ByteCostBuilder {
     pub(crate) fn new(ident: Ident) -> Self {
         Self {
             ident,
-            key: quote!{ 0 },
-            data: quote!{ 0 },
-            derived: quote!{ 0 },
+            key: quote! { 0 },
+            data: quote! { 0 },
+            derived: quote! { 0 },
         }
     }
 }
--- a/bee-message/bee-byte-cost-derive/src/enum_info.rs
+++ b/bee-message/bee-byte-cost-derive/src/enum_info.rs
@@ -1,4 +1,4 @@
-// Copyright 2021 IOTA Stiftung
+// Copyright 2021-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
--- a/bee-message/bee-byte-cost-derive/src/enum_info.rs
+++ b/bee-message/bee-byte-cost-derive/src/enum_info.rs
@@ -1,0 +1,69 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    builder::ByteCostBuilder,
+    field_info::{fields_to_info, FieldInfo},
+    weight::filter_attrs,
+};
+
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote, ToTokens};
+use syn::{Ident, Result, Variant};
+
+pub(crate) struct VariantInfo {
+    ident: Ident,
+    fields: Vec<FieldInfo>,
+}
+
+impl VariantInfo {
+    pub(crate) fn new(variant: &Variant) -> Result<Self> {
+        if filter_attrs(&variant.attrs).next().is_some() {
+            return Err(syn::Error::new(
+                variant.ident.span(),
+                "weight types cannot be applied to variants of an enum",
+            ));
+        }
+
+        Ok(Self {
+            fields: fields_to_info(&variant.fields)?,
+            ident: variant.ident.clone(),
+        })
+    }
+}
+
+impl ToTokens for VariantInfo {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let ident = &self.ident;
+        let variant_fields = (0..self.fields.len()).map(|n| format_ident!("field_{}", n));
+        let variant_fields_2 = variant_fields.clone();
+        let match_arm = quote! {
+            Self::#ident(#(#variant_fields),*) => { #(bee_byte_cost::ByteCost::weighted_bytes(#variant_fields_2, config))+* },
+        };
+        match_arm.to_tokens(tokens);
+    }
+}
+
+pub(crate) struct EnumInfo(Vec<VariantInfo>);
+
+impl EnumInfo {
+    pub(crate) fn new(variants: impl IntoIterator<Item = Variant>) -> Result<Self> {
+        let variants = variants
+            .into_iter()
+            .map(|v| VariantInfo::new(&v))
+            .collect::<Result<Vec<VariantInfo>>>();
+        Ok(Self(variants?))
+    }
+
+    pub(crate) fn add_to(self, builder: &mut ByteCostBuilder) {
+        let match_arms = self.0.iter();
+
+        // TODO: Don't hard code the length of KIND;
+        let match_tokens = quote! {
+            + 1 + match self {
+                #(#match_arms)*
+            }
+        };
+        match_tokens.to_tokens(&mut builder.derived);
+    }
+}
--- a/bee-message/bee-byte-cost-derive/src/field_info.rs
+++ b/bee-message/bee-byte-cost-derive/src/field_info.rs
@@ -1,4 +1,4 @@
-// Copyright 2021 IOTA Stiftung
+// Copyright 2021-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
--- a/bee-message/bee-byte-cost-derive/src/field_info.rs
+++ b/bee-message/bee-byte-cost-derive/src/field_info.rs
@@ -1,0 +1,96 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    builder::ByteCostBuilder,
+    weight::{parse_byte_cost_attributes, Weight},
+};
+
+use proc_macro2::TokenStream;
+use quote::{quote_spanned, ToTokens};
+use syn::{spanned::Spanned, Field, Fields, FieldsNamed, FieldsUnnamed, Ident, Index, Result};
+
+pub(crate) enum IdentOrIndex {
+    Ident(Ident),
+    Index(Index),
+}
+
+impl ToTokens for IdentOrIndex {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        match self {
+            Self::Ident(ident) => ident.to_tokens(tokens),
+            Self::Index(index) => index.to_tokens(tokens),
+        }
+    }
+}
+
+pub(crate) struct FieldInfo {
+    pub(crate) tags: Vec<Weight>,
+    pub(crate) pattern_ident: IdentOrIndex,
+}
+
+// TODO: Need better name
+pub(crate) fn fields_to_info(fields: &Fields) -> Result<Vec<FieldInfo>> {
+    match fields {
+        Fields::Named(FieldsNamed { named, .. }) => named.into_iter().map(FieldInfo::new_named).collect(),
+        Fields::Unnamed(FieldsUnnamed { unnamed, .. }) => unnamed
+            .into_iter()
+            .enumerate()
+            .map(|(index, field)| FieldInfo::new_unnamed(field, index))
+            .collect(),
+        unit @ Fields::Unit => Err(syn::Error::new(
+            unit.span(),
+            "the `ByteCost` trait cannot be derived for unit fields",
+        )),
+    }
+}
+
+impl ToTokens for FieldInfo {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let pattern_ident = &self.pattern_ident;
+
+        let token_stream = if self.tags.is_empty() {
+            quote_spanned! {pattern_ident.span()=> + bee_byte_cost::ByteCost::weighted_bytes(&self.#pattern_ident, config) }
+        } else {
+            quote_spanned! {pattern_ident.span()=> + packable::PackableExt::packed_len(&self.#pattern_ident) as u64 }
+        };
+
+        token_stream.to_tokens(tokens)
+    }
+}
+
+impl FieldInfo {
+    pub(crate) fn new_named(field: &Field) -> Result<Self> {
+        // Safety: We can unwrap here because we are looking at a named field.
+        let pattern_ident = IdentOrIndex::Ident(field.ident.as_ref().unwrap().clone());
+        Ok(Self {
+            tags: parse_byte_cost_attributes(&field.attrs)?,
+            pattern_ident,
+        })
+    }
+
+    pub(crate) fn new_unnamed(field: &Field, index: usize) -> Result<Self> {
+        let pattern_ident = IdentOrIndex::Index(Index {
+            index: index as u32,
+            span: field.span(),
+        });
+        Ok(Self {
+            tags: parse_byte_cost_attributes(&field.attrs)?,
+            pattern_ident,
+        })
+    }
+
+    pub(crate) fn add_to(self, builder: &mut ByteCostBuilder) {
+        let field = self.to_token_stream();
+        if self.tags.is_empty() {
+            field.to_tokens(&mut builder.derived);
+        } else {
+            for tag in self.tags.iter() {
+                match tag {
+                    Weight::Data => field.to_tokens(&mut builder.data),
+                    Weight::Key => field.to_tokens(&mut builder.key),
+                }
+            }
+        }
+    }
+}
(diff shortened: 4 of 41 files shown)


bee-message/bee-bytecost-derive/src/fields.rs
```diff
@@ -0,0 +1,87 @@
...
+                    let mut data_fields = Vec::new();
+                    let mut untagged_fields = Vec::new();
+
+                    for f in named {
+                        if f.attrs.is_empty() {
+                            untagged_fields.push(f.clone());
+                        }
+
+                        for a in f.attrs.iter() {
+                            if a.path.is_ident(&Ident::new("key", Span::call_site())) {
```
pvdrz: iirc you can do `a.path.is_ident("key")`

bee-message/bee-bytecost-derive/src/fields.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub(crate) fn new(input: DeriveInput) -> syn::Result<Self> {
+        match input.data {
+            Data::Struct(s) => match s.fields {
+                Fields::Named(FieldsNamed { named, .. }) => {
+                    let mut key_fields = Vec::new();
+                    let mut data_fields = Vec::new();
+                    let mut untagged_fields = Vec::new();
+
+                    for f in named {
+                        if f.attrs.is_empty() {
```
pvdrz: I think this might be problematic if there are any other attributes on the field from other macros.

bee-message/bee-bytecost-derive/src/fields.rs
```diff
@@ -0,0 +1,87 @@
...
+                        ident: input.ident,
+                        key_fields,
+                        data_fields,
+                        untagged_fields,
+                    })
+                }
+                _ => unimplemented!(),
+            },
+            Data::Enum(data) => {
+                todo!()
```
pvdrz: todo!

bee-message/bee-bytecost-derive/src/fields.rs
```diff
@@ -0,0 +1,87 @@
...
+                "The `ByteCost` trait cannot be derived for unions.",
+            )),
+        }
+    }
+}
+
+impl ToTokens for AnnotatedFields {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let ident = &self.ident;
+        let key_fields = self.key_fields.iter().map(|f| f.ident.as_ref().unwrap());
```
pvdrz: why is it ok to `unwrap` here? (same for the other two variables defined below)

bee-message/bee-bytecost-derive/src/fields.rs
```diff
@@ -0,0 +1,87 @@
...
+        let data_fields = self.data_fields.iter().map(|f| f.ident.as_ref().unwrap());
+        let untagged_fields = self.untagged_fields.iter().map(|f| f.ident.as_ref().unwrap());
+
+        let output = quote! {
+            impl ByteCost for #ident {
+                fn bytes_weighted(&self, config: &bee_bytecost::ByteCostConfig) -> u64 {
```
pvdrz: @thibault-martinez would say that you either use `bee_bytecost::BytecostConfig` or `bee_byte_cost::ByteCostConfig`.

bee-message/bee-bytecost-derive/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use fields::AnnotatedFields;
+
+use proc_macro::TokenStream;
+use proc_macro_error::{abort, proc_macro_error};
+use quote::ToTokens;
+use syn::parse_macro_input;
+
+#[proc_macro_error]
+#[proc_macro_derive(ByteCost, attributes(data, key))]
```
pvdrz: I'd suggest using `attributes(bytecost)` and using `bytecost::key` and `bytecost::data` just for readability

bee-message/bee-bytecost-derive/src/fields.rs
```diff
@@ -0,0 +1,87 @@
...
+
+                    let mut sum_key_fields = 0;
+                    #(sum_key_fields += self.#key_fields.packed_len();)*
```
pvdrz: maybe is there a way to ensure that `#key_fields` implements `Packable` during compilation instead of relying in the: "could not find `packed_len` method" error. But I'd do that in a future PR: https://docs.rs/impls/1.0.3/impls/macro.impls.html

pvdrz (changes requested):

bee-message/bee-bytecost-derive/src/fields.rs
grtlr: I think we settled for `ByteCost`. But for some reason `bee_byte_cost_derive` looked weird. So I didn't but hyphens between the crate names.

thibault-martinez: We did settle on `ByteCost` but @pvdrz has a point, if we follow the logic the crate should be `bee_byte_cost`

bee-message/bee-bytecost-derive/src/lib.rs
grtlr: Yeah that's something that was still on my todo list. 😀 It will also help discern these attributes from other attributes such as `packable()`.

bee-message/bee-bytecost-derive/src/fields.rs
grtlr: For some reason I can't get this to work. 🙁

bee-message/bee-bytecost-derive/src/lib.rs
grtlr: I've settled for `#[byte_cost(key,data)]` for the attributes. What do you think?

bee-message/bee-bytecost-derive/src/fields.rs
pvdrz: weird :/ https://docs.rs/syn/1.0.82/syn/struct.Path.html#method.is_ident

bee-message/bee-bytecost-derive/src/lib.rs
pvdrz: I like it!

bee-message/bee-bytecost-derive/src/fields.rs
grtlr: But `Ident::new` needs a `Span`: https://docs.rs/syn/1.0.82/syn/struct.Ident.html#method.new

pvdrz: yeah but check the example, and also: https://docs.rs/syn/1.0.82/syn/struct.Ident.html#impl-PartialEq%3CT%3E

grtlr: Ok, I figured it out. Thanks!

bee-message/bee-bytecost-derive/src/fields.rs
grtlr: It's acutally not that complicated, but took me a while to figure out. We'll access these methods through their trait symbols that solves the problem (see also the [heapsize example](https://github.com/dtolnay/syn/tree/master/examples/heapsize) of the `syn` crate).

The only thing left now, is setting the correct spans.

bee-message/src/output/simple.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use bee_byte_cost::ByteCost;
+use bee_common::packable::{Packable, Read, Write};
+
+use core::ops::RangeInclusive;
+
+/// Describes a simple output that can only hold IOTAs.
+#[derive(ByteCost, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
+pub struct SimpleOutput {
```
Thoralf-M: There are no simple outputs anymore

bee-message/src/output/feature_block/timelock_milestone_index.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use bee_byte_cost::ByteCost;
+use bee_common::packable::{Packable, Read, Write};
+
+use derive_more::From;
+
+/// Defines a milestone index until which the output can not be unlocked.
+#[derive(ByteCost, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, From)]
+#[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
+pub struct TimelockMilestoneIndexFeatureBlock(
```
Thoralf-M: There is a `TimelockUnlockCondition`, but not a feature block
```Rust
pub struct TimelockUnlockCondition {
    // The milestone index starting from which the output can be consumed.
    milestone_index: MilestoneIndex,
    // Unix time, seconds since unix epoch, starting from which the output can be consumed.
    timestamp: u32,
}
```

bee-message/src/output/feature_block/indexation.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use crate::Error;
+
+use bee_byte_cost::ByteCost;
+use bee_common::packable::{Packable, Read, Write};
+
+/// Defines an indexation tag to which the output will be indexed.
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
+pub struct IndexationFeatureBlock(
```
Thoralf-M: Renamed to `TagFeatureBlock`

bee-message/src/output/feature_block/expiration_milestone_index.rs
```diff
@@ -0,0 +1,56 @@
...
+use bee_common::packable::{Packable, Read, Write};
+
+use derive_more::From;
+
+/// Defines a milestone index until which only the deposit [`Address`](crate::address::Address) is allowed to unlock the
+/// output. After the milestone index, only the [`Address`](crate::address::Address) defined in the
+/// [`SenderFeatureBlock`](crate::output::feature_block::SenderFeatureBlock) can unlock it.
+#[derive(ByteCost, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, From)]
+#[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
+pub struct ExpirationMilestoneIndexFeatureBlock(
```
Thoralf-M: Such a feature block doesn't exist

bee-message/src/output/feature_block/expiration_unix.rs
```diff
@@ -0,0 +1,56 @@
...
+use bee_common::packable::{Packable, Read, Write};
+
+use derive_more::From;
+
+/// Defines a unix time until which only the deposit [`Address`](crate::address::Address) is allowed to unlock the
+/// output. After the expiration time, only the [`Address`](crate::address::Address) defined in the
+/// [`SenderFeatureBlock`](crate::output::feature_block::SenderFeatureBlock) can unlock it.
+#[derive(ByteCost, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, From)]
+#[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
+pub struct ExpirationUnixFeatureBlock(
```
Thoralf-M: Such a feature block doesn't exist

bee-message/src/output/alias.rs
```diff
@@ -153,23 +155,31 @@ impl AliasOutputBuilder {
 pub(crate) type StateMetadataLength = BoundedU16<0, { AliasOutput::STATE_METADATA_LENGTH_MAX }>;
 
 /// Describes an alias account in the ledger that can be controlled by the state and governance controllers.
-#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(ByteCost, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
 #[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
 pub struct AliasOutput {
     // Amount of IOTA tokens held by the output.
+    #[byte_cost(data)]
```
Thoralf-M, line 162: Isn't there a good way to make this a default, so it doesn't need to be added to every field?

bee-message/src/output/extended.rs
```diff
@@ -101,18 +103,22 @@ impl ExtendedOutputBuilder {
 }
 
 /// Describes an extended output with optional features.
-#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Packable)]
+#[derive(ByteCost, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Packable)]
```
thibault-martinez: Bikeshedding but we usually put our own derives at the end of the list 😬

bee-message/bee-byte-cost/src/lib.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2020-2021 IOTA Stiftung
```
Alex6323: 2022

bee-message/bee-byte-cost/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "bee-byte-cost"
+version = "0.1.0"
+authors = [ "IOTA Stiftung" ]
+edition = "2021"
+description = "Defines the traits for computing the byte cost."
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "tangle", "bee", "framework", "bytecost", "dust" ]
```
Alex6323: wasn't that limited to at most 5 keywords? :thinking:

thibault-martinez: Yes we can't release the crate if it has more than 5 keywords

bee-message/bee-byte-cost/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "bee-byte-cost"
```
Alex6323, line 2: I am not sure at all, but I think since this is the new dust protection crate, and "byte cost" is just a specific impl. of a dust protection mechanism, we may better call it `bee-dust-protection` as this is also a bit more descriptive IMO.

bee-message/bee-byte-cost-derive/src/builder.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2021 IOTA Stiftung
```
Alex6323: 2022 ... also probably the other files ;)

bee-message/src/output/alias.rs
thibault-martinez: I agree with Thoralf here, looks very redundant. Maybe make it default and create an attribute to ignore a field ?

bee-message/bee-byte-cost/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+pub use bee_byte_cost_derive::ByteCost;
+pub use packable::Packable;
```
thibault-martinez, line 6:
```suggestion

pub use packable::Packable;
```

bee-message/src/output/alias.rs
grtlr: That's a good idea. It evolved that way because there used to be more granularity in earlier versions of the TIP.

bee-message/bee-byte-cost/Cargo.toml
grtlr: Hmm, I agree these two concepts are often used interchangeably. I don't have anything agains renaming it, other than having to do the work :trollface:. Any thoughts @thibault-martinez ?

bee-message/src/output/alias.rs
grtlr: I'm getting the feeling that this is a bit harder to achieve than I originally thought. In the current implementation we use the existence of an attribute to differentiate between [two things](https://github.com/iotaledger/bee/blob/853b5c7e2c9330bcfea670605d84edb01fa9358b/bee-message/bee-byte-cost-derive/src/field_info.rs#L52-L56):

* There is a `#[byte_cost()]` attribute: _We call `Packable::packed_len` on the field_, or
* There is no attribute: _We call `ByteCost::weighted_bytes` on a field that implements `ByteCost`_.

So if we use the second case to represent `#[byte_cost(data)]` as a default, we can't easily have custom, nested, implementations of `ByteCost` for types that are used in fields. Of course, we could reserve an additional attribute for that, but then you strongly couple these two types together.

Ideally, we could check if a field already implements the `ByteCost` trait, but I'm afraid that information is not available at the step where proc macros are run.

@thibault-martinez and @Thoralf-M: Do you think it's still worth doing, considering this tradeoff?

bee-message/bee-byte-cost-derive/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[dependencies]
+proc-macro-error = { version = "1.0.4", default-features = false, features = [ "syn-error" ] }
+proc-macro2 = { version = "1.0.30", default-features = false }
+quote = { version = "1.0.10", default-features = false }
+syn = { version = "1.0.80", default-features = false, features = [ "clone-impls", "derive", "parsing", "printing", "proc-macro" ] }
+
+[dev-dependencies]
+bee-byte-cost = { path = "../bee-byte-cost", version = "0.1.0", default-features = false }
+packable = { version = "0.1.0", default-features = false }
```
pvdrz, line 24: I think `packable` now goes on the second group

bee-message/bee-byte-cost-derive/src/builder.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2021-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use proc_macro2::TokenStream;
+use quote::{quote, ToTokens};
+use syn::Ident;
+
+pub(crate) struct ByteCostBuilder {
+    ident: Ident,
+    pub key: TokenStream,
```
pvdrz, line 10: is it important that these are `pub` and not `pub(crate)` given that `ByteCostBuilder` is `pub(crate)`?

bee-message/bee-byte-cost-derive/src/enum_info.rs
```diff
@@ -0,0 +1,69 @@
...
+            ident: variant.ident.clone(),
+        })
+    }
+}
+
+impl ToTokens for VariantInfo {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let ident = &self.ident;
+        let variant_fields = (0..self.fields.len()).map(|n| format_ident!("field_{}", n));
+        let variant_fields_2 = variant_fields.clone();
```
pvdrz, line 39: would this mean that `|n| format_ident!("field_{}", n)` is invoked twice? in that sense it might be faster to collect `variant_fields` and then `let variant_fields_2 = &variant_fields;` or something.

bee-message/bee-byte-cost-derive/src/enum_info.rs
```diff
@@ -0,0 +1,69 @@
...
+            .into_iter()
+            .map(|v| VariantInfo::new(&v))
+            .collect::<Result<Vec<VariantInfo>>>();
+        Ok(Self(variants?))
+    }
+
+    pub(crate) fn add_to(self, builder: &mut ByteCostBuilder) {
+        let match_arms = self.0.iter();
+
+        // TODO: Don't hard code the length of KIND;
```
pvdrz, line 61: TODO!

bee-message/bee-byte-cost-derive/src/field_info.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl ToTokens for FieldInfo {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let pattern_ident = &self.pattern_ident;
+
+        let token_stream = if self.tags.is_empty() {
+            quote_spanned! {pattern_ident.span()=> + bee_byte_cost::ByteCost::weighted_bytes(&self.#pattern_ident, config) }
+        } else {
+            quote_spanned! {pattern_ident.span()=> + packable::PackableExt::packed_len(&self.#pattern_ident) as u64 }
```
pvdrz, line 55: I wonder if `bee_byte_cost` should reexport `packable` or not to avoid weird `crate packable not found` errors

bee-message/src/output/alias.rs
pvdrz: what if you could add an attribute to the struct such that:

```rust
#[byte_cost(data)]
struct Foo {
    bar: Bar,
    baz: Baz,
}
```

were the same as

```rust
struct Foo {
    #[byte_cost(data)]
    bar: Bar,
    #[byte_cost(data)]
    baz: Baz,
}
```

bee-message/tests/byte_cost.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use bee_byte_cost::{ByteCost, ByteCostConfig};
+
+const CONFIG: ByteCostConfig = ByteCostConfig {
+    byte_cost: 1,
+    weight_for_data: 10,
+    weight_for_key: 1,
+};
+
+const OFFSET: u64 = 34 /* OutputID */
```
pvdrz, line 15: can't we use `core::mem::size_of` instead?

bee-message/bee-byte-cost-derive/Cargo.toml
thibault-martinez: Indeed

bee-message/bee-byte-cost/Cargo.toml
thibault-martinez: It is true that the dust protection is currently based on the byte cost, but changing it would be breaking change anyway so it's not like we would benefit from a "dynamic" name. Also, AFAIK, the byte cost crate in itself only really computes the byte cost, which is on its own *not* a dust protection. Dust protection is the syntactical comparison of the output amount against the byte cost and that can only happen within the bee-message crate.

bee-message/src/output/alias.rs
grtlr: Interesting! So the first case would be an attribute macro (without the derive), while the second case would be a derive macro? Can they coexist in that way?

Or could the second case even be the same attribute macro as the first case?

Update: I think I got it mixed up, both struct and field attributes should be part of the derive macro.

bee-message/bee-byte-cost/Cargo.toml
Alex6323: makes sense

grtlr:
Closing in favor of #1095.

PR #884, opened 2021-12-15
Minimum working changes to bee-ledger

thibault-martinez:
This PR reintroduces the bee-ledger crate that was previously disabled.
It also does the minimum changes that are required to make it compile.
It will be completed later but this is needed to unblock other components that are being worked on.

## Change checklist

### DIFF
--- a/bee-ledger/src/types/balance_diff.rs
+++ b/bee-ledger/src/types/balance_diff.rs
@@ -31,6 +31,7 @@ impl BalanceDiff {
 
     /// Negates a [`BalanceDiff`].
     pub fn negate(&mut self) {
+        // TODO this can overflow
         self.0 = -self.0;
     }
 }
--- a/bee-ledger/src/types/balance_diff.rs
+++ b/bee-ledger/src/types/balance_diff.rs
@@ -3,11 +3,7 @@
 
 use crate::types::Error;
 
-use bee_message::{
-    address::Address,
-    constants::IOTA_SUPPLY,
-    output::{Output, DUST_THRESHOLD},
-};
+use bee_message::{address::Address, constant::IOTA_SUPPLY, output::Output};
 
 use std::collections::{
     hash_map::{IntoIter, Iter, IterMut},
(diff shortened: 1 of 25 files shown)


bee-ledger/src/types/balance_diff.rs
```diff
@@ -16,55 +12,26 @@ use std::collections::{
...
-
-    /// Returns whether dust allowance has been decreased or dust outputs has been increased.
-    pub fn is_dust_mutating(&self) -> bool {
-        self.dust_allowance < 0 || self.dust_outputs > 0
+        self.0
     }
 
-    /// Negates a `BalanceDiff`.
+    /// Negates a [`BalanceDiff`].
     pub fn negate(&mut self) {
```
grtlr, line 33: At a later point, we could consider implementing basic math traits (from `std::ops`) on `Balance` and `BalanceDiff`, now that they are simpler again.

The problem then is of course handling overflows.

Speaking of overflows, `negate` could potentially overflow ([live demo](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8157be5866320cc55fed31e392da7ca4)), but this will probably never happen because the smallest/largest value is defined by `IOTA_SUPPLY`, right? If so, we should consider leaving a quick note here.

bee-ledger/src/workers/error.rs
```diff
@@ -55,9 +55,6 @@ pub enum Error {
     /// Invalid ledger balance state.
     #[error("Invalid ledger balance state: {0}")]
     InvalidLedgerBalanceState(u64),
-    /// Invalid ledger dust state.
-    #[error("Invalid ledger dust state: {0:?} {1:?}")]
-    InvalidLedgerDustState(Address, Balance),
     /// Consumed amount overflow.
     #[error("Consumed amount overflow: {0}.")]
     ConsumedAmountOverflow(u128),
```
grtlr: Change capitalization maybe?

bee-ledger/src/types/balance_diff.rs
thibault-martinez: Yeah the crate will be refactored anyway and these need some due improvement. I will add a TODO for negate, good catch 👍

thibault-martinez: Yeah it's technically bound to `IOTA_SUPPLY` but I don't know yet how this can change with shimmer and all, will leave a todo for now and convert to a simple comment later when I thought about it

bee-test/src/rand/metadata.rs
```diff
@@ -13,7 +13,7 @@ use bee_tangle::{
 
 /// Generates a random conflict reason.
 pub fn rand_conflict_reason() -> ConflictReason {
-    ((rand_number::<u64>() % 7) as u8).try_into().unwrap()
```
grtlr, line 16: Are you intentionally leaving out `ConflictReason::InvalidDustAllowance` and `ConflictReason::SemanticValidationFailed`? If so, we should maybe document that.

Ideally we would have something like this though:

```rust
let mut rng = rand::thread_rng();
let conflicts = [
    ConflictReason::None,
    ConflictReason::InputUtxoAlreadySpent,
    ...
];
let conflict = conflicts.choose(&mut rng).unwrap();
```

With this we could roundtrip: From _random conflict_ to `u8` and then `try_into`.

thibault-martinez: `ConflictReason::InvalidDustAllowance` has been removed and yeah we're leaving out `ConflictReason::SemanticValidationFailed` intentionally.  I will leave a comment for now, this PR is really just the bare minimum to make every comment, the `bee-test` will be drastically refactored in an incoming PR.

grtlr (approved):

PR #895, opened 2021-12-27
DTOs and conversions for all new output types

samuel-rufi:


bee-api/bee-rest-api/src/types/dtos.rs
```diff
@@ -554,6 +707,476 @@ pub struct ReferenceUnlockBlockDto {
...
+            Self::TimelockUnix(_) => TimelockUnixFeatureBlock::KIND,
+            Self::ExpirationMilestoneIndex(_) => ExpirationMilestoneIndexFeatureBlock::KIND,
+            Self::ExpirationUnix(_) => ExpirationUnixFeatureBlock::KIND,
+            Self::Indexation(_) => IndexationFeatureBlock::KIND,
+            Self::Metadata(_) => MetadataFeatureBlock::KIND,
+        }
+    }
+}
+
+impl From<&FeatureBlock> for FeatureBlockDto {
```
pvdrz, line 887: I think `From<T>` is more flexible than `From<&T>`:

- If we implement `From<T>` those two are fine:

```rust
let t: T = ...;
let v = V::from(t); // We don't clone `t` because we won't use it anymore
```

```rust
let t: T = ...;
let v = V::from(t.clone()); // We clone `t` because we want to use it later
t.do_stuff();
```

- If we implement `From<&T>` then:

```rust
let t: T = ...;
let v = V::from(&t); // This will clone/copy stuff from `t` even though `t` won't be used anymore.
```

```rust
let t: T = ...;
let v = V::from(&t);
t.do_stuff();
```

bee-api/bee-rest-api/src/types/dtos.rs
```diff
@@ -554,6 +707,476 @@ pub struct ReferenceUnlockBlockDto {
...
+            FeatureBlockDto::TimelockUnix(v) => Self::TimelockUnix(TimelockUnixFeatureBlock::new(v.0)),
+            FeatureBlockDto::ExpirationMilestoneIndex(v) => {
+                Self::ExpirationMilestoneIndex(ExpirationMilestoneIndexFeatureBlock::new(v.0))
+            }
+            FeatureBlockDto::ExpirationUnix(v) => Self::ExpirationUnix(ExpirationUnixFeatureBlock::new(v.0)),
+            FeatureBlockDto::Indexation(v) => Self::Indexation(IndexationFeatureBlock::new(
+                &hex::decode(&v.0).map_err(|_e| Error::InvalidSemanticField("IndexationFeatureBlock"))?,
+            )?),
+            FeatureBlockDto::Metadata(v) => Self::Metadata(MetadataFeatureBlock::new(
+                &hex::decode(&v.0).map_err(|_e| Error::InvalidSemanticField("MetadataFeatureBlock"))?,
```
pvdrz, line 927: like here, this decoded hex buffer is passed by reference but nobody else uses it. I assume `MetadataFeatureBlock::new` will copy some/all of the bytes instead of moving them.

bee-api/bee-rest-api/src/types/dtos.rs
```diff
@@ -554,6 +707,476 @@ pub struct ReferenceUnlockBlockDto {
...
+    //
+    pub feature_blocks: Vec<FeatureBlockDto>,
+}
+
+impl From<&AliasOutput> for AliasOutputDto {
+    fn from(value: &AliasOutput) -> Self {
+        Self {
+            kind: AliasOutput::KIND,
+            amount: value.amount(),
+            native_tokens: value.native_tokens().iter().map(|x| x.into()).collect::<_>(),
```
pvdrz: maybe `Into::into` is better?

pvdrz (changes requested):

bee-api/bee-rest-api/src/types/dtos.rs
samuel-rufi: Do you generally think we should do `From<T>` instead of  `From<&T>` or just for the `FeatureBlock` here?

What is your take on this @thibault-martinez ?

bee-api/bee-rest-api/src/types/dtos.rs
samuel-rufi: 👍

samuel-rufi: Changed all try_into/into mappings

bee-api/bee-rest-api/src/types/dtos.rs
pvdrz: I'd do everywhere. If we need `From<&T>` we should be implementing `AsRef<T>`, `Borrow<T>` or `Deref<T>`.

samuel-rufi:
Will be replaced by a following PR.

Issue #883, opened 2021-12-14
Stardust layouts improvements


thibault-martinez:
- [ ] Documentations
- [ ] Tests
- [ ] Copy or not Copy ? https://github.com/iotaledger/bee/pull/875#discussion_r766967641
- [ ] `AliasAddress` as its own type or not ? https://github.com/iotaledger/bee/pull/875#discussion_r766968615
- [ ] `const-panic` for unicity of kinds. https://github.com/iotaledger/bee/pull/875#discussion_r766970370
- [ ] `MessageBuilder`'s `Default` derivation ? https://github.com/iotaledger/bee/pull/875#discussion_r766972668
- [ ] Use `builder` crate ? https://github.com/iotaledger/bee/pull/875#discussion_r766989431
- [ ] `Output` trait ? https://github.com/iotaledger/bee/pull/875#discussion_r767106257
- [ ] Review all ranges integer types, all `usize` ? https://github.com/iotaledger/bee/pull/875#discussion_r767190886
- [ ] `It would be nice to short circuit on the first duplicate.` https://github.com/iotaledger/bee/pull/875#discussion_r767191349
- [ ] Extract derivated traits from `impl_id`

PR #902, opened 2021-12-29
Random generation of `Output`s

grtlr:
# Description of change

Implements random generation of `Output`s for testing purposes. 

* This PR is some of the cleanup of #883

## Type of change

## Change checklist


bee-test/src/rand/output/feature_block.rs
```diff
@@ -0,0 +1,86 @@
+// Copyright 2020-2021 IOTA Stiftung
```
thibault-martinez:
```suggestion
// Copyright 2021 IOTA Stiftung
```

bee-test/src/rand/output/feature_block.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Generates a random [`TimelockMilestoneIndexFeatureBlock`].
+pub fn rand_timelock_milestone_index() -> TimelockMilestoneIndexFeatureBlock {
+    TimelockMilestoneIndexFeatureBlock::new(rand_milestone_index())
+}
+
+/// Generates a random [`TimelockUnixFeatureBlock`].
+pub fn rand_timelock_unix() -> TimelockUnixFeatureBlock {
+    TimelockUnixFeatureBlock::new(rand_number())
+}
```
thibault-martinez, line 30: Pure OCD now but can you order the functions in their actual feature block order please 😬

bee-test/src/rand/output/feature_block.rs
```diff
@@ -0,0 +1,86 @@
...
+    vec![
+        FeatureBlock::Sender(rand_sender()),
+        FeatureBlock::Issuer(rand_issuer()),
+        FeatureBlock::DustDepositReturn(rand_dust_deposit_return()),
+        FeatureBlock::TimelockMilestoneIndex(rand_timelock_milestone_index()),
+        FeatureBlock::TimelockUnix(rand_timelock_unix()),
+        FeatureBlock::ExpirationMilestoneIndex(rand_expiration_milestone_index()),
+        FeatureBlock::ExpirationUnix(rand_expiration_unix()),
+        FeatureBlock::Indexation(rand_indexation()),
+        FeatureBlock::Metadata(rand_metadata()),
```
thibault-martinez: Reverse metadata and indexation please

bee-test/src/rand/output/feature_block.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/// Generates a random [`MetadataFeatureBlock`].
+pub fn rand_metadata() -> MetadataFeatureBlock {
+    let bytes = rand_bytes(rand_number_range(1..MetadataFeatureBlock::LENGTH_MAX));
+    MetadataFeatureBlock::new(&bytes).unwrap()
+}
+
+/// Generates a random [`IndexationFeatureBlock`].
+pub fn rand_indexation() -> IndexationFeatureBlock {
```
thibault-martinez: I know it would make these function names pretty long but I would add the `_feature_block` prefix to all these
- for consistency with all other `rand_` functions in this crate
- to reduce confusion about what it's actually generating, `rand_indexation` could be an indexation payload

grtlr: I thought about this as well. Maybe one solution would be to import these functions in a qualified way like `feature_block::rand_indexation`. But ofc, that would require discipline when using these functions.

I'm fine with either way.

thibault-martinez: Yeah I usually try to avoid relying on qualified names. I personally don't think long names are bad/annoying, if that's what it takes for them to be explicit and less error/confusion-prone I'm all in

bee-test/src/rand/output/feature_block.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/// Generates a random [`IssuerFeatureBlock`].
+pub fn rand_issuer_feature_block() -> IssuerFeatureBlock {
+    IssuerFeatureBlock::new(rand_address())
+}
+
+/// Generates a random [`DustDepositReturnFeatureBlock`].
+pub fn rand_dust_deposit_return_feature_block() -> DustDepositReturnFeatureBlock {
+    DustDepositReturnFeatureBlock::new(rand_number_range(DUST_DEPOSIT_MIN..IOTA_SUPPLY)).unwrap()
```
thibault-martinez: Can you use the 3 new ranges here and below ?

bee-test/src/rand/output/feature_block.rs
```diff
@@ -0,0 +1,86 @@
...
+        FeatureBlock::ExpirationUnix(rand_expiration_unix_feature_block()),
+        FeatureBlock::Metadata(rand_metadata_feature_block()),
+        FeatureBlock::Indexation(rand_indexation_feature_block()),
+    ]
+}
+
+/// Generates a [`Vec`] of random [`FeatureBlock`]s given a set of allowed [`FeatureBlockFlags`].
+pub fn rand_allowed_feature_blocks(allowed_feature_blocks: FeatureBlockFlags) -> Vec<FeatureBlock> {
+    let mut all_feature_blocks = all_feature_blocks();
+    all_feature_blocks.retain(|feature_block| allowed_feature_blocks.contains(feature_block.flag()));
```
thibault-martinez, line 48: I'm just thinking out loud but isn't there a way to do the opposite ? From the allowed flags generate the feature blocks ? That would mean matching I guess. I think the current way is a tiny bit error prone because we would have to remember to modify `all_feature_blocks` whenever we have new blocks, the other way we wouldn't have to.

grtlr: That's a very good point which is closely intertwined with our discussion on the `KIND` associated constant. It would be great if things like uniqueness and ordering of `KIND`s were checked at compile time. One way could be to have a central register (maybe in the form of a macro) that hands out the `KIND`s (as trait implementations).

But until we have such a system, I think you're right we should turn it around.

The problem is though, that I'm not aware if there is a way to iterate over all possible values in a bitflag (apart from trying all digits in the integer that represents the bitflag).

grtlr: The problem is though, that I'm not aware if there is a way to iterate over all possible values in a bitflag (apart from trying all digits in the integer that represents the bitflag).

pvdrz: I think we are basically trying to abstract over what an `enum` "is". So if we have
```rust
struct Bar;

struct Baz;

enum Foo {
    Bar(Bar),
    Baz(Baz),
}
```
we want to somehow assign `Bar` and `Baz` a `KIND` so we can distinguish them. But at the same time `Foo` should guarantee that both `KIND`s are different. This is basically what a [discriminant](https://doc.rust-lang.org/core/mem/struct.Discriminant.html) is but we want to control its value too.

An option would be to introduce two traits
```rust
trait Kinded {
    type Kind;
    // This could be a constant.
    fn kind(&self) -> Self::Kind;
}

trait KindOf<K: Kinded> {
     const KIND: K::Kinded;
}
```
`KindOf` would be implemented for `Bar` and `Baz` and `Kinded` could be derived for `Foo` to produce the following code:

```rust
impl Kinded for Foo {
    type Kind = u8;
    
    fn kind(&self) -> Self::Kind {
        const _: () = {
            if Bar::KIND == Baz::KIND {
                panic!("here be dragons");
            }
        };
        match self {
            Self::Bar(_) => Bar::KIND,
            Self::Baz(_) => Baz::KIND,
        }
    }
}
```

Edit: We could be a bit more ambitious and just do a single macro for everything:

```rust
#[derive(Kinded)]
#[kinded(kind_type  = u8)]
enum Foo {
    #[kinded(kind = 1)]
    Bar(Bar),
    #[kinded(kind = 2)]
    Baz(Baz),
}
```

And we could add some extra functionality (more!) to `bee-packable` so we can derive `Packable` for enums that implement `Kinded` and all its variants implement `Packable + KindOf`.

thibault-martinez: I am not very fond of these traits tbh. It makes everything more complex. If the only thing we are trying to achieve with them is to be sure there are no duplicates I wouldn't worry too much about it. I really don't see how we can mess up that much. I was more interested in you having a look at the second part of the comment :p

bee-message/src/output/feature_block/mod.rs
```diff
@@ -199,22 +199,57 @@ pub(crate) fn validate_allowed_feature_blocks(
...
-        const DUST_DEPOSIT_RETURN = 1 << DustDepositReturnFeatureBlock::KIND;
-        /// Signals the presence of a [`TimelockFeatureBlock`].
-        const TIMELOCK = 1 << TimelockFeatureBlock::KIND;
-        /// Signals the presence of a [`ExpirationFeatureBlock`].
-        const EXPIRATION = 1 << ExpirationFeatureBlock::KIND;
-        /// Signals the presence of a [`MetadataFeatureBlock`].
-        const METADATA = 1 << MetadataFeatureBlock::KIND;
-        /// Signals the presence of a [`IndexationFeatureBlock`].
-        const INDEXATION = 1 << IndexationFeatureBlock::KIND;
+macro_rules! create_bitflags {
```
thibault-martinez: Can you put it in the `macro` module ? I am soon splitting the feature blocks in two and I'd need to reuse this in another module.

grtlr: I'll do that. I just first have to figure out how to get the comments to render correctly. Also, if we want to reuse this code, we need to come up with general rustdoc comments.

bee-message/src/output/extended.rs
```diff
@@ -122,8 +122,9 @@ impl ExtendedOutput {
         .union(UnlockConditionFlags::DUST_DEPOSIT_RETURN)
         .union(UnlockConditionFlags::TIMELOCK)
         .union(UnlockConditionFlags::EXPIRATION);
+
```
thibault-martinez:
```suggestion
```

bee-message/src/macro.rs
```diff
@@ -94,3 +94,29 @@ macro_rules! string_serde_impl {
...
+
+/// A convenience macro to work around the fact the `[bitflags]` crate does not yet support iterating over the
+/// individual flags. This macro essentially creates the `[bitflags]` and puts the individual flags into a associated
+/// constant `pub const ALL_FLAGS: &'static []`.
+#[macro_export]
+macro_rules! create_bitflags {
+    ($(#[$meta:meta])* $Name:ident, $TagType:ty, [$(($FlagName:ident, $TypeName:ident),)+]) => {
+        bitflags! {
+            $(#[$meta])*
+            pub struct $Name: $TagType {
```
thibault-martinez, line 106: This needs documentation

bee-message/src/macro.rs
```diff
@@ -94,3 +94,29 @@ macro_rules! string_serde_impl {
         }
     };
 }
+
+/// A convenience macro to work around the fact the `[bitflags]` crate does not yet support iterating over the
+/// individual flags. This macro essentially creates the `[bitflags]` and puts the individual flags into a associated
```
thibault-martinez:
```suggestion
/// individual flags. This macro essentially creates the `[bitflags]` and puts the individual flags into an associated
```

bee-message/src/macro.rs
```diff
@@ -94,3 +94,29 @@ macro_rules! string_serde_impl {
...
+#[macro_export]
+macro_rules! create_bitflags {
+    ($(#[$meta:meta])* $Name:ident, $TagType:ty, [$(($FlagName:ident, $TypeName:ident),)+]) => {
+        bitflags! {
+            $(#[$meta])*
+            pub struct $Name: $TagType {
+                $(
+                    #[doc = "Signals the presence of a [`"]
+                    #[doc = stringify!($TypeName)]
+                    #[doc = "`]."]
```
thibault-martinez: Look at `impl_id` you can concat the docs

grtlr: Ah, nice!

bee-message/src/output/feature_block/mod.rs
```diff
@@ -158,16 +158,32 @@ pub(crate) fn validate_allowed_feature_blocks(
...
+        (TAG, TagFeatureBlock),
+    ]
+);
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn check_length() {
```
thibault-martinez: I'm picky but that's not exactly what this test is doing 😬

thibault-martinez (approved):

PR #912, opened 2021-12-30
Add `verify_with` to `#[derive(Packable)]`

pvdrz:


bee-common/bee-packable-derive/src/field_info.rs
```diff
@@ -40,20 +41,35 @@ impl FieldInfo {
...
+            })? {
+                unpack_error_with_opt = Some(unpack_error_with);
             }
         }
+
         Ok(Self {
-            unpack_error_with: default_unpack_error_with.clone(),
+            unpack_error_with: unpack_error_with_opt.unwrap_or_else(|| default_unpack_error_with.clone()),
+            verify_with: verify_with_opt
+                .unwrap_or_else(|| parse2(quote!(|_| -> Result<(), Self::UnpackError> { Ok(()) })).unwrap()),
```
grtlr: Is there a reason against using `.unwrap_or()` to avoid the closure?

grtlr (dismissed):

bee-common/bee-packable-derive/tests/pass/verify_with.rs
```diff
@@ -0,0 +1,40 @@
...
+        Ok(())
+    } else {
+        Err(PickyError(value))
+    }
+}
+
+#[derive(Packable)]
+#[packable(unpack_error = PickyError)]
+pub struct Picky(
+    #[packable(verify_with = verify_value)]
```
thibault-martinez, line 36: What happens if the given function is not compatible with the field type ? Can we have a fail test for that ?

thibault-martinez, line 36: Can we have multiple `verify_with` or should we always combine them all in one ? Just curious, not a request.

bee-common/bee-packable-derive/src/field_info.rs
pvdrz: `parse2` and `quote!` do a non trivial amount of work. I could add a new parameter to this function so it receives a `default_verify_with` so we only compute it once given that it would be the same everywhere.

bee-common/bee-packable-derive/tests/pass/verify_with.rs
pvdrz: only one is used but I'm not sure which one.

pvdrz: sure

bee-common/bee-packable-derive/src/field_info.rs
grtlr: Oh, I forgot about the lazy aspect of `unwrap_or_else`. Considering that I think it's fine as it is!

thibault-martinez (dismissed):

bee-common/bee-packable-derive/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright 2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-//! This crate provides the [`Packable`] derive macro.
```
grtlr, line 4: Any particular reason for getting rid of the doc linking?

grtlr (approved):

bee-message/src/output/signature_locked_single.rs
```diff
@@ -3,27 +3,27 @@
...
-};
+use bee_packable::{bounded::BoundedU64, Packable};
 
 use core::ops::RangeInclusive;
 
 /// Valid amounts for a signature locked single output.
 pub const SIGNATURE_LOCKED_SINGLE_OUTPUT_AMOUNT: RangeInclusive<u64> = 1..=IOTA_SUPPLY;
 
+pub(crate) type Amount =
+    BoundedU64<{ *SIGNATURE_LOCKED_SINGLE_OUTPUT_AMOUNT.start() }, { *SIGNATURE_LOCKED_SINGLE_OUTPUT_AMOUNT.end() }>;
```
Adam-Gleave, line 14: Is there some way we can simplify this a little? 😬

bee-common/bee-packable-derive/src/lib.rs
thibault-martinez: Same question

thibault-martinez (approved):

bee-common/bee-packable-derive/src/lib.rs
pvdrz: yes, the link is pointing to the macro itself so it is misleading, technically `bee-packable` is not a dependency of `bee-packable-derive` so the `Packable` trait is not available.

bee-message/src/output/signature_locked_single.rs
pvdrz: we could introduce `SIGNATURE_LOCKED_SINGLE_OUTPUT_(MIN|MAX)_AMOUNT` but I'm not sure if that's simpler.

thibault-martinez: Yeah I don't think it would be :/

samuel-rufi (approved):

PR #944, opened 2022-01-12
Add `[profile]` that optimizes for production

### LINKED ISSUE #925: Investigate Rust binary size reduction
By default, rust binaries are pretty big but can be reduced with some options.
https://www.reddit.com/r/rust/comments/rxnbu7/why_do_rust_executables_produce_such_huge_files/
https://www.reddit.com/r/rust/comments/rw0jad/cargos_strip_profile_option_has_been_stabilized/

grtlr:
# Description of change

Creates a profile that optimizes for binary size. 

Benchmarks and discussion in #925.

## Type of change

Benchmarks in #925.

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,6 @@
+# We can get rid of this once Rust 1.58 lands
+cargo-features = ["strip"]
+
 [workspace]
 members = [
 	"bee-api/bee-rest-api",
@@ -10,3 +13,11 @@ panic = "abort"
 
 [profile.release]
 panic = "abort"
+
+[profile.min-size]
+inherits = "release"
+opt-level = "s"
+lto = true
+codegen-units = 1
+panic = "abort"
+strip = "symbols"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,9 @@ panic = "abort"
 panic = "abort"
 
 [profile.min-size]
-inherits = "release"
-opt-level = "s"
-lto = true
 codegen-units = 1
+inherits = "release"
+lto = true
+opt-level = "s"
 panic = "abort"
 strip = "symbols"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -10,3 +13,11 @@ panic = "abort"
 
 [profile.release]
 panic = "abort"
+
+[profile.min-size]
+codegen-units = 1
+inherits = "release"
+lto = true
+opt-level = "s"
+panic = "abort"
```
thibault-martinez, line 21: Is this needed if we inherit from `release` ? Assuming my understanding of `inherits` is correct

grtlr: I don't think it's necessary, but if we ever change `panic` in `release` we have to keep in mind to change it in the new profile too. Since it directly influences the size (https://github.com/johnthagen/min-sized-rust#abort-on-panic) I thought it was good to keep it here just in case. But I can remove it if you want.

pvdrz:
what happened with the `opt-level = "z"` flag?

grtlr:
`opt-level = "z"` would turn off loop vectorization, which I wouldn't do in a production build as this kind of code duplication will probably be negligible compared to the rest of Bee.

From the documentation (https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level):
> "z": optimize for binary size, but also turn off loop vectorization.

thibault-martinez (approved):

pvdrz:
what about `opt-level = "s"` then? :trollface:

grtlr:
I don't think we can properly benchmark the difference that it makes to the node software. So in that case, I would rather err on the side of performance.

pvdrz (approved):

thibault-martinez (approved):

PR #984, opened 2022-01-18
Add JWT support for API and Dashboard

samuel-rufi:
Generate a JWT token with: `cargo release -- jwt-api`

Example request for protected `/api/v1/tips`: 
```
curl -X 'GET' \
  'http://127.0.0.1:14265/api/v1/tips' \
  -H 'accept: application/json' \
  -H 'authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiIxMkQzS29vV1BydFg1cE1hYlZuaUJ5ZVo4OXh6OWVnWnhQcTRtNnNHNGV0V1Fwb2pjMkNmIiwic3ViIjoiQmVlIiwiYXVkIjoiYXBpIiwibmJmIjoxNjQ3Mjc2NDUxLCJpYXQiOjE2NDcyNzY0NTF9.WMprcjWTEVrszhfeoCEfg3-0nRRFrnRlUUtVeD78Xqs'
  ```

You can protect routes with the node config file. Simply add the route you want to protect to the `protectedRoutes` array. To make routes public use the `publicRoutes` array.

### CODE REVIEW

File: bee-api/bee-rest-api/src/endpoints/filters.rs [hunk 1]
```diff
@@ -1,101 +1,16 @@
...
-pub(crate) fn with_consensus_worker(
-    consensus_worker: mpsc::UnboundedSender<ConsensusWorkerCommand>,
-) -> impl Filter<Extract = (mpsc::UnboundedSender<ConsensusWorkerCommand>,), Error = Infallible> + Clone {
-    warp::any().map(move || consensus_worker.clone())
-}
+pub(crate) fn with_args<B: StorageBackend>(
+    args: Arc<ApiArgs<B>>
+) -> impl Filter<Extract = (Arc<ApiArgs<B>>,), Error = Infallible> + Clone {
+    warp::any().map(move || args.clone())
+}
```
grtlr: Add newline at the end.

```suggestion
}

```

bee-api/bee-rest-api/Cargo.toml
```diff
@@ -30,6 +31,7 @@ iota-crypto = { version = "0.9.1", default-features = false, features = [ "blake
 log = { version = "0.4.14", default-features = false, optional = true }
 multiaddr = { version = "0.13.0", default-features = false }
 num_cpus = { version = "1.13.0", default-features = false, optional = true }
+regex = "1.5.4"
```
grtlr: We should probably use the same syntax as the other entries.

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -147,24 +125,48 @@ impl RestApiConfigBuilder {
...
+        let public_routes = {
+            let mut ret = Vec::new();
+            for route in self
+                .public_routes
+                .unwrap_or_else(|| DEFAULT_PUBLIC_ROUTES.iter().map(|s| s.to_string()).collect())
+            {
+                let mut escaped: String = regex::escape(&route);
+                escaped = escaped.replace("\\*", ".*");
+                escaped.push('$');
+                ret.push(Regex::new(&escaped).expect(&format!("invalid public route {}", route)))
```
grtlr: Factoring this out into a separate method could help document what this regex is doing.

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -174,10 +176,17 @@ impl RestApiConfigBuilder {
...
 #[derive(Clone)]
 pub struct RestApiConfig {
+    /// REST API binding address.
     pub(crate) binding_socket_addr: SocketAddr,
-    pub(crate) public_routes: Box<[String]>,
-    pub(crate) allowed_ips: Box<[IpAddr]>,
+    /// JWT salt for REST API.
+    pub(crate) jwt_salt: String,
+    /// Routes that are available for public use and don't need JWT authentication.
+    pub(crate) public_routes: Box<[Regex]>,
```
grtlr: Just out of curiosity: What's the motivation for using `Box<[Regex]>` over `Vec<Regex>`?

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -16,30 +16,35 @@ use storage::StorageBackend;
...
+use std::{any::TypeId, convert::Infallible, sync::Arc};
 
 pub(crate) type NetworkId = (String, u64);
 pub(crate) type Bech32Hrp = String;
 
 pub(crate) const CONFIRMED_THRESHOLD: u32 = 5;
 
 pub async fn init<N: Node>(
+    node_id: PeerId,
+    keypair: Keypair,
```
grtlr: Nitpicking: Although the chance for an error is probably small, maybe we could only pass `keypair` to the constructor and recreate the `PeerId` internally.

bee-api/bee-rest-api/src/endpoints/permission.rs
```diff
@@ -1,30 +1,65 @@
...
+};
 
-use std::net::{IpAddr, SocketAddr};
+use std::sync::Arc;
 
-pub fn has_permission(
-    route: &'static str,
-    public_routes: Box<[String]>,
-    allowed_ips: Box<[IpAddr]>,
+const BEARER: &str = "Bearer ";
```
grtlr, line 17: The whitespace is probably important here? If so, I would document that.

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -2,79 +2,51 @@
...
+    "/api/v1/addresses*",
+    "/api/v1/treasury",
+    "/api/v1/receipts*",
 ];
+/// Default routes that are protected and need JWT authentication.
+pub(crate) const DEFAULT_PROTECTED_ROUTES: [&str; 2] = ["/api/v1/*", "/api/plugins/*"];
+/// Enables the proof-of-work feature on the node per default.
 pub(crate) const DEFAULT_FEATURE_PROOF_OF_WORK: bool = true;
+/// Default value for the white flag solidification timeout.
 pub(crate) const DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT: u64 = 2;
```
grtlr, line 68: Consider making this a `Duration` and use `from_secs`.

bee-api/bee-rest-api/src/endpoints/routes/api/v1/message_children.rs
```diff
@@ -25,24 +20,19 @@ fn path() -> impl Filter<Extract = (MessageId,), Error = warp::Rejection> + Clon
...
 
 pub async fn message_children<B: StorageBackend>(
     message_id: MessageId,
-    tangle: ResourceHandle<Tangle<B>>,
+    args: Arc<ApiArgs<B>>,
 ) -> Result<impl Reply, Rejection> {
-    let mut children = Vec::from_iter(tangle.get_children(&message_id).await.unwrap_or_default());
+    let mut children = Vec::from_iter(args.tangle.get_children(&message_id).await.unwrap_or_default());
     let count = children.len();
     let max_results = 1000;
```
grtlr, line 46: I know this is not part of your PR, but should this maybe be a `const`?

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -71,47 +83,32 @@ where
...
+            bus: node.bus(),
+            node_info: node.info(),
+            tangle: node.resource::<Tangle<N::Backend>>(),
+            peer_manager: node.resource::<PeerManager>(),
+            requested_messages: node.resource::<RequestedMessages>(),
+            network_command_sender: node.resource::<NetworkCommandSender>(),
+            message_submitter: node.worker::<MessageSubmitterWorker>().unwrap().tx.clone(),
+            message_requester: node.worker::<MessageRequesterWorker>().unwrap().clone(),
+            consensus_worker: node.worker::<ConsensusWorker>().unwrap().tx.clone(),
+        });
```
grtlr: Do you think we might loose performance by putting all of these into a single `Arc`?

bee-api/bee-rest-api/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2020-2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+extern crate regex;
+extern crate warp;
+
```
grtlr: Could this be omitted?

bee-node/src/lib.rs
```diff
@@ -5,6 +5,10 @@
 //! * A Bee full node;
 //! * A Bee entry node (autopeering);
 
+extern crate rpassword;
+extern crate auth_helper;
+extern crate bee_rest_api;
+
```
grtlr: Same here

bee-node/src/tools/jwt_api.rs
```diff
@@ -0,0 +1,37 @@
...
+pub fn exec<B: NodeStorageBackend>(
+    _tool: &JwtApiTool,
+    local: &Local,
+    node_config: &NodeConfig<B>,
+) -> Result<(), JwtApiError> {
+    let jwt = JsonWebToken::new(
+        local.peer_id().to_string(),
+        node_config.rest_api_config.jwt_salt().to_owned(),
+        AUDIENCE_CLAIM.to_owned(),
+        10000, // JWT tokens shouldn't expire.
```
grtlr: Maybe move this to a `const`.

grtlr (changes requested):
It's cool that this adds functionality, but also makes parts of the code easier to read/maintain. :thumbsup:

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: @pvdrz suggested to do this to avoid some cloning if I remember correctly. With this PR we use `RegexSet` instead of `Box<[Regex]>` - just for info.

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: Done!

bee-node/src/tools/jwt_api.rs
samuel-rufi: API JWT tokens don't expire anymore, so this isn't an issue anymore

bee-api/bee-rest-api/src/endpoints/filters.rs
samuel-rufi: This has been fixed

bee-api/bee-rest-api/src/endpoints/mod.rs
samuel-rufi: Seems like putting "API arguments" into an Arc and passing them to the warp endpoits is the general recommandation.
Instead of cloning only the specific arguments we now clone the Arc in addition.
This 1 reference extra is imo worth it as this design makes the API more clean and much easier to maintain.

bee-node/src/lib.rs
samuel-rufi: Removed

bee-api/bee-rest-api/src/lib.rs
samuel-rufi: Removed

bee-api/bee-rest-api/src/endpoints/permission.rs
samuel-rufi: Done

bee-api/bee-rest-api/Cargo.toml
samuel-rufi: Done

bee-api/bee-rest-api/src/endpoints/routes/api/v1/message_children.rs
samuel-rufi: Changed! 👍

bee-api/bee-rest-api/src/endpoints/mod.rs
samuel-rufi: This is something that could be done true...
I guess however it would be better to not calculate the id here manually and instead rely on the node. Also the `Local` type in the node stores them separately. So even if the chance is small for an error, maybe keep it like it is

bee-api/bee-rest-api/src/endpoints/config.rs
rufsam: done 👍 @thibault-martinez

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +134,36 @@ impl RestApiConfigBuilder {
...
-            .unwrap_or_else(|| panic!("Unsupported protocol"));
-        let public_routes: Box<[String]> = self
-            .public_routes
-            .unwrap_or_else(|| DEFAULT_PUBLIC_ROUTES.iter().map(|s| s.to_string()).collect())
-            .into_boxed_slice();
-        let allowed_ips: Box<[IpAddr]> = self
-            .allowed_ips
-            .unwrap_or_else(|| DEFAULT_ALLOWED_IPS.to_vec())
-            .into_boxed_slice();
+            .unwrap_or_else(|| panic!("unsupported protocol"));
```
pvdrz: what about `.expect("unsupported protocol")` instead?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +134,36 @@ impl RestApiConfigBuilder {
...
+
+        let jwt_salt = self.jwt_salt.unwrap_or_else(|| DEFAULT_JWT_SALT.to_string());
+
+        let public_routes = {
+            let routes = self
+                .public_routes
+                .unwrap_or_else(|| DEFAULT_PUBLIC_ROUTES.iter().map(|r| r.to_string()).collect());
+            RegexSet::new(routes.iter().map(|r| route_to_regex(&r)).collect::<Vec<String>>())
+                .expect(&format!("invalid public route provided"))
```
pvdrz: I think you don't need the `&(format!(slice))` you can do just `slice`.

pvdrz (dismissed):

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: not sure what i did here 😃

grtlr (dismissed):

pvdrz (dismissed):

pvdrz (dismissed):

grtlr (dismissed):

grtlr (dismissed):

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -2,83 +2,56 @@
...
-    ROUTE_RECEIPTS,
-    ROUTE_RECEIPTS_AT,
-    ROUTE_TREASURY,
-    ROUTE_TRANSACTION_INCLUDED_MESSAGE,
-];
-pub(crate) const DEFAULT_ALLOWED_IPS: [IpAddr; 2] = [
-    IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
-    IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1)),
+/// Default JWT salt for REST API.
+pub(crate) const DEFAULT_JWT_SALT: &str = "Bee";
```
Alex6323, line 16: We already have a constant holding "Bee" in `bee-node` I think. Should we use that instead, otherwise we'ld have to make sure those 2 constants remain in sync, or would that not pose a real problem?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -2,83 +2,56 @@
...
+    "/api/v1/addresses*",
+    "/api/v1/treasury",
+    "/api/v1/receipts*",
 ];
+/// Default routes that are protected and need JWT authentication.
+pub(crate) const DEFAULT_PROTECTED_ROUTES: [&str; 2] = ["/api/v1/*", "/api/plugins/*"];
+/// Enables the proof-of-work feature on the node per default.
 pub(crate) const DEFAULT_FEATURE_PROOF_OF_WORK: bool = true;
+/// Default value for the white flag solidification timeout.
 pub(crate) const DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT: u64 = 2;
```
Alex6323: I assume that means seconds? What about using `Duration` type?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -100,15 +73,21 @@ impl RestApiConfigBuilder {
         self
     }
 
+    /// Sets the JWT salt.
+    pub fn jwt_salt(mut self, jwt_salt: &str) -> Self {
+        self.jwt_salt.replace(jwt_salt.to_string());
```
Alex6323: I'ld consider changing the argument to `jwt_salt: String` to make clear to the user that we need to own one of those.

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +131,36 @@ impl RestApiConfigBuilder {
...
+                .unwrap_or_else(|| DEFAULT_PUBLIC_ROUTES.iter().map(|r| r.to_string()).collect());
+            RegexSet::new(routes.iter().map(|r| route_to_regex(r)).collect::<Vec<String>>())
+                .expect("invalid public route provided")
+        };
+
+        let protected_routes = {
+            let routes = self
+                .protected_routes
+                .unwrap_or_else(|| DEFAULT_PROTECTED_ROUTES.iter().map(|r| r.to_string()).collect());
+            RegexSet::new(routes.iter().map(|r| route_to_regex(r)).collect::<Vec<String>>())
```
Alex6323: A few redundant map closures here I think. Isn't clippy complaining? :thinking: 
Also I think Rust can infer the type of Vec:
`routes.iter().map(route_to_regex).collect::<Vec<_>>()`

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -179,10 +170,17 @@ impl RestApiConfigBuilder {
 /// REST API configuration.
 #[derive(Clone)]
 pub struct RestApiConfig {
+    /// REST API binding address.
     pub(crate) binding_socket_addr: SocketAddr,
```
Alex6323: I'ld suggest renaming to `bind_socket_addr`.

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -197,14 +195,19 @@ impl RestApiConfig {
         self.binding_socket_addr
     }
 
+    /// Returns the JWT salt.
+    pub fn jwt_salt(&self) -> &String {
```
Alex6323: the return type can just be a `&str`

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -217,3 +220,14 @@ impl RestApiConfig {
         self.white_flag_solidification_timeout
     }
 }
+
+pub fn route_to_regex(route: &str) -> String {
```
Alex6323: this is `pub`, but doesn't require documentation? Can this be `pub(crate)` then?

bee-api/bee-rest-api/src/endpoints/config.rs
Alex6323: ah, that will cause a circular dep, I guess. Maybe we should move this constant to `bee-identity` eventually to have only one constant holding "Bee" :thinking:

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +131,36 @@ impl RestApiConfigBuilder {
...
+                .public_routes
+                .unwrap_or_else(|| DEFAULT_PUBLIC_ROUTES.iter().map(|r| r.to_string()).collect());
+            RegexSet::new(routes.iter().map(|r| route_to_regex(r)).collect::<Vec<String>>())
+                .expect("invalid public route provided")
+        };
+
+        let protected_routes = {
+            let routes = self
+                .protected_routes
+                .unwrap_or_else(|| DEFAULT_PROTECTED_ROUTES.iter().map(|r| r.to_string()).collect());
```
Adam-Gleave:
```suggestion
                .unwrap_or_else(|| DEFAULT_PROTECTED_ROUTES.iter().map(ToString::to_string).collect());
```

Adam-Gleave: (Similar below)

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -16,40 +16,82 @@ use storage::StorageBackend;
...
+
+pub struct ApiArgsFullNode<B: StorageBackend> {
+    pub node_id: PeerId,
+    pub node_keypair: Keypair,
+    pub rest_api_config: RestApiConfig,
+    pub protocol_config: ProtocolConfig,
+    pub network_id: NetworkId,
+    pub bech32_hrp: Bech32Hrp,
+    #[cfg(feature = "dashboard")]
+    pub dashboard_username: DashboardUsername,
```
Adam-Gleave: This is really nitpicky, but since this is reliant on a feature flag I would usually place this at the end of the list of fields, so there's some separation.

bee-api/bee-rest-api/src/endpoints/permission.rs
```diff
@@ -1,30 +1,124 @@
...
+                // Decode the JWT payload to find out how to validate it.
+                let jwt_payload = {
+                    let jwt_string = jwt.to_string();
+                    let split = jwt_string.split('.').collect::<Vec<&str>>();
+                    if split.len() < 3 {
+                        return Err(reject::custom(CustomRejection::Forbidden));
+                    }
+                    let encoded = split[1];
+                    let decoded = base64::decode(encoded).map_err(|_| reject::custom(CustomRejection::Forbidden))?;
+                    String::from_utf8(decoded).map_err(|_| CustomRejection::Forbidden)?
```
Adam-Gleave: I would maybe comment this a little (and add some newlines) just to explain what is going on here/why these operations are necessary.

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -72,47 +114,64 @@ where
...
-                peer_manager,
-                network_controller,
-                node_info,
-                bus,
-                message_requester,
-                requested_messages,
-                consensus_worker,
-            )
-            .recover(handle_rejection);
+            let health = warp::path("health").map(|| StatusCode::OK).recover(handle_rejection);
```
Alex6323: "health" should probably be a constant.

Alex6323: also: `warp` here still?

bee-api/bee-rest-api/src/endpoints/permission.rs
```diff
@@ -1,30 +1,124 @@
...
+    static ref DASHBOARD_ROUTES: RegexSet = {
+        let routes = vec![
+            "/api/v1/info",
+            "/api/v1/messages/*",
+            "/api/v1/outputs/*",
+            "/api/v1/addresses/*",
+            "/api/v1/milestones/*",
+            "/api/v1/peers*",
+        ];
+        RegexSet::new(routes.iter().map(|r| route_to_regex(r)).collect::<Vec<String>>()).unwrap()
```
Alex6323: same as earlier

bee-api/bee-rest-api/src/endpoints/routes/api/v1/messages_find.rs
```diff
@@ -40,35 +34,35 @@ pub(crate) fn filter<B: StorageBackend>(
...
-    };
+    let mut fetched =
+        match Fetch::<PaddedIndex, Vec<MessageId>>::fetch(&*args.storage, &hashed_index).map_err(|_| {
+            reject::custom(CustomRejection::ServiceUnavailable(
+                "can not fetch from storage".to_string(),
+            ))
+        })? {
+            Some(ids) => ids,
+            None => vec![],
+        };
```
Adam-Gleave: I would probably break this down into a couple statements.

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: Good idea!

bee-api/bee-rest-api/src/endpoints/permission.rs
```diff
@@ -1,30 +1,124 @@
...
+                if args.rest_api_config.public_routes.is_match(path.as_str()) {
+                    return Ok(());
+                }
+
+                let jwt = extract_jwt(&headers)?;
+                // Decode the JWT payload to find out how to validate it.
+                let jwt_payload = {
+                    let jwt_string = jwt.to_string();
+                    let split = jwt_string.split('.').collect::<Vec<&str>>();
+                    if split.len() < 3 {
```
Alex6323: Can you document magic numbers if they don't make sense putting them in a constant like it is probably the case here?

bee-api/bee-rest-api/src/endpoints/routes/api/v1/output.rs
```diff
@@ -71,12 +58,13 @@ pub(crate) async fn output<B: StorageBackend>(
...
-                        "unable to fetch the output".to_string(),
-                    ))
-                })?;
+                let consumed_output =
+                    Fetch::<OutputId, ConsumedOutput>::fetch(&*args.storage, &output_id).map_err(|e| {
+                        error!("unable to fetch the output: {}", e);
+                        reject::custom(CustomRejection::ServiceUnavailable(
+                            "unable to fetch the output".to_string(),
+                        ))
+                    })?;
```
Adam-Gleave: I prefer using `match` rather than `map_err` if there are any side-effects:

```suggestion
                let consumed_output = match Fetch::<OutputId, ConsumedOutput>::fetch(&*args.storage, &output_id) {
                    Err(e) => {
                        error!("unable to fetch the output: {}", e);
                        return reject::custom(CustomRejection::ServiceUnavailable(
                            "unable to fetch the output".to_string(),
                        ));
                    }
                    Ok(output) => output,
                };
```

bee-node/bee-plugin/bee-plugin-dashboard/src/lib.rs
```diff
@@ -237,6 +246,38 @@ where
     }
 }
 
+async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
+    let (http_code, err_code, reason) = match err.find() {
+        // handle custom rejections
```
Adam-Gleave: (Same below)
```suggestion
        // Handle custom rejections.
```

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -72,47 +114,64 @@ where
...
-        let tangle = node.resource::<Tangle<N::Backend>>();
-        let storage = node.storage();
-        let message_submitter = node.worker::<MessageSubmitterWorker>().unwrap().tx.clone();
-        let message_requester = node.worker::<MessageRequesterWorker>().unwrap().clone();
-        let requested_messages = node.resource::<RequestedMessages>();
-        let peer_manager = node.resource::<PeerManager>();
-        let network_controller = node.resource::<NetworkCommandSender>();
-        let node_info = node.info();
-        let bus = node.bus();
+        let args = Arc::new(ApiArgsFullNode {
```
Alex6323: Have you considered:
```rust
#[derive(Clone)]
struct ApiArgsFullNode(Arc<ApiArgsFullNodeInner>);
```
to make the API a bit nicer to use?

samuel-rufi: Yes, but would you implement `Deref` for it or how would you access the inner?

bee-api/bee-rest-api/src/endpoints/mod.rs
samuel-rufi: Yes, the mainnet branch uses warp, also I don't think we will replace it with axum here (the APIs are more and more diverging)
I wouldn't put "health" into a constant, having it directly here improves comprehensibility imo (for this case)
Also other paths for other endpoints are declared similarly

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: > A few redundant map closures here I think. Isn't clippy complaining? 🤔

I tried to change it but seems like it's necessary (and maybe thats the reason clippy isn't complaining).
Changing here the closures requires changes to the `route_to_regex ` or requires additional conversions. 

`routes.iter().map(route_to_regex)` would require `route_to_regex(&String)`, but the definition currently is `route_to_regex(&str)`.

> Also I think Rust can infer the type of Vec:
routes.iter().map(route_to_regex).collect::<Vec<_>>()

Changed this 👍

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: done

bee-api/bee-rest-api/src/endpoints/mod.rs
samuel-rufi: Good idea done :d

bee-api/bee-rest-api/src/endpoints/permission.rs
samuel-rufi: I documented the issue, it should be more clear now

bee-api/bee-rest-api/src/endpoints/mod.rs
Alex6323: alright then

bee-api/bee-rest-api/src/endpoints/mod.rs
Alex6323: yea, I'ld probably do that :+1: The two benefits are:
* method signatures become nicer 
* the `Arc` becomes an implementation detail, and can more easily replaced with something else without breaking APIs

bee-api/bee-rest-api/src/endpoints/config.rs
Alex6323: alright then :+1:

bee-api/bee-rest-api/src/endpoints/permission.rs
samuel-rufi: Added some more comments to the decoding process!

bee-api/bee-rest-api/src/endpoints/mod.rs
samuel-rufi: Implemented it this way:

```
pub struct ApiArgsFullNode<B: StorageBackend>(Arc<ApiArgsFullNodeInner<B>>);

impl<B: StorageBackend> Deref for ApiArgsFullNode<B> {
    type Target = Arc<ApiArgsFullNodeInner<B>>;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl<B: StorageBackend> Clone for ApiArgsFullNode<B> {
    fn clone(&self) -> Self {
        Self(self.0.clone())
    }
}
```

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: done!

bee-api/bee-rest-api/src/endpoints/routes/api/v1/messages_find.rs
samuel-rufi: done

bee-api/bee-rest-api/src/endpoints/config.rs
Alex6323: Are you sure?

bee-api/bee-rest-api/src/endpoints/config.rs
Alex6323: one instance was changed, but not the other in line 153

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -179,11 +176,18 @@ impl RestApiConfigBuilder {
...
+    pub(crate) jwt_salt: String,
+    /// Routes that are available for public use and don't need JWT authentication.
+    pub(crate) public_routes: RegexSet,
+    /// Routes that are protected and need JWT authentication.
+    pub(crate) protected_routes: RegexSet,
+    /// Enables/disables the proof-of-work feature on the node.
     pub(crate) feature_proof_of_work: bool,
-    pub(crate) white_flag_solidification_timeout: u64,
+    /// Describes the white flag solidification timeout.
+    pub(crate) white_flag_solidification_timeout: Duration,
```
Alex6323: I think those don't need to be `pub(crate)` because of the builder and the already `pub` getters.

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -10,42 +10,102 @@ pub mod rejection;
...
-) -> N::Builder
+pub fn init_full_node<N: Node>(init_config: InitConfigFullNode, node_builder: N::Builder) -> N::Builder
 where
     N::Backend: StorageBackend,
 {
-    node_builder.with_worker_cfg::<ApiWorkerFullNode>((rest_api_config, protocol_config, network_id, bech32_hrp))
+    node_builder.with_worker_cfg::<ApiWorkerFullNode>(init_config)
+}
+
+pub fn init_entry_node<N: Node>(init_config: InitConfigEntryNode, node_builder: N::Builder) -> N::Builder
```
Alex6323, line 52: this function doesn't require documentation, so I guess, it's a crate-internal function.  Can we therefore reduce its visibility to `pub(crate)`?

Alex6323: just checked `lib.rs`. Why are we not warning about missing docs?

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: thanks - fixed it

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: Now I'm sure :trollface:

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -1,84 +1,60 @@
...
+    "/api/v1/treasury",
+    "/api/v1/receipts*",
 ];
+/// Default routes that are protected and need JWT authentication.
+pub(crate) const DEFAULT_PROTECTED_ROUTES: [&str; 2] = ["/api/v1/*", "/api/plugins/*"];
+/// Enables the proof-of-work feature on the node per default.
 pub(crate) const DEFAULT_FEATURE_PROOF_OF_WORK: bool = true;
-pub(crate) const DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT: u64 = 2;
+/// Default value for the white flag solidification timeout.
+pub(crate) const DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT: Duration = Duration::new(2, 0);
```
Alex6323: nice, but can you use `Duration::from_secs(2)`  as it's more readable? 😊

bee-api/bee-rest-api/src/endpoints/permission.rs
Alex6323: maybe search for `collect::<Vec<String>>()` to find all those occurences where you can ommit the concrete type `String`.

Also...document the `unwrap`?

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: fixed

bee-api/bee-rest-api/src/endpoints/permission.rs
```diff
@@ -1,30 +1,132 @@
...
+
+fn extract_jwt(headers: &HeaderMap) -> Result<JsonWebToken, Rejection> {
+    let header = headers.get(AUTHORIZATION).ok_or(CustomRejection::Forbidden)?;
+    let auth_header = std::str::from_utf8(header.as_bytes()).map_err(|_| CustomRejection::Forbidden)?;
+
+    if !auth_header.starts_with(BEARER) {
+        return Err(reject::custom(CustomRejection::Forbidden));
+    }
+
+    Ok(JsonWebToken::from(auth_header.trim_start_matches(BEARER).to_owned()))
```
Alex6323, line 106: out of interest: BEARER contains a whitespace at the end, right? `trim_start_matches` is used to remove it here everytime it's called? I guess there's also a `start_matches`?

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: much better, changed! 😸

bee-api/bee-rest-api/src/endpoints/permission.rs
samuel-rufi: Fixed

bee-api/bee-rest-api/src/endpoints/mod.rs
samuel-rufi: > just checked lib.rs. Why are we not warning about missing docs?

that´s a good question, not fully sure why we had it like this - this should change ofc.
i think this however should be done in a separate PR do you agree? @Alex6323

samuel-rufi: > this function doesn't require documentation, so I guess, it's a crate-internal function. Can we therefore reduce its visibility to pub(crate)?

the node builder makes use of the functions so they must be public

bee-api/bee-rest-api/src/endpoints/permission.rs
samuel-rufi: Hmm seems like there is only `trim_start_matches` https://doc.rust-lang.org/std/primitive.str.html#method.trim_start_matches

bee-api/bee-rest-api/src/endpoints/mod.rs
Alex6323: right, I realized that when I saw that the `deny(missing_docs)` was disabled in `lib.rs`.

Alex6323: Can you enable it again, and fix the docs?

bee-api/bee-rest-api/src/endpoints/permission.rs
Alex6323: alright, I was just curious. Ignore me!

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +134,38 @@ impl RestApiConfigBuilder {
...
+                .expect("invalid protected route provided")
+        };
+
         let feature_proof_of_work = self.feature_proof_of_work.unwrap_or(DEFAULT_FEATURE_PROOF_OF_WORK);
-        let white_flag_solidification_timeout = self
-            .white_flag_solidification_timeout
-            .unwrap_or(DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT);
+
+        let white_flag_solidification_timeout = match self.white_flag_solidification_timeout {
+            Some(timeout) => Duration::new(timeout, 0),
```
Alex6323: `Duration::from_secs(timeout)` :trollface:

bee-api/bee-rest-api/src/endpoints/mod.rs
samuel-rufi: As discussed in Discord, this should be handled in a separate PR. I created an issue for it: https://github.com/iotaledger/bee/issues/1277

bee-api/bee-rest-api/src/endpoints/permission.rs
```diff
@@ -1,30 +1,133 @@
...
+    static ref DASHBOARD_ROUTES: RegexSet = {
+        let routes = vec![
+            "/api/v1/info",
+            "/api/v1/messages/*",
+            "/api/v1/outputs/*",
+            "/api/v1/addresses/*",
+            "/api/v1/milestones/*",
+            "/api/v1/peers*",
+        ];
+        // It is safe to use unwrap() because route_to_regex() generates valid expressions.
```
Alex6323:
```suggestion
        // Panic: unwrapping is fine because all strings in `routes` can be turned into valid regular expressions.
```

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: Thanks I will never use Duration::new() again 😃

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -1,84 +1,60 @@
...
+    #[serde(alias = "jwtSalt")]
+    jwt_salt: Option<String>,
+    /// Routes that are available for public use and don't need JWT authentication.
     #[serde(alias = "publicRoutes")]
     public_routes: Option<Vec<String>>,
-    #[serde(alias = "allowedIps")]
-    allowed_ips: Option<Vec<IpAddr>>,
+    /// Routes that are protected and need JWT authentication.
+    #[serde(alias = "protectedRoutes")]
+    protected_routes: Option<Vec<String>>,
```
pvdrz, line 53: Classical question: does the difference between `Some(vec![])` and `None` actually matter?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -100,15 +76,21 @@ impl RestApiConfigBuilder {
         self
     }
 
+    /// Sets the JWT salt.
+    pub fn jwt_salt(mut self, jwt_salt: String) -> Self {
```
pvdrz: I'd call all these methods `with_<field>` or `set_<field>` instead, what do you think?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +134,38 @@ impl RestApiConfigBuilder {
...
-            .into_boxed_slice();
+            .expect("unsupported protocol");
+
+        let jwt_salt = self.jwt_salt.unwrap_or_else(|| DEFAULT_JWT_SALT.to_string());
+
+        let public_routes = {
+            let routes = self
+                .public_routes
+                .unwrap_or_else(|| DEFAULT_PUBLIC_ROUTES.iter().map(ToString::to_string).collect());
+            RegexSet::new(routes.iter().map(|r| route_to_regex(r)).collect::<Vec<_>>())
```
pvdrz, line 145: I think you can do `map(route_to_regex)` directly

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +134,38 @@ impl RestApiConfigBuilder {
...
+                .expect("invalid public route provided")
+        };
+
+        let protected_routes = {
+            let routes = self
+                .protected_routes
+                .unwrap_or_else(|| DEFAULT_PROTECTED_ROUTES.iter().map(ToString::to_string).collect());
+            RegexSet::new(routes.iter().map(|r| route_to_regex(r)).collect::<Vec<_>>())
```
pvdrz, line 153: same

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +134,38 @@ impl RestApiConfigBuilder {
...
+                .expect("invalid protected route provided")
+        };
+
         let feature_proof_of_work = self.feature_proof_of_work.unwrap_or(DEFAULT_FEATURE_PROOF_OF_WORK);
-        let white_flag_solidification_timeout = self
-            .white_flag_solidification_timeout
-            .unwrap_or(DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT);
+
+        let white_flag_solidification_timeout = match self.white_flag_solidification_timeout {
```
pvdrz: I think you can do something like `self.white_flag_solidification_timeout.map_or_else(Duration::from_secs, DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT)`.

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -213,7 +221,18 @@ impl RestApiConfig {
     }
 
     /// Returns the white flag solidification timeout.
-    pub fn white_flag_solidification_timeout(&self) -> u64 {
-        self.white_flag_solidification_timeout
+    pub fn white_flag_solidification_timeout(&self) -> &Duration {
```
pvdrz: `Duration` implements `Copy` so I think it is better to return the value instead of a reference,

bee-api/bee-rest-api/src/endpoints/filters.rs
```diff
@@ -3,98 +3,12 @@
...
-}
-
-pub(crate) fn with_consensus_worker(
-    consensus_worker: mpsc::UnboundedSender<ConsensusWorkerCommand>,
-) -> impl Filter<Extract = (mpsc::UnboundedSender<ConsensusWorkerCommand>,), Error = Infallible> + Clone {
-    warp::any().map(move || consensus_worker.clone())
+pub(crate) fn with_args<B: StorageBackend>(
+    args: ApiArgsFullNode<B>,
+) -> impl Filter<Extract = (ApiArgsFullNode<B>,), Error = Infallible> + Clone {
+    warp::any().map(move || args.clone())
```
pvdrz, line 13: do you need both `move` and `.clone()`?

bee-api/bee-rest-api/src/endpoints/permission.rs
```diff
@@ -1,30 +1,133 @@
...
+                    // Get the base64 encoded payload.
+                    let encoded_payload = split[1];
+                    // Base64-decode the payload to bytes.
+                    let decoded_payload_bytes =
+                        base64::decode(encoded_payload).map_err(|_| reject::custom(CustomRejection::Forbidden))?;
+                    // Create a UTF-8 string from the decoded bytes.
+                    String::from_utf8(decoded_payload_bytes).map_err(|_| CustomRejection::Forbidden)?
+                };
+
+                if jwt_payload.contains(&format!("\"aud\":\"{}\"", API_AUDIENCE_CLAIM)) {
```
pvdrz: is there any way to put the argument of `contains` into a constant to avoid computing it on every call?

bee-api/bee-rest-api/src/endpoints/routes/health.rs
```diff
@@ -3,21 +3,14 @@
...
-use crate::endpoints::{
-    config::ROUTE_HEALTH,
-    filters::{with_peer_manager, with_tangle},
-    permission::has_permission,
-    storage::StorageBackend,
-};
+use crate::endpoints::{filters::with_args, storage::StorageBackend, ApiArgsFullNode};
 
 const HEALTH_CONFIRMED_THRESHOLD: u32 = 2; // in milestones
 const HEALTH_MILESTONE_AGE_MAX: u64 = 5 * 60; // in seconds
```
pvdrz: I know this is not part of the changes but could this be a `Duration`s instead?

pvdrz (changes requested):

bee-api/bee-rest-api/src/endpoints/config.rs
grtlr: If not, I'd probably go for just `Vec<String>`.

bee-api/bee-rest-api/src/endpoints/permission.rs
```diff
@@ -1,30 +1,133 @@
...
+                    if validate_api_jwt(&jwt, &args).is_ok()
+                        && args.rest_api_config.protected_routes().is_match(path.as_str())
+                    {
+                        return Ok(());
+                    }
+                } else {
+                    #[cfg(feature = "dashboard")]
+                    if jwt_payload.contains(&format!("\"aud\":\"{}\"", DASHBOARD_AUDIENCE_CLAIM))
```
grtlr: Same as Christian's comment above.

bee-api/bee-rest-api/src/endpoints/config.rs
thibault-martinez: It does matter, `None` notifies that it wasn't set at all, triggering the default value for it.

samuel-rufi: So let's keep it

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: Good point, added the `with` prefix

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: Hmm also @Alex6323 mentioned this, but doesn't really work
Seems like it's necessary (and maybe thats the reason also clippy isn't complaining).
Changing here the closures requires changes to the route_to_regex or requires additional conversions.

`routes.iter().map(route_to_regex)` would require `route_to_regex(&String)`, but it's defined as `route_to_regex(&str)`.
Or do you have any other thought how to change this?

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: done

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: done

bee-api/bee-rest-api/src/endpoints/filters.rs
samuel-rufi: Yes, seems like

Without `move` it may outlive borrowed value `args`.
And witout `clone`: 

```
13 |     warp::any().map(move || args)
   |                 --- ^^^^^^^^----
   |                 |   |       |
   |                 |   |       closure is `FnOnce` because it moves the variable `args` out of its environment
   |                 |   this closure implements `FnOnce`, not `Fn`
   |                 the requirement to implement `Fn` derives from here
```

bee-api/bee-rest-api/src/endpoints/permission.rs
samuel-rufi: I changed it to:
```
pub const API_AUDIENCE_CLAIM: &str = "api";
pub const DASHBOARD_AUDIENCE_CLAIM: &str = "dashboard";
const API_JWT_HINT: &str = "\"aud\":\"api\"";
#[cfg(feature = "dashboard")]
const DASHBOARD_JWT_HINT: &str = "\"aud\":\"dashboard\"";
```

However it would be better if we could reuse the `API_AUDIENCE_CLAIM` and `DASHBOARD_AUDIENCE_CLAIM`. Do you have a  better idea in mind? Maybe use some static refs?

bee-api/bee-rest-api/src/endpoints/routes/health.rs
samuel-rufi: fixed

bee-api/bee-rest-api/src/endpoints/config.rs
pvdrz: you could change `route_to_regex` argument to receive `impl core::ops::Deref<Target = str>` instead of `&str`. But I think I would leave this unmodified instead.

bee-api/bee-rest-api/src/endpoints/permission.rs
pvdrz: I think you can do
```rust
const API_JWT_HINT: &str =  concat!("\"aud\":\"", API_AUDIENCE_CLAIM, "\"");
const DASHBOARD_JWT_HINT: &str =  concat!("\"aud\":\"", DASHBOARD_AUDIENCE_CLAIM, "\"");

File: bee-api/bee-rest-api/src/endpoints/filters.rs (hunk 44 above)
Comment (pvdrz) [reply bee-api/bee-rest-api/src/endpoints/filters.rs:13]: So weird, I'm really looking forward to your axum PR :trollface:

File: bee-api/bee-rest-api/src/endpoints/permission.rs (hunk 45 above)
Comment (samuel-rufi) [reply bee-api/bee-rest-api/src/endpoints/permission.rs]: Hmm not sure if this works:

```
error: expected a literal
  --> bee-api\bee-rest-api\src\endpoints\permission.rs:20:51
   |
20 | const API_JWT_HINT: &str =  concat!("\"aud\":\"", API_AUDIENCE_CLAIM, "\"");
   |                                                   ^^^^^^^^^^^^^^^^^^
   |
   = note: only literals (like `"foo"`, `42` and `3.14`) can be passed to `concat!()`
```

Comment (samuel-rufi) [reply bee-api/bee-rest-api/src/endpoints/permission.rs]: Limited with the constants, so we will simply do:

```
pub const API_AUDIENCE_CLAIM: &str = "api";
pub const DASHBOARD_AUDIENCE_CLAIM: &str = "dashboard";
const API_JWT_HINT: &str = "\"aud\":\"api\"";
#[cfg(feature = "dashboard")]
const DASHBOARD_JWT_HINT: &str = "\"aud\":\"dashboard\"";
```

**Review by grtlr — DISMISSED** [review]

**Review by Alex6323 — DISMISSED** [review]

**Review by Alex6323 — DISMISSED** [review]

**Review by grtlr — DISMISSED** [review]

**Review by pvdrz — DISMISSED** [review]

File: documentation/docs/examples.md [hunk 48]
```diff
@@ -5,3 +5,17 @@ keywords:
 - examples
 ---
 # Examples
+
+## Generate a JWT for the REST API for your node
+To generate a JWT for the REST API for your node run following command:
```
thibault-martinez, line 11:
```suggestion

To generate a JWT for the REST API for your node run following command: 
```

grtlr (dismissed):

pvdrz (approved):

grtlr (approved):

PR #967, opened 2022-01-15
Make it easier to copy & paste p2p-identity and password into config

atomicMSR:
# Description of change

Change the output of the `p2p-identity` and `password` tool commands to match the format of the .toml config file, for easier copy&pasting.

## Type of change

Choose a type of change, and delete any options that are not relevant.

- Enhancement (a non-breaking change which adds functionality)

## How the change has been tested

Local test run.

## Change checklist

Tick the boxes that are relevant to your changes, and delete any items that are not.

- [ X I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code

PR #986, opened 2022-01-19
Switch to `.key` file for identity

### LINKED ISSUE #325: Move secret key out of the config into a separate protected file
This is to ensure, that users don't accidentally share their secret key the `PeerId` is derived from as configs happen to be copied quite often without caution.

grtlr:
# Description of change

Store the P2P identity in a separate `.pem` file, which is automatically created if it doesn't exist yet. This should streamline the process of setting up and running a node. 

Inspired by PR #967.

## Type of change

Ran locally, with different configurations.

## Change checklist


bee-node/src/entrynode/config.rs
```diff
@@ -33,12 +33,13 @@ impl EntryNodeConfig {
...
 
-impl<S: NodeStorageBackend> From<NodeConfig<S>> for EntryNodeConfig {
-    fn from(node_cfg: NodeConfig<S>) -> Self {
+    pub fn from<S>(local: Local, node_cfg: NodeConfig<S>) -> Self
+    where
+        S: NodeStorageBackend,
+    {
         Self {
-            local: node_cfg.local,
+            local: local,
```
Alex6323: simplify to just `local,`

bee-node/src/main.rs
```diff
@@ -36,41 +41,58 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
 
     // Initialize the logger.
     let logger_cfg = config.logger_config().clone();
     bee_common::logger::logger_init(logger_cfg)?;
 
+    // Establish identity.
+
+    // The identity used to be stored in the config file.
+    if has_identity_field {
+        warn!("The config file contains an identity field which will be ignored.");
```
Alex6323: Should we give a hint to the user, what to do about it?

bee-node/src/main.rs
```diff
@@ -36,41 +41,58 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                "There is no identity file at `{}`. Generating a new one.",
+                identity_path.display()
+            );
+            let keypair = Keypair::generate();
+            if let Err(e) = write_keypair_to_pem_file(identity_path, &keypair) {
+                panic!("Failed to write PEM file: {}", e);
+            };
+            keypair
+        }
+        Err(e) => panic!("Failed to establish identity: {}", e),
```
Alex6323: What about: "Failed to create a new identity:" ?

bee-node/src/pem_file.rs
```diff
@@ -0,0 +1,61 @@
...
+use pem::Pem;
+
+use std::{fs, path::Path};
+
+const PRIVATE_KEY_TAG: &str = "PRIVATE KEY";
+
+#[derive(Debug, thiserror::Error)]
+pub enum PemFileError {
+    #[error("reading the identity file failed: {0}")]
+    FileRead(std::io::Error),
```
Alex6323: I think the first 3 error variants could be simply: `Read`, `Write`, `Parse` because the enum already contains the word `file`. I am not sure about the last one, but I think I'ld prefer swapping verb and noun, so `DecodeKeypair`.

bee-node/src/pem_file.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+fn pem_entry_to_keypair(pem_entry: String) -> Result<Keypair, PemFileError> {
+    let mut pem = pem::parse(pem_entry).or(Err(PemFileError::InvalidPemFile))?;
+    Keypair::decode(&mut pem.contents).or(Err(PemFileError::KeypairDecode))
+}
+
+pub fn read_keypair_from_pem_file<P: AsRef<Path>>(path: P) -> Result<Keypair, PemFileError> {
+    match fs::read_to_string(path) {
+        // Todo: Should we handle PEM files with multiple keys?
```
Alex6323: I guess we should probably enforce one key only, so if multiple keys are found, we probably should panic, but that's just my first guess. Any other idea how to go about it?

.gitignore
```diff
@@ -5,8 +5,12 @@ target/
 # These are backup files generated by rustfmt
 **/*.rs.bk
 
+# Private key files
+*.pem
+
 # bee-node
 bee-node/config.toml
+bee-node/identity.toml
```
Alex6323: I guess that line is wrong. It should be `bee-node/identity.pem`, right? Do we need the general exclusion of `.pem` files (above) then?

bee-node/src/main.rs
grtlr: Actually, we should never reach this arm, because `read_keypair_from_pem_file` only ever returns `PemFileError::Read`.

bee-node/src/pem_file.rs
grtlr: We could iterate over all keys, check if their tag matches `PRIVATE_KEY` and then take the first one. But `panic` on multiple entries also sounds reasonable.

What do you say @thibault-martinez ?

bee-node/src/config.rs
```diff
@@ -143,14 +143,7 @@ impl<S: NodeStorageBackend> NodeConfigBuilder<S> {
...
     /// Returns the built node config.
-    pub fn finish(self) -> NodeConfig<S> {
-        // Create the local node entity.
-        let local = if let Some(keypair) = self.identity {
-            Local::from_keypair(keypair, self.alias).expect("error creating local from hex encoded keypair")
-        } else {
-            Local::new(self.alias)
-        };
-
+    pub fn finish(self) -> (Option<String>, bool, NodeConfig<S>) {
```
thibault-martinez: Actually the alias needs to be in the `NodeConfig`, it was there through `local` but then it's not anymore. Just add `alias` to the config and you don't have to return it anymore.

thibault-martinez: As for the bool, I don't really like it but I guess it's temporary, we leave it for one release for people to migrate and then we can remove it at some point.

bee-node/src/pem_file.rs
thibault-martinez: I agree we shouldn't handle multiple keys but we should just error, not panic.

bee-node/src/main.rs
```diff
@@ -36,41 +41,58 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    let keypair = match read_keypair_from_pem_file(&identity_path) {
+        Ok(keypair) => keypair,
+        Err(PemFileError::Read(_)) => {
+            info!(
+                "There is no identity file at `{}`. Generating a new one.",
+                identity_path.display()
+            );
+            let keypair = Keypair::generate();
+            if let Err(e) = write_keypair_to_pem_file(identity_path, &keypair) {
+                panic!("Failed to write PEM file: {}", e);
```
thibault-martinez: Log an error and return instead of panic.

bee-node/src/main.rs
```diff
@@ -36,41 +44,103 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+
+            keypair
+        }
+        Err(e) => {
+            error!("Could not extract private key from PEM file: {}", e);
+            std::process::exit(-1);
+        }
+    };
+
+    let local = Local::from_keypair(keypair, Some(config.alias().clone()));
```
thibault-martinez: The alias shouldn't be an Option anymore. By this time it's already been defaulted so no need to add a Some

bee-node/src/lib.rs
```diff
@@ -11,6 +11,7 @@ mod core;
 mod entrynode;
 mod fullnode;
 mod local;
+mod pem_file;
```
thibault-martinez: I think I would call this simply `pem` for simplicity

bee-node/src/config.rs
```diff
@@ -9,10 +9,9 @@
 
 use crate::{
     cli::ClArgs,
-    local::Local,
     plugins::mqtt::config::{MqttConfig, MqttConfigBuilder},
     storage::NodeStorageBackend,
-    util, BECH32_HRP_DEFAULT, NETWORK_NAME_DEFAULT,
+    util, BECH32_HRP_DEFAULT, LOCAL_ALIAS_DEFAULT, NETWORK_NAME_DEFAULT,
```
thibault-martinez: Can you move LOCAL_ALIAS_DEFAULT to this file ? And drop the LOCAL part

bee-node/src/pem_file.rs
```diff
@@ -0,0 +1,100 @@
...
+        // Safety: We just checked the length.
+        Keypair::decode(&mut pems[0].contents).or(Err(PemFileError::DecodeKeypair))
+    } else {
+        Err(PemFileError::MultipleEntries)
+    }
+}
+
+pub fn read_keypair_from_pem_file<P: AsRef<Path>>(path: P) -> Result<Keypair, PemFileError> {
+    match fs::read_to_string(path) {
+        // Todo: Should we handle PEM files with multiple keys?
```
thibault-martinez: Remove ?

bee-node/src/lib.rs
grtlr: `pem` is already the name of the crate we are using to parse these files.

Alex6323: maybe `pemf` ? :jk:

Alex6323: how about `identity`? This makes the module independent of what we use internally to store that identity.

thibault-martinez: I think I would prefer that

bee-node/src/main.rs
```diff
@@ -36,41 +44,103 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
     Ok(())
 }
 
-fn deserialize_config(cl_args: ClArgs) -> NodeConfig<Storage> {
-    #[cfg(feature = "rocksdb")]
-    let config = match NodeConfigBuilder::<Storage>::from_file(
-        cl_args.config_path().unwrap_or_else(|| Path::new(CONFIG_PATH)),
-    ) {
+#[derive(Debug, thiserror::Error)]
+enum MigrationError {
```
Alex6323: Please correct my older Self with this PR, and change the variants of this enum to `DecodeHex`, `DecodeKeypair`.

grtlr:
When I trying to use the `pkcs8` crate, I run into the following issue: https://github.com/RustCrypto/formats/issues/350

pvdrz (approved):

Alex6323 (approved):

bee-node/config.chrysalis-comnet.toml
```diff
@@ -1,6 +1,5 @@
 # Duplicate this file to config.toml to use it
 
-# identity   = ""
```
Adam-Gleave, line 3: Should these lines be removed now?

bee-node/CHANGELOG.md
```diff
@@ -10,13 +10,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - New `./dashboard` route;
+- Automatically generate new identity if none is present;
 
 ### Changed
 
+- P2P identity is now stored in a `*.pem` file;
```
thibault-martinez: .key

bee-node/CHANGELOG.md
```diff
@@ -10,13 +10,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - New `./dashboard` route;
+- Automatically generate new identity if none is present;
```
thibault-martinez: Maybe also say something about migration ?

bee-node/src/config.rs
```diff
@@ -77,8 +83,11 @@ impl<S: NodeStorageBackend> NodeConfig<S> {
 #[derive(Default, Deserialize)]
 #[must_use]
 pub struct NodeConfigBuilder<S: NodeStorageBackend> {
+    // We don't store the identity in the config file anymore.
+    // This is here for legacy reasons to warn the user of that fact.
     #[serde(rename = "identity")]
-    pub(crate) identity: Option<String>,
+    #[deprecated(since = "0.2.0")]
```
thibault-martinez: 0.3.0 actually

bee-node/src/main.rs
```diff
@@ -36,31 +44,103 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+            // existing config or generate a new identity.
+            let keypair = if let Some(identity_encoded) = identity_field {
+                warn!(
+                    "There is no identity file at `{}`. Migrating identity from the existing config file.",
+                    identity_path.display(),
+                );
+
+                migrate_keypair(identity_encoded).unwrap_or_else(|e| {
+                    error!("Failed to migrate keypair: {}", e);
+                    std::process::exit(-1);
```
thibault-martinez: Why exit ? The main can return an error.

bee-node/src/main.rs
```diff
@@ -36,31 +44,103 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                    "There is no identity file at `{}`. Generating a new one.",
+                    identity_path.display()
+                );
+
+                Keypair::generate()
+            };
+
+            if let Err(e) = write_keypair_to_pem_file(identity_path, &keypair) {
+                error!("Failed to write PEM file: {}", e);
+                std::process::exit(-1);
```
thibault-martinez: Same

bee-node/src/main.rs
```diff
@@ -36,31 +44,103 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+            };
+
+            keypair
+        }
+        Err(e) => {
+            error!("Could not extract private key from PEM file: {}", e);
+            std::process::exit(-1);
```
thibault-martinez: Same

bee-node/src/main.rs
```diff
@@ -36,31 +44,103 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-        start_fullnode(config).await;
+        start_fullnode(local, config).await;
     }
 
     Ok(())
 }
 
-fn deserialize_config(cl_args: ClArgs) -> NodeConfig<Storage> {
+#[derive(Debug, thiserror::Error)]
+enum MigrationError {
```
thibault-martinez: Maybe prefix Identity ? Migration is very broad.

bee-node/src/main.rs
```diff
@@ -36,31 +44,103 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    #[error("hex decoding failed")]
+    DecodeHex,
+    #[error("keypair decoding failed")]
+    DecodeKeypair,
+    #[error("invalid keypair")]
+    InvalidKeypair,
+}
+
+fn migrate_keypair(encoded: String) -> Result<Keypair, MigrationError> {
```
thibault-martinez: If it relates to the identity, maybe it should be part of the module ?

grtlr: I've kept it here, so we don't populate the module with functions that will be removed in the next versions anyways. It's also set up so that the compiler guides us in removing all remnants of the previous identity file. Of course, we could also move it to the module and slap a `#[deprecated]` on it, if you prefer that.

thibault-martinez (approved):

PR #988, opened 2022-01-19
Docs - Add test build action

lucas-tortora:
# Description of change

As any build error in the docs can stop the wiki from updating, I've added a `test build` action to catch these before they are merged into the published branches, in this case `dev`

## Type of change

- Documentation Fix

The action was tested in my fork for both passing and failure.

## Change checklist


lucas-tortora:
@Dr-Electron, could you please review this when you have some time? Thanks!

.github/workflows/test-docs-build.yml
```diff
@@ -0,0 +1,41 @@
+name: Test Docs Wiki Build
+
+on:
+  workflow_dispatch:
```
grtlr: Since it's empty we could remove this.

```suggestion
```

.github/workflows/test-docs-build.yml
```diff
@@ -0,0 +1,41 @@
...
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+      - name: Install dependencies
+        run: |
+          cd documentation
+          if [ -e package-lock.json ]; then
```
grtlr: Maybe we should always have a `package-lock.json` for reproducibility? In that case the check would be unnecessary.

.github/workflows/test-docs-build.yml
```diff
@@ -0,0 +1,41 @@
...
+          pwd
+      - name: Setup Wiki and Test Build
+        run: |
+          cd documentation
+          export config=$(cat EXTERNAL_DOCS_CONFIG)
+          npm run setup
+          cd ../
+          rsync -av --progress ./ ./documentation/local-wiki/iota-wiki/external/bee --exclude .git --exclude node_modules --exclude local-wiki
+          cd documentation/local-wiki/iota-wiki
+          yarn
```
grtlr, line 35: Why is `yarn` mixed with `npm`?

.github/workflows/test-docs-build.yml
```diff
@@ -0,0 +1,41 @@
...
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+      - name: Install dependencies
+        run: |
+          cd documentation
+          if [ -e package-lock.json ]; then
+          npm ci
+          else
+          npm i
```
grtlr: Personally, I prefer the long form of these commands because it helps to document the code.

.github/workflows/test-docs-build.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          cd documentation
+          export config=$(cat EXTERNAL_DOCS_CONFIG)
+          npm run setup
+          cd ../
+          rsync -av --progress ./ ./documentation/local-wiki/iota-wiki/external/bee --exclude .git --exclude node_modules --exclude local-wiki
+          cd documentation/local-wiki/iota-wiki
+          yarn
+          export replace_string='/\* AUTO GENERATED EXTERNAL DOCS CONFIG \*/'
+          perl -0pe 's#$ENV{replace_string}#$ENV{config}#' docusaurus.config.js > docusaurus.config.js.cpy
```
grtlr, line 38: It would be great if these steps would be documented a little bit using comments.

Dr-Electron (dismissed):

Dr-Electron:
lgtm. But why don't we use the iota-wiki-cli anymore?

grtlr (changes requested):
Thanks for submitting the PR! What do you think about it @thibault-martinez? I also left a couple of minor comments.

lucas-tortora:
> lgtm. But why don't we use the iota-wiki-cli anymore?

We are, but the way it's set up the process doesn't exit on error.  So it's not useful in this particular scenario.

.github/workflows/test-docs-build.yml
lucas-tortora: The iota-wiki-cli is installed with npm.  This is used to clone an instance of the iota-wiki, which uses docusarus, which uses yarn.  If there's a better/less confusing way to do this we could modify the script.

.github/workflows/test-docs-build.yml
lucas-tortora: Is there a long form for `npm ci`?

.github/workflows/test-docs-build.yml
```diff
@@ -0,0 +1,44 @@
+name: Test Docs Wiki Build
+
+on:
+  pull_request:
+    branches:
+      - dev
+jobs:
+  checks:
+    if: github.event_name == 'pull_request'
```
thibault-martinez: Out of curiosity, is this necessary since we already have a `on.pull_request` ?

Dr-Electron:
> > lgtm. But why don't we use the iota-wiki-cli anymore?
> 
> We are, but the way it's set up the process doesn't exit on error. So it's not useful in this particular scenario.

I don't fully remember the code of the cli, but couldn't we change that? That way also locally you can du a quick check without running everything? Would that make sense?

lucas-tortora:
> > > lgtm. But why don't we use the iota-wiki-cli anymore?
> > 
> > 
> > We are, but the way it's set up the process doesn't exit on error. So it's not useful in this particular scenario.
> 
> I don't fully remember the code of the cli, but couldn't we change that? That way also locally you can du a quick check without running everything? Would that make sense?

I don't really know how to do that, but if it's possible I'd be 100% onboard.  It would make everything simpler and tidier.

Dr-Electron:
I can think about it this weekend. We can still merge this in the meantime. Don't know when I have the time to look at it. Already wanted to do that a long time ago 😅

lucas-tortora:
> I can think about it this weekend. We can still merge this in the meantime. Don't know when I have the time to look at it. Already wanted to do that a long time ago sweat_smile

We had added a test method already, I hope it helps.  We would just need to find a way for the gh action to know if it failed or not.

Dr-Electron:
> > I can think about it this weekend. We can still merge this in the meantime. Don't know when I have the time to look at it. Already wanted to do that a long time ago sweat_smile
> 
> We had added a test method already, I hope it helps. We would just need to find a way for the gh action to know if it failed or not.

I know. That's exactly what I wanted to work on... I just never find the time 😆

.github/workflows/test-docs-build.yml
grtlr: Sorry, it's been a while since I last used `npm`. `npm ci` is fine, of course.

.github/workflows/test-docs-build.yml
grtlr: Ok, cool. Thanks for clearing that up!

thibault-martinez (approved):

grtlr (approved):

pvdrz (approved):

Issue #976, opened 2022-01-17
Autopeering Error processing command. Cause: Not present peer


thibault-martinez:
`2022-01-17 12:01:04 (UTC) bee_gossip::service::host                  ERROR Error processing command. Cause: Not present peer:`

Issue #975, opened 2022-01-17
Invalid peering request


thibault-martinez:
`2022-01-17 11:57:23 (UTC) bee_autopeering::peering::manager          WARN  Received invalid peering request from 12D3KooWHqdw3M7E. Reason: PeerNotVerified`

Issue #974, opened 2022-01-17
Incorrect autopeering peer denial


thibault-martinez:
It says 3/4 but then it fails to connect a new one.

```
2022-01-17 11:36:27 (UTC) bee_gossip::service::host                  INFO  Connected peers: known 0/0 unknown 2/4 discovered 3/4.
2022-01-17 11:36:27 (UTC) bee_gossip::service::host                  INFO  Trying to connect to: NGdNpS (NGdNpS).
2022-01-17 11:36:27 (UTC) bee_gossip::network::host                  WARN  Dialing peer NGdNpS denied. Cause: ExceedsDiscoveredPeerLimit(4)
2022-01-17 11:36:27 (UTC) bee_gossip::network::host                  WARN  Dialing peer NGdNpS failed. Cause: Dialing peer NGdNpS was denied.
```

thibault-martinez:
Even weirder...

```
2022-01-17 11:58:22 (UTC) bee_gossip::service::host                  INFO  Connected peers: known 0/0 unknown 3/4 discovered 0/4.
2022-01-17 11:58:22 (UTC) bee_gossip::service::host                  INFO  Trying to connect to: rn6zCY (rn6zCY).
2022-01-17 11:58:22 (UTC) bee_gossip::service::host                  INFO  Trying to connect to: Rrvp4M (Rrvp4M).
2022-01-17 11:58:22 (UTC) bee_gossip::service::host                  INFO  Trying to connect to: rnBNvE (rnBNvE).
2022-01-17 11:58:22 (UTC) bee_gossip::service::host                  INFO  Trying to connect to: oHEdFa (oHEdFa).
2022-01-17 11:58:22 (UTC) bee_gossip::network::host                  WARN  Dialing peer rn6zCY denied. Cause: ExceedsDiscoveredPeerLimit(4)
2022-01-17 11:58:22 (UTC) bee_gossip::network::host                  WARN  Dialing peer rn6zCY failed. Cause: Dialing peer rn6zCY was denied.
```

Alex6323:
yup, thx

Issue #920, opened 2022-01-10
Autopeering is slower than its hornet counterpart


thibault-martinez:

Alex6323:
Can you provide a little bit more detail? To come to this conclusion you would need at least 10 samples from both node implementations respectively to be somewhat on the statistical safe side. And even more importantly add to your description what's the start and end point of the measurement. I am assuming you mean the time it takes until a valid gossip peer has been connected and the node starts syncing!?.

Although I cannot reject your statement right away, I want you to consider the following things:
(a) autopeering is a probabilistic system by design: 
     - if we are queried we pick a random set of at most 6 peers to send in our discovery response; same goes for the remote peer.
     - each auto-peer creates 2 random salts (every 2 hours), by which distances to possible candidates are determined: when we (as a new peer) try to get integrated into the network possible peers very likely already have full neighborhoods, and would only accept us if we got "lucky" and are evaluated to have a shorter distance than their current furthest-away neighbor. If we're unlucky this can take some time.
 (b) autopeering in general is not supposed to work as fast as possible. For example: the time interval in which we query a peer (to discover new peers) is identical in Bee and Hornet and set to 60(!!!) seconds. Hence finding a suitable peer can take some time.

Dr-Electron:
I will do some tests tonight. But most of the time it takes 3-6min for a peer to show op on the dashboard which !seemed! slower than it is in Hornet. As i said, i will investigate more

Dr-Electron:
This are the results if I look at the logs. Start time was the autopeering start time and I stopped once I got a peer which send me new milestones.
But I think I made a mistake. Because if I stop and start the node to quickly the other peer may connect with me right at the beginning? Also bee seems to start the gossip plugin after auto peering but I think Hornet starts it before. So there could be a difference of some sec.
But in general you can see that sometimes Bee takes way longer to connect to its first peer. 
I didn't have too much time today. So I can do some more test sometime later if needed.

```
Hornet:
21:37:15Z	21:38:25Z  1m10s
21:51:40Z	21:52:59Z  1m19s
22:00:07Z	22:00:36Z  29s
22:02:18Z	22:04:07Z  1m49s
22:04:48Z	22:06:37Z  1m49s

Bee:
22:20:26    22:23:34  2m9s
22:33:15    22:38:23  5m8s
22:42:31    22:43:09  38s
22:45:11    22:45:20  9s
22:49:33    22:54:42  5m9s
```

PR #989, opened 2022-01-19
Autopeering fixes for RC3

Alex6323:
# Description of change

Fixes a few more bugs with the autopeering integration and interplay with `bee-gossip`.

#920 (most of the time it should now be fairly quick to find its first valid peer)
#970 
#974 
#975 
#976 

## Type of change

VPS.

## Change checklist

### DIFF
--- a/bee-network/bee-autopeering/Cargo.toml
+++ b/bee-network/bee-autopeering/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bee-autopeering"
-version = "0.1.1"
+version = "0.1.2"
 authors = [ "IOTA Stiftung" ]
 edition = "2021"
 description = "Allows peers in the same IOTA network to automatically discover each other."
--- a/bee-network/bee-autopeering/Cargo.toml
+++ b/bee-network/bee-autopeering/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bee-autopeering"
-version = "0.1.2"
+version = "0.2.0"
 authors = [ "IOTA Stiftung" ]
 edition = "2021"
 description = "Allows peers in the same IOTA network to automatically discover each other."
--- a/bee-network/bee-autopeering/src/discovery/query.rs
+++ b/bee-network/bee-autopeering/src/discovery/query.rs
@@ -13,7 +13,7 @@ use crate::{
     task::Repeat,
 };
 
-use rand::{thread_rng, Rng};
+use rand::{thread_rng, Rng as _};
 
 #[derive(Clone)]
 pub(crate) struct QueryContext {
--- a/bee-network/bee-autopeering/src/init.rs
+++ b/bee-network/bee-autopeering/src/init.rs
@@ -162,16 +162,16 @@ where
 
     // Reverify old peers regularly.
     let f = query::reverify_fn();
-    let bootup_burst = 10.min(active_peers.read().len());
+    let bootstrap_verif_count = 10.min(active_peers.read().len());
     let delay = iter::repeat(Duration::from_millis(100))
-        .take(bootup_burst)
+        .take(bootstrap_verif_count)
         .chain(iter::repeat(DEFAULT_REVERIFY_INTERVAL));
     task_mngr.repeat(f, delay, ctx.clone(), "Reverification", MAX_SHUTDOWN_PRIORITY);
 
     // Discover new peers regularly.
     let f = query::query_fn();
-    let delay = iter::once(Duration::from_millis(2000))
-        .chain(iter::repeat(Duration::from_millis(0)).take(1))
+    let bootstrap_query_delay = 2;
+    let delay = iter::once(Duration::from_secs(bootstrap_query_delay))
         .chain(iter::repeat(DEFAULT_QUERY_INTERVAL));
     task_mngr.repeat(f, delay, ctx, "Discovery", MAX_SHUTDOWN_PRIORITY);
 
--- a/bee-network/bee-autopeering/src/init.rs
+++ b/bee-network/bee-autopeering/src/init.rs
@@ -7,7 +7,7 @@ use crate::{
     config::AutopeeringConfig,
     delay,
     discovery::{
-        manager::{DiscoveryManager, DiscoveryManagerConfig, DEFAULT_QUERY_INTERVAL, DEFAULT_REVERIFY_INTERVAL},
+        manager::{DiscoveryManager, DiscoveryManagerConfig, QUERY_INTERVAL_DEFAULT, REVERIFY_INTERVAL_DEFAULT},
         query::{self, QueryContext},
     },
     event::{self, EventRx},
@@ -28,11 +28,16 @@ use crate::{
     request::{self, RequestManager, EXPIRED_REQUEST_REMOVAL_INTERVAL},
     server::{server_chan, IncomingPacketSenders, Server, ServerConfig, ServerSocket},
     task::{TaskManager, MAX_SHUTDOWN_PRIORITY},
+    time::SECOND,
 };
 
 use std::{error, future::Future, iter, time::Duration};
 
 const NUM_TASKS: usize = 9;
+const BOOTSTRAP_MAX_VERIFICATIONS: usize = 10;
+const BOOTSTRAP_VERIFICATION_DELAY: Duration = Duration::from_millis(100);
+const BOOTSTRAP_QUERY_DELAY: Duration = Duration::from_secs(2 * SECOND);
+const BOOTSTRAP_UPDATE_DELAY: Duration = Duration::from_secs(4 * SECOND);
 
 /// Initializes the autopeering service.
 pub async fn init<S, I, Q, V>(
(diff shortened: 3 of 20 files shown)


bee-network/bee-autopeering/src/init.rs
```diff
@@ -162,12 +162,17 @@ where
 
     // Reverify old peers regularly.
     let f = query::reverify_fn();
-    let delay = iter::repeat(DEFAULT_REVERIFY_INTERVAL);
+    let bootstrap_verif_count = 10.min(active_peers.read().len());
```
thibault-martinez: There are a couple of magic numbers in this file, maybe constify them ?
10, 100, 2 and 4000

bee-network/bee-gossip/src/service/host.rs
```diff
@@ -198,7 +198,7 @@ async fn event_processor(shutdown: Shutdown, events: InternalEventReceiver, send
 async fn peerstate_checker(shutdown: Shutdown, senders: Senders, peerlist: PeerList) {
     debug!("Peer checker running.");
 
-    let Senders { internal_commands, .. } = senders;
+    // let Senders { internal_commands, .. } = senders;
```
thibault-martinez: Remove ?

bee-network/bee-autopeering/src/init.rs
Alex6323: not lying....I've read your comment after the fact :trollface:

bee-network/bee-autopeering/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bee-autopeering"
-version = "0.1.1"
+version = "0.1.2"
```
grtlr: @thibault-martinez I'm not sure if it applies here, but I thought we were always bumping minor?

bee-network/bee-autopeering/src/discovery/query.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
     task::Repeat,
 };
 
-use rand::{thread_rng, Rng};
+use rand::{thread_rng, Rng as _};
```
grtlr, line 16: Why do we need the rename here?

bee-network/bee-autopeering/src/init.rs
```diff
@@ -28,11 +28,16 @@ use crate::{
...
 };
 
-use std::{error, future::Future, iter};
+use std::{error, future::Future, iter, time::Duration};
 
 const NUM_TASKS: usize = 9;
+const BOOTSTRAP_MAX_VERIFICATIONS: usize = 10;
+const BOOTSTRAP_VERIFICATION_DELAY: Duration = Duration::from_millis(100);
+const BOOTSTRAP_QUERY_DELAY: Duration = Duration::from_secs(2 * SECOND);
+const BOOTSTRAP_UPDATE_DELAY: Duration = Duration::from_secs(4 * SECOND);
```
grtlr, line 40: This looks a bit weird, shouldn't `from_secs` be enough?

bee-network/bee-autopeering/src/init.rs
```diff
@@ -162,12 +167,14 @@ where
 
     // Reverify old peers regularly.
     let f = query::reverify_fn();
-    let delay = iter::repeat(DEFAULT_REVERIFY_INTERVAL);
+    let delay = iter::repeat(BOOTSTRAP_VERIFICATION_DELAY)
+        .take(BOOTSTRAP_MAX_VERIFICATIONS.min(active_peers.read().len()))
+        .chain(iter::repeat(REVERIFY_INTERVAL_DEFAULT));
```
grtlr, line 172: This took some time to wrap my head around 🙃 , but I wouldn't know a cleaner way to do it 👍.

bee-network/bee-autopeering/src/init.rs
```diff
@@ -181,7 +188,7 @@ where
 
     // Update the outbound neighborhood regularly (interval depends on whether slots available or not).
     let f = update::update_outbound_neighborhood_fn();
-    let delay = delay::ManualDelayFactory::new(OPEN_OUTBOUND_NBH_UPDATE_SECS);
+    let delay = iter::once(BOOTSTRAP_UPDATE_DELAY).chain(delay::ManualDelayFactory::new(OPEN_OUTBOUND_NBH_UPDATE_SECS));
```
grtlr, line 191: The more I think about it, the more I like this approach.

thibault-martinez (dismissed):

bee-network/bee-gossip/src/init.rs
```diff
@@ -190,8 +190,10 @@ fn init(
         .send(Event::LocalIdCreated { local_id })
         .map_err(|_| Error::LocalIdAnnouncementFailed)?;
 
+    // TODO: rename to PeerStateMap.
```
grtlr, line 193: Reminder ✏️

grtlr (commented):
I have to admit, I don't understand the intrinsics of many of these steps, but the code looks good 👍

bee-network/bee-autopeering/Cargo.toml
thibault-martinez: For consistency it would be best to only bump minor yeah. But since the previous versions of this weren't published this shouldn't be an issue.

bee-network/bee-autopeering/src/init.rs
Alex6323: I know it looks weird in this case. In my own defense: things have changed a bit over time, because the base time unit originally was a millisecond IIRC, so the constant `SECOND` was set to `1000`. Then line 30 would look like:
`const BOOTSTRAP_QUERY_DELAY: Duration = Duration::from_millis(2 * SECOND)`, which I think is a tiny bit more
readable than: 
`const BOOTSTRAP_QUERY_DELAY: Duration = Duration::from_millis(2000)`. Analogously more readable:
`60 * SECOND` rather than `60000` (which actually is an interval autopeering makes use of, and the more readable the bigger intervals you have.
I can also not deny the fact, that this was inspired by the reference Go code I looked at a lot, so it might have had some (negative?) influence on me. haha :trollface:

thibault-martinez: IMO `Duration::from_millis(2 * SECOND)` is extremely confusing to read

thibault-martinez: And `Duration::from_secs(2 * SECOND)` a bit redundant

samuel-rufi (dismissed):

bee-network/bee-autopeering/src/init.rs
samuel-rufi: Agree with @thibault-martinez

bee-network/bee-autopeering/src/discovery/query.rs
Alex6323: haha, we had this discussion already a while back. My explanations for doing this are: 
(a) Because I know for sure, that I won't ever implement that trait in that module I am releasing my claim on that symbol, and whoever might change that module in the future and finding this symbol useful as a name is free to use it.
(b) consistency with the rest of this crate (because I did that quite a few times).
(c) I find it useful in terms of readability, because it's a statement to the reader.
(d) I can, because clippy doesn't complain about it :trollface:

bee-network/bee-autopeering/Cargo.toml
Alex6323: Isn't this PR literally patching `bee-autopeering`, so bumbing the patch version should apply. Bumping the minor is for backwards compatible feature additions. Major is for backwards incompatible changes. At least this is how I understood semantic versioning. But maybe I forgot or didn't read a certain discussion we might have had in the past. :thinking:

thibault-martinez: Cargo semver is different from the usual semver

thibault-martinez: It's preferred to bump the minor and not the patch in every case for crates < 1.0.0. Pretty much the whole ecosystem is following this rule.

bee-network/bee-autopeering/src/init.rs
Alex6323: yea, I can understand the confusion. Maybe I should add a comment explaining why that works. (the peers live in a ring, and sending a verification request moves the ring by 1, so `BOOTSTRAP_VERIFICATION_DELAY` will be used for the maximum of 10 different(!) peers. After that the normal reverification interval applies.

bee-network/bee-gossip/src/init.rs
Alex6323: Yes, that's something I gonna do with the rewrite of `bee-gossip` I am now working on. Different PR though :slightly_smiling_face:

bee-network/bee-autopeering/src/init.rs
Alex6323: I find `Duration::from_secs(86400)` not only confusing to read, but also error prone, when `Duration::from_secs(24 * HOUR)`, could be used instead (which undenyably is more readable and just another example of the same thing I did above).

thibault-martinez: Why is it error prone ? You ask for milliseconds and then give a raw number. It doesn't get any easier/clearer than that IMO

Alex6323: update your browser...you were to quick. I changed my post to make things a bit clearer

thibault-martinez: Yeah yeah I see the point, I didn't initially comment so I'm fine with it anyway, I just find it a bit convoluted for the very simple scenario you are actually having in this file.

bee-network/bee-autopeering/src/discovery/query.rs
grtlr: If clippy doesn't complain I won't complain either 😁

Adam-Gleave (dismissed):

bee-network/bee-autopeering/Cargo.toml
Alex6323: Done

thibault-martinez (approved):

PR #1006, opened 2022-01-24
Add unlock conditions

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-message/src/error.rs
+++ b/bee-message/src/error.rs
@@ -103,88 +103,88 @@ impl std::error::Error for Error {}
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Error::CryptoError(e) => write!(f, "cryptographic error: {}.", e),
+            Error::CryptoError(e) => write!(f, "cryptographic error: {}", e),
             Error::DuplicateSignatureUnlockBlock(index) => {
                 write!(f, "duplicate signature unlock block at index: {0}", index)
             }
-            Error::DuplicateUtxo(utxo) => write!(f, "duplicate UTXO {:?} in inputs.", utxo),
-            Error::FeatureBlocksNotUniqueSorted => write!(f, "feature blocks are not unique and/or sorted."),
+            Error::DuplicateUtxo(utxo) => write!(f, "duplicate UTXO {:?} in inputs", utxo),
+            Error::FeatureBlocksNotUniqueSorted => write!(f, "feature blocks are not unique and/or sorted"),
             Error::InputUnlockBlockCountMismatch {
                 input_count,
                 block_count,
             } => {
                 write!(
                     f,
-                    "input count and unlock block count mismatch: {} != {}.",
+                    "input count and unlock block count mismatch: {} != {}",
                     input_count, block_count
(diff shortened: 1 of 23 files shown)


bee-message/src/error.rs
```diff
@@ -190,6 +194,8 @@ impl fmt::Display for Error {
             Error::InvalidUnlockBlockNft(index) => {
                 write!(f, "invalid unlock block nft: {0}", index)
             }
+            Error::InvalidUnlockConditionCount(count) => write!(f, "invalid unlock condition count: {}.", count),
```
grtlr: Who let the _dots_ out? 🐶

bee-message/src/output/alias.rs
```diff
@@ -217,13 +234,31 @@ impl AliasOutput {
     ///
     #[inline(always)]
     pub fn state_controller(&self) -> &Address {
-        &self.state_controller
+        if let UnlockCondition::StateControllerAddress(address) = self
+            .unlock_conditions
+            .get(StateControllerAddressUnlockCondition::KIND)
+            .unwrap()
```
grtlr, line 229: These unwraps are safe too because they are part of every `AliasOutput`?

grtlr (changes requested):

bee-message/src/output/alias.rs
thibault-martinez: Yeah, I moved the comment up so that it's clearer it applies to the whole function

thibault-martinez: test

thibault-martinez: est2

bee-message/src/output/alias.rs
```diff
@@ -38,26 +38,15 @@ pub struct AliasOutputBuilder {
...
-    pub fn new(
-        amount: u64,
-        alias_id: AliasId,
-        state_controller: Address,
-        governor: Address,
-    ) -> Result<AliasOutputBuilder, Error> {
-        validate_controller(&state_controller, &alias_id)?;
-        validate_controller(&governor, &alias_id)?;
-
+    pub fn new(amount: u64, alias_id: AliasId) -> Result<AliasOutputBuilder, Error> {
```
grtlr, line 41: I liked the previous code better because the user was not able to create an invalid object. If you don't like them as part of `new`, you could also consider requiring the unlock conditions as arguments to `AliasOutputBuilder::finish`. That way these conditions also wouldn't be part of the interface in `AliasOutput::build`.

grtlr (approved):
A small comment that's only about personal preference, but nothing that would block this PR.

bee-message/src/output/alias.rs
thibault-martinez: > the user was not able to create an invalid object

It's still not possible, it's all part of the builder so you never have an invalid object. The role of a builder is that it can at any point be in an invalid state until the very end when it's converted to a valid object. I conceptually also preferred the old version but it's too confusing because unlock conditions would come from two different places (and that would be the same if we put the params in finish).

I tried many different options but every times it was way more complex than it should have.

samuel-rufi (approved):

Adam-Gleave (approved):

PR #1027, opened 2022-01-28
Some more clean and compliance

thibault-martinez:
## Change checklist


bee-message/src/output/native_token.rs
```diff
@@ -64,12 +64,11 @@ impl NativeTokens {
     /// Maximum possible number of different native tokens that can reside in one output.
     pub const COUNT_MAX: u16 = 256;
 
-    /// Creates a new `NativeTokens`.
+    /// Creates a new [`NativeTokens`].
```
grtlr, line 67:
```suggestion
    /// Creates new [`NativeToken`]s.
```

grtlr (approved):

bee-message/src/output/native_token.rs
thibault-martinez: But we are actually creating a `NativeTokens`, not `NativeToken`s

grtlr: The _a_ and the plural are throwing me off. How about:

```suggestion
    /// Creates a new [`NativeTokens`] object.
```

thibault-martinez: I can add it for now if you want but it wouldn't be very consistent with the rest, we never put `object` in other docs and it's not the only plural form...

pvdrz: image

bee-message/src/output/unlock_condition/timelock.rs
```diff
@@ -40,3 +48,39 @@ impl TimelockUnlockCondition {
         self.timestamp
     }
 }
+
```
pvdrz, line 51: I think you can derive this and use a function that receives `impl Deref<Target=u32>` in `verify_with`.

pvdrz (approved):

bee-message/src/output/unlock_condition/timelock.rs
thibault-martinez: The verification needs 2 fields

bee-message/src/output/native_token.rs
pvdrz: Jokes aside, i would use the term `value` instead of `object`.

bee-message/src/output/unlock_condition/timelock.rs
pvdrz: but the verification checks that each field is not zero. If you verified each field with a different function it would produce the same code.

thibault-martinez: It checks that they are not BOTH zero at the same time, one of them is allowed to be 0

pvdrz: ohhhhhhhhhhh damn, nevermind then

PR #1054, opened 2022-02-01
Fix peer manager deadlock

### LINKED ISSUE #985: Bee doesn't stop
## Bug description

If I stop my bee service not all plugins get stopped and I can wait for more than 15min but Bee still doesn't stop...

## Rust version

- Rust version: 1.57.0

## Bee version

- Bee version (version number, commit, or branch): 7fbbfabb

## Hardware specification

- Operating system: Ubuntu
- RAM: 4GB
- Cores: 3 
 
## Steps To reproduce the bug

Stop Bee

## Errors

[bee_logs.txt](https://github.com/iotaledger/bee/files/7892715/bee_logs.txt)

pvdrz:
# Description of change

Avoid nested read guard acquisition over the `PeerManager` lock in the same function which causes a deadlock. See https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html for more details. In particular:

> This lock uses a task-fair locking policy which avoids both reader and writer starvation. This means that readers trying to acquire the lock will block even if the lock is unlocked when there are writers waiting to acquire the lock. Because of this, attempts to recursively acquire a read lock within a single thread may result in a deadlock.

Also the `RwLock` field has been made private to keep all the ways of accessing this lock as methods of the struct that contains it.

Fixes https://github.com/iotaledger/bee/issues/985 

## Type of change

The node was run twice for 30 and 60 minutes and the deadlock did not happen

## Change checklist

### DIFF
--- a/bee-protocol/src/workers/peer/manager_res.rs
+++ b/bee-protocol/src/workers/peer/manager_res.rs
@@ -54,7 +54,7 @@ pub struct PeerManagerInner {
 
 #[derive(Default)]
 // TODO private
-pub struct PeerManager(pub(crate) RwLock<PeerManagerInner>);
+pub struct PeerManager(RwLock<PeerManagerInner>);
 
 impl PeerManager {
     pub(crate) fn new() -> Self {
@@ -94,6 +94,26 @@ impl PeerManager {
 
     pub(crate) fn for_each<F: Fn(&PeerId, &Peer)>(&self, f: F) {
         self.0.read().peers.iter().for_each(|(id, (peer, _))| f(id, peer));
+    }
+
+    pub(crate) fn foo(&self, counter: &mut usize, f: impl Fn(&PeerId, &Peer) -> bool) -> Option<PeerId> {
+        let guard = self.0.read();
+
+        for _ in 0..guard.keys.len() {
+            let peer_id = &guard.keys[*counter % guard.keys.len()];
+
+            *counter += 1;
+
+            if let Some((peer, _)) = guard.peers.get(peer_id) {
+                if f(peer_id, peer.as_ref()) {
+                    return Some(peer_id.clone());
+                }
+            }
+        }
+
+        drop(guard);
+
+        None
     }
 
     pub fn is_connected(&self, id: &PeerId) -> bool {
--- a/bee-protocol/src/workers/peer/manager_res.rs
+++ b/bee-protocol/src/workers/peer/manager_res.rs
@@ -48,12 +48,10 @@ type PeerTuple = (Arc<Peer>, Option<(GossipSender, oneshot::Sender<()>)>);
 pub struct PeerManagerInner {
     peers: HashMap<PeerId, PeerTuple>,
     // This is needed to ensure message distribution fairness as iterating over a HashMap is random.
-    // TODO private
-    pub(crate) keys: Vec<PeerId>,
+    keys: Vec<PeerId>,
 }
 
 #[derive(Default)]
-// TODO private
 pub struct PeerManager(RwLock<PeerManagerInner>);
 
 impl PeerManager {
--- a/bee-protocol/src/workers/peer/manager_res.rs
+++ b/bee-protocol/src/workers/peer/manager_res.rs
@@ -13,7 +13,14 @@ use futures::channel::oneshot;
 use log::debug;
 use parking_lot::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 
-use std::{collections::HashMap, convert::Infallible, sync::Arc};
+use std::{
+    collections::HashMap,
+    convert::Infallible,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+};
 
 pub struct PeerManagerResWorker {}
 
@@ -30,7 +37,7 @@ impl<N: Node> Worker<N> for PeerManagerResWorker {
 
     async fn stop(self, node: &mut N) -> Result<(), Self::Error> {
         if let Some(peer_manager) = node.remove_resource::<PeerManager>() {
-            for (_, (_, sender)) in peer_manager.0.into_inner().peers {
+            for (_, (_, sender)) in peer_manager.inner.into_inner().peers {
                 if let Some(sender) = sender {
                     // TODO: Should we handle this error?
                     let _ = sender.1.send(());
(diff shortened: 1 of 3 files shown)


bee-protocol/src/workers/peer/manager_res.rs
```diff
@@ -54,7 +54,7 @@ pub struct PeerManagerInner {
 
 #[derive(Default)]
 // TODO private
```
thibault-martinez: remove

samuel-rufi (dismissed):

bee-protocol/src/workers/peer/manager_res.rs
```diff
@@ -96,6 +94,26 @@ impl PeerManager {
         self.0.read().peers.iter().for_each(|(id, (peer, _))| f(id, peer));
     }
 
+    pub(crate) fn find_first_fairly(&self, counter: &mut usize, f: impl Fn(&Peer) -> bool) -> Option<PeerId> {
```
thibault-martinez: Do you think we can make the counter internal ? API would be cleaner.

thibault-martinez: Actually I think it should be internal, it's the only way to guarantee 100% fairness.

pvdrz: I will make it internal. I was thinking on removing the `first` as I think it is redundant. `Iterator::find` already stops on the first match

thibault-martinez: Yeah I'd remove the `first` as well

thibault-martinez: `fair_find` or ` find_fairly` ?

pvdrz: whatever you prefer. I'd do `find_fairy` :trollface:

thibault-martinez: Ahah, whatever, you choose

pvdrz: 🧚‍♂️

bee-protocol/src/workers/peer/manager_res.rs
```diff
@@ -48,64 +55,95 @@ type PeerTuple = (Arc<Peer>, Option<(GossipSender, oneshot::Sender<()>)>);
...
     peers: HashMap<PeerId, PeerTuple>,
     // This is needed to ensure message distribution fairness as iterating over a HashMap is random.
-    // TODO private
-    pub(crate) keys: Vec<PeerId>,
+    keys: Vec<PeerId>,
 }
 
 #[derive(Default)]
-// TODO private
-pub struct PeerManager(pub(crate) RwLock<PeerManagerInner>);
```
Adam-Gleave, line 57: This feels like it was always a bad idea :eyes:

Adam-Gleave (dismissed):

bee-protocol/src/workers/peer/manager_res.rs
Alex6323: I agree.

pvdrz: Yeah I think we should have some guidelines or compilation checks about locks because I believe that every time we had an issue with the node freezing or not shutting down correctly is because we are using some `RwLock` or `Mutex` in an unstructured way.

bee-protocol/src/workers/peer/manager_res.rs
```diff
@@ -48,64 +55,95 @@ type PeerTuple = (Arc<Peer>, Option<(GossipSender, oneshot::Sender<()>)>);
...
         lock.keys.retain(|peer_id| peer_id != id);
         lock.peers.remove(id)
     }
 
     pub(crate) fn for_each<F: Fn(&PeerId, &Peer)>(&self, f: F) {
-        self.0.read().peers.iter().for_each(|(id, (peer, _))| f(id, peer));
+        self.inner.read().peers.iter().for_each(|(id, (peer, _))| f(id, peer));
+    }
+
+    pub(crate) fn fair_find(&self, f: impl Fn(&Peer) -> bool) -> Option<PeerId> {
```
Alex6323, line 141: Not necessarily related to that PR, but I think a little prose to explain why this `fair_find` method is needed could be helpful. Could we use an `IndexMap` here?

Isn't a better method name: `select_peer_randomly` or something?

pvdrz: I think I'm not the right person to answer this as I didn't write the code that inspired this method. 

I also have a couple questions about the `Vec<PeerId>`. The iteration order in `HashMap::keys` is not deterministic between runs but it is consistent if the hashmap has not changed, why is this not enough?

Regarding the name. I favor `find` against `select` because of [`Iterator::find`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find). About the `randomly` instead of `fairly` I'd say this is not a random traverse, it is consistent if the hashmap does not change.

thibault-martinez: > why this fair_find method is needed could be helpful. Could we use an IndexMap here?

You would still need the method with IndexMap, this is unrelated. The methods ensures a cycle around the peers so that the distribution is fair.

> Isn't a better method name: select_peer_randomly or something?

It's not random though.

> but it is consistent if the hashmap has not changed, why is this not enough?

Precisely because the maps does change a lot.

pvdrz: @thibault-martinez but if the map changes a lot, the vec changes a lot so what difference does it make?

thibault-martinez: Sure but the order is not completely shuffled, it keeps being fair for remaining peers

pvdrz: but removing a key from a hashmap doesn't change the iteration order for the other keys.

Alex6323: Right, I now do better understand why the method was written the way it is, but - coming back to my original point - in order to reduce confusion I think the code is "odd" (with the atomic) enough to add a little bit of explanation. Instead of that rotating "pointer" (modulus counter), one could also go with a ring data structure here, which feels like it would better encode the situation we have here.

thibault-martinez: The point of the counter is to keep the state of where you stopped cycling for a future iteration. See Discord protocol, Christian will try something.

thibault-martinez (approved):

Alex6323 (approved):

Adam-Gleave (approved):

grtlr (approved):

PR #1063, opened 2022-02-02
Add Output amount check

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-message/src/output/mod.rs
+++ b/bee-message/src/output/mod.rs
@@ -44,6 +44,7 @@ pub use unlock_condition::{UnlockCondition, UnlockConditions};
 use crate::Error;
 
 use derive_more::From;
+use packable::bounded::BoundedU64;
 
 use core::ops::RangeInclusive;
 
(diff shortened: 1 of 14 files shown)


bee-message/src/output/mod.rs
```diff
@@ -56,6 +57,8 @@ pub const OUTPUT_INDEX_MAX: u16 = OUTPUT_COUNT_MAX - 1; // 126
 /// The range of valid indices of outputs of a transaction .
 pub const OUTPUT_INDEX_RANGE: RangeInclusive<u16> = 0..=OUTPUT_INDEX_MAX; // [0..126]
 
+pub(crate) type OutputAmount = BoundedU64<1, { u64::MAX }>;
```
Thoralf-M: Shouldn't this be the max IOTA supply?

samuel-rufi (dismissed):

bee-message/src/output/mod.rs
thibault-martinez: I would have said so as well but it doesn't seem to be in the TIP

samuel-rufi: hmm good point

Thoralf-M: In https://github.com/lzpap/tips/blob/tx-updates/tips/TIP-0020/tip-0020.md there is `The sum of all Amount fields must not exceed Max IOTA Supply.`, that implies that also a single output can't have more

thibault-martinez: Yeah I agree but it really needs to be in TIP-18, otherwise a potentially invalid Output is considered valid until it's put in a Transaction. I'll ask Levente to clarify.

Adam-Gleave (dismissed):

pvdrz (dismissed):

Thoralf-M (approved):

grtlr (approved):

samuel-rufi (approved):

PR #1068, opened 2022-02-02
Add NativeTokens amount check

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-api/bee-rest-api/src/types/dtos.rs
+++ b/bee-api/bee-rest-api/src/types/dtos.rs
@@ -887,7 +887,7 @@ impl TryFrom<&NativeTokenDto> for NativeToken {
     type Error = Error;
 
     fn try_from(value: &NativeTokenDto) -> Result<Self, Self::Error> {
-        Ok(Self::new(
+        Self::new(
             (&value.token_id).try_into()?,
             value
                 .amount
@@ -895,7 +895,7 @@ impl TryFrom<&NativeTokenDto> for NativeToken {
                 .parse::<U256>()
                 .map_err(|_| Error::InvalidField("amount"))?,
         )
-        .map_err(|_| Error::InvalidField("NativeTokens"))?)
+        .map_err(|_| Error::InvalidField("nativeTokens"))
     }
 }
 
--- a/bee-api/bee-rest-api/src/types/dtos.rs
+++ b/bee-api/bee-rest-api/src/types/dtos.rs
@@ -894,7 +894,8 @@ impl TryFrom<&NativeTokenDto> for NativeToken {
                 .0
                 .parse::<U256>()
                 .map_err(|_| Error::InvalidField("amount"))?,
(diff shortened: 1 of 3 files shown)


bee-api/bee-rest-api/src/types/dtos.rs
```diff
@@ -894,7 +894,8 @@ impl TryFrom<&NativeTokenDto> for NativeToken {
                 .0
                 .parse::<U256>()
                 .map_err(|_| Error::InvalidField("amount"))?,
-        ))
+        )
+        .map_err(|_| Error::InvalidField("NativeTokens"))?)
```
Alex6323: ocd trigger alarm: shouldn't "amount" in line 896 also be uppercase?

Thoralf-M: Also, wouldn't this and the token_id error never be returned because of `map_err(|_|`?

bee-message/src/output/native_token.rs
```diff
@@ -40,12 +44,21 @@ impl NativeToken {
     }
 }
 
+#[inline]
```
Alex6323, line 47: Are we sure about the explicit inlining? I still hear Josh screaming about such things :trollface:

thibault-martinez: The function is needed to avoid passing a closure to packable but realistically we want to avoid the call if possible. Also this is just `inline`, not `inline(always)` so the compiler will do whatever it wants

bee-api/bee-rest-api/src/types/dtos.rs
thibault-martinez: No but actually it's NativeTokens that should be nativeTokens, I'll change

bee-message/src/output/native_token.rs
Alex6323: yea, I noticed that...I am wondering now if there's a real difference between not adding the attribute and adding `inline` only? :thinking:

Thoralf-M (approved):

Alex6323 (approved):

bee-api/bee-rest-api/src/types/dtos.rs
thibault-martinez: > Also, wouldn't this and the token_id error never be returned because of map_err(|_|?

You are totally right and I told myself the same but that's how this whole crate is done, it definitely needs a refactoring

bee-message/src/output/native_token.rs
thibault-martinez: Ask Christian :trollface:

pvdrz: I've seen cases where the compiler doesn't inline short functions (just returning a field or calling a method of a field). Given that the function is only used inside the packable implementation I think `#[inline]` puts us in the safe side.

PR #1132, opened 2022-02-12
Fix ImmutableAliasAddressUnlockCondition

Thoralf-M:
# Description of change

Fix ImmutableAliasAddressUnlockCondition

## Type of change

### DIFF
--- a/bee-message/src/output/unlock_condition/immutable_alias_address.rs
+++ b/bee-message/src/output/unlock_condition/immutable_alias_address.rs
@@ -1,14 +1,17 @@
 // Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::address::{Address, AliasAddress};
+use crate::{
+    address::{Address, AliasAddress},
+    Error,
+};
 
 use derive_more::From;
 
 /// Defines the permanent [`AliasAddress`] that owns this output.
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, From, packable::Packable)]
 #[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
-pub struct ImmutableAliasAddressUnlockCondition(Address);
+pub struct ImmutableAliasAddressUnlockCondition(#[packable(verify_with = verify_alias_address)] Address);
 
 impl ImmutableAliasAddressUnlockCondition {
     /// The [`UnlockCondition`](crate::output::UnlockCondition) kind of an [`ImmutableAliasAddressUnlockCondition`].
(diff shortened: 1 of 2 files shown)


bingyanglin (dismissed):

bee-message/src/output/unlock_condition/immutable_alias_address.rs
```diff
@@ -1,14 +1,14 @@
...
-use crate::address::AliasAddress;
+use crate::address::{Address, AliasAddress};
 
 use derive_more::From;
 
 /// Defines the permanent [`AliasAddress`] that owns this output.
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, From, packable::Packable)]
 #[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
-pub struct ImmutableAliasAddressUnlockCondition(AliasAddress);
+pub struct ImmutableAliasAddressUnlockCondition(Address);
```
thibault-martinez: But then we need to add a check to Packable::unpack because it can now technically put any kind of address in an `ImmutableAliasAddressUnlockCondition`

thibault-martinez (changes requested):

bee-message/src/output/unlock_condition/immutable_alias_address.rs
Thoralf-M: Like this?
```Rust
impl Packable for ImmutableAliasAddressUnlockCondition {
    type UnpackError = Error;

    fn pack<P: Packer>(&self, packer: &mut P) -> Result<(), P::Error> {
        self.0.pack(packer)?;
        Ok(())
    }

    fn unpack<U: Unpacker, const VERIFY: bool>(
        unpacker: &mut U,
    ) -> Result<Self, UnpackError<Self::UnpackError, U::Error>> {
        let address = Address::unpack::<_, VERIFY>(unpacker)?;

        if VERIFY {
            if let Address::Alias(_) = address {
            } else {
                return Err(packable::error::UnpackError::Packable(Error::InvalidAddressKind(
                    address.kind(),
                )));
            }
        }

        Ok(Self(address))
    }
}

```

thibault-martinez: Yeah that or a function that you would pass to a packable `verify_with` attribute so that you can derive Packable

thibault-martinez (approved):

pvdrz (approved):

bee-message/src/output/unlock_condition/immutable_alias_address.rs
```diff
@@ -1,14 +1,17 @@
...
+    Error,
+};
 
 use derive_more::From;
 
 /// Defines the permanent [`AliasAddress`] that owns this output.
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, From, packable::Packable)]
 #[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
-pub struct ImmutableAliasAddressUnlockCondition(AliasAddress);
+pub struct ImmutableAliasAddressUnlockCondition(#[packable(verify_with = verify_alias_address)] Address);
```
samuel-rufi, line 14: If an `ImmutableAliasAddressUnlockCondition` needs an `AliasAddress` then maybe pass that instead of `Address`? More safety and less checks

samuel-rufi (changes requested):

samuel-rufi (approved):

PR #1139, opened 2022-02-14
Avoid unbounded `RwLock` guards in `PeerManager`

pvdrz:


grtlr (dismissed):

Adam-Gleave (dismissed):

bee-protocol/src/workers/peer/manager_res.rs
```diff
@@ -101,13 +101,18 @@ impl PeerManager {
...
     }
 
-    // TODO find a way to only return a ref to the peer.
-    pub fn get(&self, id: &PeerId) -> Option<impl std::ops::Deref<Target = PeerTuple> + '_> {
-        RwLockReadGuard::try_map(self.inner.read(), |map| map.get(id)).ok()
+    pub fn get_map<T>(&self, id: &PeerId, f: impl FnOnce(&PeerTuple) -> T) -> Option<T> {
+        let guard = self.inner.read();
+        let output = guard.get(id).map(f);
+        drop(guard);
+        output
```
Alex6323, line 108: I guess the body is so complicated because of lifetime issues. Does the lifetime of `guard` has to be extended due to it being dropped after `get`. I honestly hate this issue if that's the case here. Otherwise we should really be able to write `self.inner.read().get(id).map(f)` here, right?

bee-protocol/src/workers/peer/manager_res.rs
```diff
@@ -101,13 +101,18 @@ impl PeerManager {
...
     }
 
-    pub fn get_mut(&self, id: &PeerId) -> Option<impl std::ops::DerefMut<Target = PeerTuple> + '_> {
-        RwLockWriteGuard::try_map(self.inner.write(), |map| map.get_mut(id)).ok()
+    pub fn get_mut_map<T>(&self, id: &PeerId, f: impl FnOnce(&mut PeerTuple) -> T) -> Option<T> {
+        let mut guard = self.inner.write();
+        let output = guard.get_mut(id).map(f);
+        drop(guard);
+        output
```
Alex6323, line 115: same as above

bee-protocol/src/workers/peer/manager_res.rs
pvdrz: I just wrote it like that to avoid any misinterpretations. I guess I could write it in the shorter way but I like knowing that the guard is dropped just when we need it to.

samuel-rufi (dismissed):

Adam-Gleave (approved):

Alex6323 (approved):

grtlr (approved):

samuel-rufi (approved):

PR #1150, opened 2022-02-17
Support `protocolVersion` and `networkId` changes in submit_message()

samuel-rufi:


bee-api/bee-rest-api/src/endpoints/routes/api/v1/submit_message.rs
```diff
@@ -141,6 +152,28 @@ pub(crate) async fn submit_message<B: StorageBackend>(
     } else {
         let payload_dto = serde_json::from_value::<PayloadDto>(payload_v.clone())
             .map_err(|e| reject::custom(CustomRejection::BadRequest(e.to_string())))?;
+
+        // TODO: is this check really necessary here? this check should be done by the node message processing pipe?
```
grtlr: Do you plan on working on this in this PR? Otherwise we could open an issue.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
```diff
@@ -78,30 +81,38 @@ pub(crate) async fn submit_message<B: StorageBackend>(
     rest_api_config: RestApiConfig,
     protocol_config: ProtocolConfig,
 ) -> Result<impl Reply, Rejection> {
-    let network_id_v = &value["networkId"];
+    let protocol_version_v = &value["protocolVersion"];
```
Alex6323: what does `_v` stand for? Should this be obvious? If not, I'ld consider writing it out.

samuel-rufi: It refers to the incoming JSON value, I wrote it out as you suggested 👍

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
```diff
@@ -78,40 +81,39 @@ pub(crate) async fn submit_message<B: StorageBackend>(
...
+    if !protocol_version_json.is_null() {
+        let parsed = protocol_version_json.as_u64().ok_or_else(|| {
+            reject::custom(CustomRejection::BadRequest(
+                "invalid protocol version: expected an unsigned integer < 256".to_string(),
+            ))
+        })? as u8;
 
-    let parents: Vec<MessageId> = if parents_v.is_null() {
+        // TODO: is this check really necessary here considering that the "node processing pipe" anyways checks the
+        // field for incoming messages?
```
Alex6323: Maybe make this a NOTE rather than a TODO, because I feel like this could become something that keeps hanging around for all eternity. You could argue that this additional check is probably not necessary due to the processing pipe, but to make the crate more self-contained/self-sufficient (not relying on other crates to perform the check) it is performed anyway.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
```diff
@@ -122,32 +124,55 @@ pub(crate) async fn submit_message<B: StorageBackend>(
...
+    let payload = if payload_json.is_null() {
         None
     } else {
-        let payload_dto = serde_json::from_value::<PayloadDto>(payload_v.clone())
+        let payload_dto = serde_json::from_value::<PayloadDto>(payload_json.clone())
             .map_err(|e| reject::custom(CustomRejection::BadRequest(e.to_string())))?;
+
+        // TODO: is this check really necessary here considering that the "node processing pipe" anyways checks the
+        // field for incoming messages? In case of a transaction payload, check if its `network_id` matches the
+        // one configured in the node.
```
Alex6323: same here

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
```diff
@@ -78,40 +81,39 @@ pub(crate) async fn submit_message<B: StorageBackend>(
...
-            })?
-            .parse::<u64>()
-            .map_err(|_| {
-                reject::custom(CustomRejection::BadRequest(
-                    "invalid network id: expected an u64-string".to_string(),
-                ))
-            })?
-    };
+    if !protocol_version_json.is_null() {
+        let parsed = protocol_version_json.as_u64().ok_or_else(|| {
```
Alex6323: I suggest renaming to `parsed_version` or even `parsed_protocol_version` to increase readability of the below code without having to come back to this line 93 to understand what `parsed` actually contains. Ignore that advise if you think this is too obvious, because admittedly I am not too familiar with this crate :slightly_smiling_face:

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
```diff
@@ -122,32 +124,55 @@ pub(crate) async fn submit_message<B: StorageBackend>(
...
+
         Some(Payload::try_from(&payload_dto).map_err(|e| reject::custom(CustomRejection::BadRequest(e.to_string())))?)
     };
 
-    let nonce = if nonce_v.is_null() {
+    let nonce = if nonce_json.is_null() {
         None
     } else {
-        let parsed = nonce_v
+        let parsed = nonce_json
```
Alex6323: `parsed` ~> `parsed_nonce` maybe?

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
samuel-rufi: Thanks for the input.
Your comment makes me think that we actually don't want to have these checks in here. The crate shouldn't assume things about the node or network imo. It should simply take the users JSON, deserialize it and submit it to the node processing pipe.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
```diff
@@ -78,40 +81,39 @@ pub(crate) async fn submit_message<B: StorageBackend>(
...
+
+    let parents: Vec<MessageId> = if parents_json.is_null() {
         tangle.get_messages_to_approve().await.ok_or_else(|| {
             reject::custom(CustomRejection::ServiceUnavailable(
                 "can not auto-fill parents: no tips available".to_string(),
             ))
         })?
     } else {
-        let array = parents_v.as_array().ok_or_else(|| {
+        let array = parents_json.as_array().ok_or_else(|| {
```
Alex6323: Do you have something more descriptive than just `array`? From this line it looks like `array` could be renamed to `parents` but I am not sure.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
samuel-rufi: One argument for doing the checks in the API would be to not do PoW if these fields are wrong. However in case the user sends message raw bytes, there is no PoW to do as it's already done and the message gets directly injected in the node processing pipe. So this argument only holds for JSON.


Arguments against would be "double checks" and that the API is doing things it doesn't really need to do. Also consider in future however there might be more checks to do, and maybe we would need to check even different payload types. And we need the checks for JSON + RAW Byte inputs.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
Alex6323: and in the loop below I would rename: `s` ~> `message_id`. It's parsed into an actual `MessageId` type, right?, so I would just shadow variables here.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
Alex6323: It's your call :slightly_smiling_face:

thibault-martinez: I think the protocol version and network ID are different regarding this. If the protocol version is wrong, the JSON deserialization will most probably fail so we should even try. Network ID doesn't influence deserialization.

rufsam: >  If the protocol version is wrong, the JSON deserialization will most probably fail
Why is the JSON deserialization failing? Not sure I follow

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
samuel-rufi: Done 👍

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
samuel-rufi: Done

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
samuel-rufi: Done

bee-api/bee-rest-api/src/endpoints/routes/api/v2/submit_message.rs
samuel-rufi: Removed the network check as we will rely on the node pipe for this

thibault-martinez (approved):

grtlr (approved):

Adam-Gleave (approved):

PR #1171, opened 2022-02-24
bee-autopeering: fix ambiguous method names

Alex6323:


bee-network/bee-autopeering/src/peering/neighbor.rs
```diff
@@ -199,8 +199,10 @@ impl<const N: usize, const INBOUND: bool> NeighborhoodInner<N, INBOUND> {
         }
     }
 
-    fn is_preferred(&mut self, candidate: &Neighbor) -> bool {
-        if let Some(furthest) = self.find_furthest_if_full() {
+    /// Note: If the neighborhood is not full, than any `candidate` is just good enough, and will be preferred
```
thibault-martinez:
```suggestion
    /// Note: If the neighborhood is not full, then any `candidate` is just good enough and will be preferred
```

Alex6323: Done

PR #1176, opened 2022-02-25
Check storage deposit for `Output`s

### LINKED ISSUE #1104: Add syntactic `ByteCost` checks

grtlr:
# Description of change

Dust protection :sparkles:.

Closes #1104.

## Type of change

At some point we should think about how to the whole storage deposit logic. @thibault-martinez any ideas on that?

## Change checklist


pvdrz (dismissed):

pvdrz (dismissed):

pvdrz (approved):

pvdrz (dismissed):

bee-message/src/error.rs
```diff
@@ -121,6 +126,7 @@ pub enum Error {
     MissingPayload,
     MissingRequiredSenderBlock,
     MissingStateControllerUnlockCondition,
+    MissingStorageDepositReturnUnlockCondition,
```
Thoralf-M: That can't be an error, either you have one or not,  but since it's optional it can't be missing

grtlr: Hmm, what about the case where the output amount is too small, but no `StorageDepositReturnUnlockCondition` is posted? Should we give it a different name, or unify under the `InvalidStorageDepositReturnAmount` error?

grtlr: I think it would be nice to differentiate between these two cases though.

Thoralf-M: I think it should be `InvalidStorageDepositReturnAmount`, otherwise you would always get `StorageDepositReturnUnlockCondition` as error unless you have one? I think in most output you don't expect that unlock condition

bee-message/src/output/mod.rs
```diff
@@ -199,6 +202,47 @@ impl Output {
             Self::Nft(output) => Some(output.immutable_feature_blocks()),
         }
     }
+
+    /// Checks if a sufficient storage deposit was made for the given [`Output`].
+    pub fn check_sufficient_storage_deposit(&self, config: &ByteCostConfig) -> Result<(), Error> {
+        let maybe_storage_deposit_condition = match self {
```
thibault-martinez: If you rebase to include my recent PR, you can call `unlock_conditions` directly on the output and then map `storage_deposit_return` :)

bee-message/src/error.rs
grtlr: If introduced two new variants `InsufficientStorageDepositAmount` and `InsufficientStorageDepositReturnAmount` because `InvalidStorageDepositAmount` is used by `Packable` to denote out-of-bounds errors. I'm not settled on the names yet, so if you have suggestions... 😉

bee-message/src/output/mod.rs
grtlr: I need to check for the type of `Output` to return early for types that don't require storage deposits, or can I assume that `check_sufficient_storage_deposit` will only be called in a context where this is not important?

bee-message/src/output/mod.rs
```diff
@@ -199,6 +202,50 @@ impl Output {
...
+                .unlock_conditions()
+                .get(StorageDepositReturnUnlockCondition::KIND),
+        };
+
+        let required = minimum_storage_deposit(config, self);
+
+        if let Some(UnlockCondition::StorageDepositReturn(storage_deposit_condition)) = maybe_storage_deposit_condition
+        {
+            let deposit = storage_deposit_condition.amount();
+            if deposit >= required {
```
Thoralf-M: Why not
```suggestion
            if storage_deposit_condition.amount() >= required {
```
? Then it's similar to below

bee-message/src/error.rs
Thoralf-M: I think they're fine

bee-message/src/output/mod.rs
```diff
@@ -201,6 +204,52 @@ impl Output {
...
+                return Ok(());
+            } else {
+                return Err(Error::InsufficientStorageDepositReturnAmount {
+                    required,
+                    deposit: storage_deposit_condition.amount(),
+                });
+            }
+        }
+
+        if self.amount() >= required {
```
Thoralf-M: Can we check the amount first and only if that passes check StorageDepositReturn? I think the amount in the output is more "basic" and should be returned as error if both are wrong
Same would be then in the match in processor.rs

bee-message/src/output/mod.rs
```diff
@@ -201,6 +204,43 @@ impl Output {
...
+                None
+            };
+
+        if self.amount() >= required {
+            match maybe_deposit {
+                Some(deposit) if deposit >= required => Ok(()),
+                Some(deposit) => Err(Error::InsufficientStorageDepositReturnAmount { required, deposit }),
+                None => Ok(()),
+            }
+        } else {
```
Thoralf-M: Sorry if I maybe explained it wrong, but the amount always needs to be >= required, independent of `StorageDepositReturn` UnlockCondition
So in this case we always need to return 
```Rust
Err(Error::InsufficientStorageDepositAmount {
           required,
           amount: self.amount(),
})
```
And then additionally we need to check if there is such an unlock condition
```Rust
if let Some(UnlockCondition::StorageDepositReturn(storage_deposit_condition)) = maybe_deposit_condition {
    if storage_deposit_condition.amount() < required {
        return Err(Error::InsufficientStorageDepositReturnAmount { required, storage_deposit_condition.amount() })
    }
}
```

bee-message/src/error.rs
```diff
@@ -188,6 +197,18 @@ impl fmt::Display for Error {
             Error::InvalidStorageDepositAmount(amount) => {
                 write!(f, "invalid storage deposit amount: {}", amount)
             }
+            Error::InsufficientStorageDepositAmount { required, amount } => {
```
Thoralf-M: How about changing the order of `required, amount`?

bee-message/src/output/mod.rs
Thoralf-M: https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#additional-syntactic-transaction-validation-rule
This also needs to be checked if there is a StorageDepositReturn

grtlr: https://github.com/iotaledger/tips/pull/38#discussion_r818507694

Thoralf-M: So you need to add the check where you compare the amount in the StorageDepositReturn with the calculated storage deposit for a basic output with the address type from the StorageDepositReturn

bee-message/src/output/mod.rs
```diff
@@ -201,10 +204,60 @@ impl Output {
...
+            Output::Treasury(_) => return Ok(()), // `TreasuryOutput`s don't have `UnlockConditions`.
+            Output::Basic(_) | Output::Alias(_) | Output::Foundry(_) | Output::Nft(_) => self
+                .unlock_conditions()
+                .and_then(|conds| conds.get(StorageDepositReturnUnlockCondition::KIND)),
+        };
+
+        if let Some(UnlockCondition::StorageDepositReturn(return_condition)) = maybe_return_condition {
+            let required_deposit = minimum_storage_deposit(config, return_condition.return_address());
+            let minimum = self.amount() - required_deposit;
+            let maximum = self.amount();
```
Thoralf-M: There is no maximum (unless you take the IOTA supply as maximum)

Edit: No maximum for all StorageDepositReturn UnlockConditions, because it always depends on the amount of the output

bee-message/src/output/mod.rs
```diff
@@ -201,10 +204,60 @@ impl Output {
...
+            if !(minimum..=maximum).contains(&return_condition.amount()) {
+                return Err(Error::InvalidStorageDepositReturnAmount {
+                    deposit: return_condition.amount(),
+                    minimum,
+                    maximum,
+                });
+            }
+        };
```
thibault-martinez: I would rework these 2 blocks into something like
```rust
        if let Some(return_condition) = self
            .unlock_conditions()
            .and_then(|conds| conds.storage_deposit_return())
        {
            // Do stuff
        } else {
            return Ok(());
        }
```

bee-message/src/output/mod.rs
```diff
@@ -201,10 +204,71 @@ impl Output {
...
+                    deposit: return_condition.amount(),
+                    required: minimum_deposit,
+                });
+            }
+
+            // Check if the storage deposit return was required in the first place.
+            // `Amount` - `Return Amount` ≤ `Required Storage Deposit of the Output`
+            if self.amount() - return_condition.amount() > required_output {
+                return Err(Error::UnnecessaryStorageDepositReturnCondition {
+                    remainder: self.amount() - return_condition.amount(),
```
Thoralf-M: I don't like the name `remainder` here, because this will be the amount the receiver gets and remainder would rather be the amount in the `StorageDepositReturn`

grtlr: How about the `logical_amount` of an `Output`, instead of `remainder`?

Thoralf-M: Would be fine for me

Thoralf-M (dismissed):

Thoralf-M (dismissed):

pvdrz (dismissed):

Thoralf-M (dismissed):

bee-message/src/output/mod.rs
```diff
@@ -53,13 +54,15 @@ pub(crate) use treasury::TreasuryOutputAmount;
...
 
-use crate::{constant::IOTA_SUPPLY, Error};
+use crate::{address::Address, constant::IOTA_SUPPLY, Error};
 
 use derive_more::From;
 use packable::{bounded::BoundedU64, PackableExt};
 
 use core::ops::RangeInclusive;
 
+use self::unlock_condition::AddressUnlockCondition;
```
thibault-martinez: Can you put this with the other mod levels imports please ?

bee-message/src/output/mod.rs
```diff
@@ -201,10 +204,71 @@ impl Output {
             Self::Nft(output) => Some(output.immutable_feature_blocks()),
         }
     }
+
+    /// Checks if a sufficient storage deposit was made for the given [`Output`].
+    pub fn check_sufficient_storage_deposit(&self, config: &ByteCostConfig) -> Result<(), Error> {
```
thibault-martinez: I would call this simply `check_storage_deposit` as given the errors it checks a bit more than just sufficiency. For a bit more consistency with other verification methods I would even call it `verify_storage_deposit`.

bee-protocol/src/workers/message/processor.rs
```diff
@@ -137,13 +144,59 @@ where
...
-                                continue;
+                                continue 'next_event;
+                            }
+
+                            for (i, output) in essence.outputs().iter().enumerate() {
+                                if let Err(error) = output.check_sufficient_storage_deposit(&config.byte_cost) {
+                                    match error {
+                                        MessageError::InsufficientStorageDepositAmount{required, amount} => {
+                                            notify_invalid_message(
+                                                format!("Insufficient amount in output {i}: {amount} (should be at least {required})"),
```
thibault-martinez: Why not let the error format itself ? You already wrote the display impl.

bee-message/src/output/byte_cost.rs
```diff
@@ -99,7 +104,48 @@ impl<T: ByteCost, const N: usize> ByteCost for [T; N] {
...
 }
 
-/// Computes the storage cost for `[crate::output::Output]`s.
-pub fn minimum_storage_deposit(config: &ByteCostConfig, output: &impl ByteCost) -> u64 {
-    config.v_byte_cost * output.weighted_bytes(config) + config.v_byte_offset
+#[cfg(test)]
+mod test {
+    use crate::output::{ByteCost, ByteCostConfig, ByteCostConfigBuilder, Output};
+
+    use bee_test::rand::output::{rand_alias_output, rand_basic_output, rand_foundry_output, rand_nft_output};
```
thibault-martinez, line 111:
```suggestion
    use bee_test::rand::output::{rand_alias_output, rand_basic_output, rand_foundry_output, rand_nft_output};

```

bee-message/src/output/mod.rs
```diff
@@ -201,10 +204,71 @@ impl Output {
             Self::Nft(output) => Some(output.immutable_feature_blocks()),
         }
     }
+
+    /// Checks if a sufficient storage deposit was made for the given [`Output`].
```
thibault-martinez: Maybe adapt the comment to what it actually does

bee-message/src/output/mod.rs
```diff
@@ -201,10 +204,71 @@ impl Output {
...
+    pub fn check_sufficient_storage_deposit(&self, config: &ByteCostConfig) -> Result<(), Error> {
+        let required_output = self.byte_cost(config);
```
thibault-martinez: Name is a bit weird IMO

thibault-martinez: Reopening because this was not addressed, just checking if it's a mistake or not. If you like this name just close again and ignore me.

thibault-martinez (dismissed):

thibault-martinez (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

Thoralf-M (dismissed):

thibault-martinez (dismissed):

thibault-martinez (approved):

pvdrz (approved):

Thoralf-M (approved):

PR #1187, opened 2022-02-28
Fix metadata update race condition

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-protocol/src/workers/propagator.rs
+++ b/bee-protocol/src/workers/propagator.rs
@@ -49,9 +49,9 @@ async fn propagate<B: StorageBackend>(
             // Note: There are two types of solidity carriers:
             // * solid messages (with available history)
             // * solid entry points / sep (with verified history)
-            // Because we know both parents are solid, we also know that they have set OMRSI/YMRSI values, hence
-            // we can simply unwrap. We also try to minimise unnecessary Tangle API calls if - for example - the
-            // parent in question turns out to be a SEP.
+            // Because we know all parents are solid, we also know that they have set OMRSI/YMRSI values, hence we can
+            // simply unwrap. We also try to minimise unnecessary Tangle API calls if - for example - the parent in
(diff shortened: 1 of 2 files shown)


grtlr (approved):

bee-protocol/src/workers/propagator.rs
```diff
@@ -62,14 +62,11 @@ async fn propagate<B: StorageBackend>(
                     .await
                 {
                     Some(parent_sepi) => (IndexId::new(parent_sepi, *parent), IndexId::new(parent_sepi, *parent)),
-                    None => match tangle.get_metadata(parent).await {
-                        // 'unwrap' is safe (see explanation above)
-                        Some(parent_md) => (
-                            parent_md.omrsi().expect("solid msg with unset omrsi"),
-                            parent_md.ymrsi().expect("solid msg with unset ymrsi"),
-                        ),
-                        None => continue 'outer,
```
grtlr, line 71: Is the new code really equivalent to the older version? It seems like the old code was skipping later statements due to `continue 'outer`.

grtlr (changes requested):

bee-protocol/src/workers/propagator.rs
thibault-martinez: It is not equivalent but it was really weird. Failing to load the metadata would have been silent after discovering that a message should be set solid but before actually marking it as solid; hence losing the flag forever and potentially stopping solidification as well. But this should never happen anyway.

grtlr (dismissed):

samuel-rufi (dismissed):

pvdrz (dismissed):

pvdrz (approved):

samuel-rufi (approved):

grtlr (approved):

Adam-Gleave (approved):

PR #1188, opened 2022-02-28
Use sync locks in `bee-tangle`

pvdrz:
I still have to test this code for any performance issues

### DIFF
--- a/bee-tangle/src/vertices.rs
+++ b/bee-tangle/src/vertices.rs
@@ -6,8 +6,8 @@ use crate::vertex::Vertex;
 use bee_message::MessageId;
 
 use hashbrown::{hash_map::DefaultHashBuilder, raw::RawTable};
+use parking_lot::RwLock;
 use rand::Rng;
-use tokio::sync::RwLock;
 
 use std::{
     hash::{BuildHasher, Hash, Hasher},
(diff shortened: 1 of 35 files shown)


bee-tangle/src/vertices.rs
```diff
@@ -54,36 +54,58 @@ impl Vertices {
...
-        let table = self.get_table(hash).read().await;
 
-        RwLockReadGuard::try_map(table, |table| match table.get(hash, equivalent_id(message_id)) {
-            Some((_, v)) => Some(v),
-            None => None,
-        })
-        .ok()
+        let guard = self.get_table(hash).read();
+        let output = guard.get(hash, equivalent_id(message_id)).map(|(_, v)| f(v));
+        drop(guard);
```
grtlr, line 62: Does the `drop(guard)` influence the performance compared to just dropping the lock when the `guard` goes out of scope? If not, I'm in favor of keeping omitting the `drop()` as it can lead to errors. I might even go as far as to provide a reason in a comment for why manually dropping a lock is safe, similar to the comments that we write for `.unwrap()`.

grtlr (changes requested):

bee-tangle/src/vertices.rs
pvdrz: I think it actually does not matter but i prefer to have the explicitness of knowing that the guard is dropped just after computing the output.

grtlr: Yeah, I can see that point. I guess it's a matter of preference 😄.

grtlr (approved):

pvdrz:
This still deadlocks and sync locks performance oscillates a lot. Closing until we come up with a better idea of what's causing this behavior

Issue #1144, opened 2022-02-17
Build dashboard in `build.rs`


grtlr:
Waiting on: #1056

grtlr:
The general idea would be that the dashboard provides a precompiled version of the project. But I will wait until the plugin situation is resolved.

PR #1192, opened 2022-03-01
Fetch dashboard release in `build.rs`

Adam-Gleave:
Addresses #1144.

Adds to the `build.rs` script in `bee-node` to fetch the latest dashboard release from the [node-dashboard repository](https://github.com/iotaledger/node-dashboard), and place it in a subdirectory of `OUT_DIR` (defined by cargo at build-time) so that it can be used by `rust-embed`.

This will automatically run if `OUT_DIR/dashboard` fails to exist, or if `FETCH_DASHBOARD=1` is set.

Cargo recommends that build scripts *only* write any output files to this directory. Unfortunately that does mean that the files will hang around when the package is rebuilt and the build hash changes, so we might need to `cargo clean` more than normal...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ members = [
 	"bee-ternary",
 	"bee-test",
 ]
+resolver = "2"
 
 [profile.dev]
 panic = "abort"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ members = [
 	"bee-ternary",
 	"bee-test",
 ]
-resolver = "2"
 
 [profile.dev]
 panic = "abort"
--- a/bee-node/Cargo.toml
+++ b/bee-node/Cargo.toml
@@ -56,6 +56,14 @@ tracing = { version = "0.1.29", default-features = false, optional = true }
 warp = { version = "0.3.1", default-features = false }
 warp-reverse-proxy = { version = "0.4.0", default-features = false, optional = true }
 
+[build-dependencies]
+reqwest = { version = "0.11.5", default-features = false, features = [ "default-tls", "json" ] }
+serde = { version = "1.0.130", default-features = false, features = [ "derive" ] }
+serde_json = { version = "1.0.68", default-features = false }
+tempfile = { version = "3.3.0", default-features = false }
+tokio = { version = "1.12.0", default-features = false, features = [ "signal", "rt", "macros", "rt-multi-thread" ] }
+zip = { version = "0.5.13", default-features = false, features = [ "bzip2", "deflate" ] }
+
 [lib]
 name = "bee_node"
 path = "src/lib.rs"
--- a/bee-node/Cargo.toml
+++ b/bee-node/Cargo.toml
@@ -73,6 +73,6 @@ path = "src/main.rs"
 [features]
 default = [ "rocksdb" ]
 
-dashboard = [ "cap", "mime_guess", "reqwest", "rust-embed", "sha2", "tempfile", "warp-reverse-proxy", "zip" ]
+dashboard = [ "cap", "mime_guess", "reqwest", "rust-embed", "serde_repr", "sha2", "tempfile", "warp-reverse-proxy", "zip" ]
 rocksdb = [ "bee-storage-rocksdb" ]
 sled = [ "bee-storage-sled" ]
--- a/bee-node/Cargo.toml
+++ b/bee-node/Cargo.toml
@@ -57,8 +57,8 @@ warp = { version = "0.3.1", default-features = false }
 warp-reverse-proxy = { version = "0.4.0", default-features = false, optional = true }
 
 [build-dependencies]
-checksums = { version = "0.9.1", default-features = false, optional = true }
 reqwest = { version = "0.11.5", default-features = false, features = [ "blocking", "default-tls", "json" ], optional = true }
+sha2 = { version = "0.9.6", default-features = false, optional = true }
 tempfile = { version = "3.3.0", default-features = false, optional = true }
 zip = { version = "0.5.13", default-features = false, features = [ "bzip2", "deflate" ], optional = true }
 
@@ -73,6 +73,6 @@ path = "src/main.rs"
 [features]
 default = [ "rocksdb" ]
 
-dashboard = [ "cap", "checksums", "mime_guess", "reqwest", "rust-embed", "serde_repr", "tempfile", "warp-reverse-proxy", "zip" ]
+dashboard = [ "cap", "mime_guess", "reqwest", "rust-embed", "serde_repr", "sha2", "tempfile", "warp-reverse-proxy", "zip" ]
 rocksdb = [ "bee-storage-rocksdb" ]
 sled = [ "bee-storage-sled" ]
--- a/bee-node/Cargo.toml
+++ b/bee-node/Cargo.toml
@@ -57,11 +57,10 @@ warp = { version = "0.3.1", default-features = false }
 warp-reverse-proxy = { version = "0.4.0", default-features = false, optional = true }
 
 [build-dependencies]
-reqwest = { version = "0.11.5", default-features = false, features = [ "default-tls", "json" ], optional = true }
+reqwest = { version = "0.11.5", default-features = false, features = [ "blocking", "default-tls", "json" ], optional = true }
 serde = { version = "1.0.130", default-features = false, features = [ "derive" ], optional = true }
 serde_json = { version = "1.0.68", default-features = false, optional = true }
 tempfile = { version = "3.3.0", default-features = false, optional = true }
-tokio = { version = "1.12.0", default-features = false, features = [ "signal", "rt", "macros", "rt-multi-thread" ], optional = true }
 zip = { version = "0.5.13", default-features = false, features = [ "bzip2", "deflate" ], optional = true }
 
 [lib]
@@ -75,6 +74,6 @@ path = "src/main.rs"
 [features]
 default = [ "rocksdb" ]
 
-dashboard = [ "cap", "mime_guess", "reqwest", "rust-embed", "serde", "serde_repr", "serde_json", "tempfile", "tokio", "warp-reverse-proxy", "zip" ]
+dashboard = [ "cap", "mime_guess", "reqwest", "rust-embed", "serde", "serde_repr", "serde_json", "tempfile", "warp-reverse-proxy", "zip" ]
 rocksdb = [ "bee-storage-rocksdb" ]
 sled = [ "bee-storage-sled" ]
--- a/bee-node/Cargo.toml
+++ b/bee-node/Cargo.toml
@@ -57,12 +57,12 @@ warp = { version = "0.3.1", default-features = false }
 warp-reverse-proxy = { version = "0.4.0", default-features = false, optional = true }
 
 [build-dependencies]
-reqwest = { version = "0.11.5", default-features = false, features = [ "default-tls", "json" ] }
-serde = { version = "1.0.130", default-features = false, features = [ "derive" ] }
-serde_json = { version = "1.0.68", default-features = false }
-tempfile = { version = "3.3.0", default-features = false }
-tokio = { version = "1.12.0", default-features = false, features = [ "signal", "rt", "macros", "rt-multi-thread" ] }
-zip = { version = "0.5.13", default-features = false, features = [ "bzip2", "deflate" ] }
+reqwest = { version = "0.11.5", default-features = false, features = [ "default-tls", "json" ], optional = true }
+serde = { version = "1.0.130", default-features = false, features = [ "derive" ], optional = true }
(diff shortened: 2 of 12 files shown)


Adam-Gleave:
Do we actually want this in a build script specifically? It feels a bit odd to be downloading things within a build script, and also means that we could possibly establish a permanent directory for the dashboard files to be placed, rather than in every `OUT_DIR` of each package build. All that `rust-embed` needs to find the file is an environment variable. 

This is definitely something that we could do in an external script, but we would need to maintain scripts for each platform: a cargo build script at least makes this cross-platform.

thibault-martinez (changes requested):
You need to remove the npm stuff from
- bee-node/README.md
- documentation/docs/setup_a_node.mdx

grtlr:
I don't think downloading things in the `build.rs` script is bad, and a lot of crates do that. For example [here](https://github.com/rust-rocksdb/rust-rocksdb/blob/180e65e1b00f6b032e1513d94a4f0a4de7cafba1/librocksdb-sys/build.rs#L299) `librocksdb-sys` calls `git submodule init`. We just need to make sure to set `cargo-rerun` in a way that it doesn't always trigger a rebuild.

bee-node/Cargo.toml
```diff
@@ -56,6 +56,14 @@ tracing = { version = "0.1.29", default-features = false, optional = true }
 warp = { version = "0.3.1", default-features = false }
 warp-reverse-proxy = { version = "0.4.0", default-features = false, optional = true }
 
+[build-dependencies]
+reqwest = { version = "0.11.5", default-features = false, features = [ "default-tls", "json" ], optional = true }
+serde = { version = "1.0.130", default-features = false, features = [ "derive" ], optional = true }
+serde_json = { version = "1.0.68", default-features = false, optional = true }
+tempfile = { version = "3.3.0", default-features = false, optional = true }
+tokio = { version = "1.12.0", default-features = false, features = [ "signal", "rt", "macros", "rt-multi-thread" ], optional = true }
+zip = { version = "0.5.13", default-features = false, features = [ "bzip2", "deflate" ], optional = true }
```
grtlr: These will mostly move to the new dashboard plugin crate in the future, right?

bee-node/README.md
```diff
@@ -9,22 +9,22 @@
 ```sh
 apt-get update
 apt-get upgrade
-apt-get install git npm build-essential cmake pkg-config librocksdb-dev llvm clang libclang-dev libssl-dev
```
grtlr, line 12: 😍

bee-node/build.rs
```diff
@@ -1,15 +1,58 @@
...
+    }
 }
 
+#[cfg(not(feature = "dashboard"))]
 fn main() -> Result<(), BuildError> {
+    parse_git_information()
+}
+
+#[cfg(feature = "dashboard")]
+#[tokio::main]
```
grtlr: I think I would use `reqwest::blocking` to get rid of the `async`/`tokio` dependencies here.

bee-node/build.rs
```diff
@@ -1,15 +1,58 @@
...
+async fn main() -> Result<(), BuildError> {
+    parse_git_information()?;
+
+    let should_fetch = std::env::var("FETCH_DASHBOARD").map(|val| val == "1").unwrap_or(false);
```
grtlr: What is the motivation behind the `env` variable? Shouldn't `#[cfg(feature = "dashboard")]` be enough?

bee-node/build.rs
```diff
@@ -36,3 +79,89 @@ fn main() -> Result<(), BuildError> {
...
+    use serde::Deserialize;
+    use zip::ZipArchive;
+
+    use std::{
+        io::{self, BufReader, Cursor},
+        path::Path,
+    };
+
+    const ASSET_PREFIX: &str = "node-dashboard-bee-";
+    const RELEASES_URL: &str = "https://api.github.com/repos/iotaledger/node-dashboard/releases/latest";
```
grtlr: Maybe we should pin the version of the dashboard release so that we don't get any surprises in the future? That way we also wouldn't need the `LatestReleaseAssets` logic.

bee-node/build.rs
```diff
@@ -36,3 +79,89 @@ fn main() -> Result<(), BuildError> {
...
+                BuildError::DashboardRequest(
+                    e.status().map(|code| code.as_u16()),
+                    release_asset.browser_download_url.clone(),
+                )
+            })?;
+
+        let mut content = Cursor::new(response.bytes().await.map_err(|_| BuildError::DashboardDownload)?);
+        io::copy(&mut content, &mut archive).expect("copying failed");
+        let mut archive = ZipArchive::new(BufReader::new(archive)).map_err(|_| BuildError::DashboardInvalidArchive)?;
+
```
grtlr, line 137: We should also check the SHA checksum to make sure the download was not corrupted or MITM'ed.

bee-node/docker/Dockerfile
```diff
@@ -20,23 +20,7 @@ COPY . .
...
-    && nvm alias default $NODE_VERSION \
-    && nvm use default \
-    && git submodule update --init --recursive \
-    && cd bee-node/src/plugins/dashboard/frontend \
-    && npm install \
-    && npm run build-bee \
-    ; fi
-
-RUN if [ "$WITH_DASHBOARD" = true ] ; then cargo build --profile production --features dashboard --bin bee ; else cargo build --profile production --bin bee ; fi
+RUN if [ "$WITH_DASHBOARD" = true ] ; then FETCH_DASHBOARD=1 cargo build --profile production --features dashboard --bin bee ; else cargo build --profile production --bin bee ; fi
```
grtlr: The `FETCH_DASHBOARD=1` could then also be removed here.

grtlr (changes requested):
I'm very much looking forward to this. My IDE has always been acting up when I forgot to initialize the submodule 😀.

bee-node/README.md
thibault-martinez: Why is all of this gone ?

bee-node/Cargo.toml
Adam-Gleave: Yes, this will all be in the dashboard crate 👍

bee-node/README.md
Adam-Gleave: I don't know why I deleted this! Will add back all apart from `npm`.

bee-node/build.rs
Adam-Gleave: Yes, I was thinking of maybe being able to specify the release in a later PR (maybe through an env variable or a new config file section?)

bee-node/build.rs
Adam-Gleave: Just pushed a couple of changes -- `FETCH_DASHBOARD` is there to force a rebuild/redownload of the release, if you wanted to update or redownload it without changing the code in the crate. The new change adds the `cargo:rerun-if-env-changed` to do this properly.

bee-node/README.md
Adam-Gleave: Oh wait, I didn't delete it (it's on the next line of the diff without `npm`, but this comment doesn't show it :trollface:)

bee-node/build.rs
Adam-Gleave: Yep, much better!

bee-node/README.md
thibault-martinez: Ahhh confusing, my bad 👍

bee-node/build.rs
grtlr: I would just hard code it as a `const &str`, same for the SHA sum because this will only very rarely change. You can get it directly from here: `https://github.com/iotaledger/node-dashboard/releases/download/v1.0.0/node-dashboard-bee-1.0.0.zip`

Adam-Gleave: Sure, I can certainly do that.

grtlr (dismissed):

grtlr (dismissed):

grtlr (dismissed):

bee-node/build.rs
```diff
@@ -36,3 +89,60 @@ fn main() -> Result<(), BuildError> {
...
+        let mut archive = tempfile::tempfile().expect("could not create temp file");
+
+        let response = client
+            .get(RELEASE_URL)
+            .send()
+            .and_then(|resp| resp.error_for_status())
+            .map_err(|e| BuildError::DashboardRequest(e.status().map(|code| code.as_u16()), RELEASE_URL.to_string()))?;
+
+        let content = response.bytes().map_err(|_| BuildError::DashboardDownload)?;
```
thibault-martinez: Is there no way to avoid having the whole content in the RAM ?

bee-node/build.rs
```diff
@@ -36,3 +89,60 @@ fn main() -> Result<(), BuildError> {
...
+        let checksum = format!("{:x}", sha256.finalize());
+
+        if checksum != RELEASE_CHECKSUM {
+            return Err(BuildError::DashboardInvalidChecksum);
+        }
+
+        println!("checksum ok");
+
+        let mut content = Cursor::new(content);
+        io::copy(&mut content, &mut archive).expect("copying failed");
```
thibault-martinez: Like, is there no crate that does a checksum on a file ? You could do it here

bee-node/build.rs
```diff
@@ -36,3 +89,60 @@ fn main() -> Result<(), BuildError> {
...
+        "https://github.com/iotaledger/node-dashboard/releases/download/v1.0.0/node-dashboard-bee-1.0.0.zip";
+    const RELEASE_CHECKSUM: &str = "d61bce7d70b51ea88536b22c5341ccdc3c4651bf8e02e093a771312f3f046c30";
+
+    pub(super) fn fetch<P: AsRef<Path>>(dashboard_dir: P) -> Result<(), BuildError> {
+        println!("downloading latest dashboard release from {}", RELEASE_URL);
+
+        let client = reqwest::blocking::Client::builder()
+            .user_agent("bee-fetch-dashboard")
+            .build()
+            .expect("could not create client");
```
Alex6323, line 112: Should we have an error variant for this as well?

bee-node/build.rs
```diff
@@ -36,3 +89,60 @@ fn main() -> Result<(), BuildError> {
...
+
+        let mut archive = tempfile::tempfile().expect("could not create temp file");
```
Alex6323: same question here

bee-node/build.rs
Adam-Gleave: I think we can do something -- I'll look into it.

bee-node/build.rs
Adam-Gleave: Done :+1:

Cargo.toml
```diff
@@ -23,6 +23,7 @@ members = [
 	"bee-ternary",
 	"bee-test",
 ]
+resolver = "2"
```
thibault-martinez: Is this necessary ? It's the default since edition 2021

Adam-Gleave: Unfortunately yes, it fails to build with the `checksums` crate without it (it causes some strange error in `libp2p-noise`).

bee-node/build.rs
```diff
@@ -36,3 +89,53 @@ fn main() -> Result<(), BuildError> {
...
+            .expect("could not create client");
+
+        let mut tmp_file = tempfile::NamedTempFile::new().expect("could not create temp file");
+
+        client
+            .get(RELEASE_URL)
+            .send()
+            .and_then(|resp| resp.error_for_status())
+            .map_err(|e| BuildError::DashboardRequest(e.status().map(|code| code.as_u16()), RELEASE_URL.to_string()))?
+            .copy_to(&mut tmp_file)
```
thibault-martinez, line 121: Just to confirm: this is streaming from the client to the file, right ? Not first downloading then writing ?

thibault-martinez (dismissed):

bee-node/build.rs
Adam-Gleave: I don't believe this should ever fail.

bee-node/build.rs
Adam-Gleave: I'm not sure we really need another error variant for this, since it shouldn't really happen, and is pretty self-explanatory if it does ever get triggered, but I'm fine with adding one if we feel it's necessary.

bee-node/build.rs
Adam-Gleave: Yes, this should stream bytes directly from the response to the file.

grtlr (dismissed):

Alex6323 (dismissed):

bee-node/build.rs
Alex6323: that's what I was curious about, too, but I can also confirm from the reqwest docs.

samuel-rufi (dismissed):

thibault-martinez (approved):

samuel-rufi (approved):

grtlr (approved):

PR #1193, opened 2022-03-01
Derive PartialEq for InfoResponse

Thoralf-M:
# Description of change

Derive PartialEq for InfoResponse

## Type of change

## Change checklist

### DIFF
--- a/bee-api/bee-rest-api/src/types/responses.rs
+++ b/bee-api/bee-rest-api/src/types/responses.rs
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 
 /// Response of GET /api/v2/info.
 /// Returns general information about the node.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct InfoResponse {
     pub name: String,
     pub version: String,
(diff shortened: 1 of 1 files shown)


thibault-martinez (approved):
Out of curiosity, why ?

pvdrz (approved):

Thoralf-M:
> Out of curiosity, why ?

I wanted to store the `ProtocolResponse` inside of iota.rs so we don't have to request it from the node again every time we need it next to some other data like if local pow should be used or not and that's stored inside of a struct with PartialEq, but I just noticed that we don't need it there
But could still be useful for something else in the future I think

samuel-rufi (approved):

thibault-martinez:
> > Out of curiosity, why ?
> 
> I wanted to store the `ProtocolResponse` inside of iota.rs so we don't have to request it from the node again every time we need it next to some other data like if local pow should be used or not and that's stored inside of a struct with PartialEq, but I just noticed that we don't need it there But could still be useful for something else in the future I think

What if since you stored the info, it changed ?

Thoralf-M:
> > > Out of curiosity, why ?
> > 
> > 
> > I wanted to store the `ProtocolResponse` inside of iota.rs so we don't have to request it from the node again every time we need it next to some other data like if local pow should be used or not and that's stored inside of a struct with PartialEq, but I just noticed that we don't need it there But could still be useful for something else in the future I think
> 
> What if since you stored the info, it changed ?

You mean in the future if it would get updated via a milestone?
The stored state also gets updated (every 60 seconds by default)

PR #1197, opened 2022-03-02
Version checker plugin

Adam-Gleave:
Adds a version checker plugin to notify the user if there is a more recent release version of the `bee-node` crate available (ignoring pre-release versions). Addresses #219.

### DIFF
--- a/bee-node/bee-node/src/fullnode/builder.rs
+++ b/bee-node/bee-node/src/fullnode/builder.rs
@@ -154,7 +154,7 @@ impl<S: NodeStorageBackend> NodeBuilder<FullNode<S>> for FullNodeBuilder<S> {
         let builder = initialize_tangle(builder);
 
         // Start the version checker.
-        let builder = builder.with_worker::<VersionCheckerPlugin>();
+        let builder = builder.with_worker_cfg::<VersionCheckerPlugin>(std::env!("CARGO_PKG_VERSION").to_string());
 
         // Start serving the dashboard (if enabled).
         #[cfg(feature = "dashboard")]
--- a/bee-node/bee-node/src/fullnode/builder.rs
+++ b/bee-node/bee-node/src/fullnode/builder.rs
@@ -8,7 +8,7 @@ use crate::{
     core::{Core, CoreError, ResourceRegister, TopologicalOrder, WorkerStart, WorkerStop},
     shutdown,
     storage::NodeStorageBackend,
-    util, AUTOPEERING_VERSION,
+    util, AUTOPEERING_VERSION, BEE_VERSION,
 };
 
 use bee_autopeering::{
@@ -154,7 +154,7 @@ impl<S: NodeStorageBackend> NodeBuilder<FullNode<S>> for FullNodeBuilder<S> {
         let builder = initialize_tangle(builder);
 
         // Start the version checker.
-        let builder = builder.with_worker_cfg::<VersionCheckerPlugin>(std::env!("CARGO_PKG_VERSION").to_string());
+        let builder = builder.with_worker_cfg::<VersionCheckerPlugin>(BEE_VERSION.to_string());
 
         // Start serving the dashboard (if enabled).
         #[cfg(feature = "dashboard")]
--- a/bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
+++ b/bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
@@ -41,7 +41,9 @@ impl<N: Node> Worker<N> for VersionCheckerPlugin {
             );
 
             while ticker.next().await.is_some() {
-                show_version_status(&current_version).await;
+                if let Some(release_info) = get_release_info().await {
+                    show_version_status(&current_version, release_info).await;
+                }
             }
 
             log::info!("Stopped.");
@@ -53,27 +55,32 @@ impl<N: Node> Worker<N> for VersionCheckerPlugin {
 
 /// Alerts the user (through a `WARN` level log event) if there is a newer release version of Bee available,
 /// and provides a URL to the release's summary webpage.
-async fn show_version_status(current_version: &str) {
-    if let Some(release_info) = get_release_info().await {
-        match (semver::Version::parse(current_version), latest(release_info)) {
-            (Err(e), _) => log::error!("Error parsing current Bee version ({}): {}.", current_version, e),
-            (Ok(current), Some(latest)) => {
-                if latest.version > current {
-                    log::warn!(
-                        "Found a more recent release version ({}), available at {}.",
-                        latest.version,
-                        latest.html_url,
-                    );
-                } else {
-                    log::info!("On the latest release version ({}).", current_version);
-                }
-            }
-            _ => log::warn!("Could not identify the latest release version."),
+async fn show_version_status(current_version: &str, release_info: Vec<ReleaseInfoBuilder>) {
+    let current_version = match semver::Version::parse(current_version) {
+        Err(e) => {
+            log::error!("Error parsing current Bee version ({}): {}.", current_version, e);
+            return;
         }
+        Ok(current_version) => current_version,
+    };
+
+    match latest(release_info, !current_version.pre.is_empty()) {
+        Some(latest) => {
+            if latest.version > current_version {
+                log::warn!(
+                    "Found a more recent release version ({}), available at {}.",
+                    latest.version,
+                    latest.html_url,
+                );
+            } else {
+                log::info!("On the latest release version ({}).", current_version);
+            }
+        }
+        None => log::warn!("Could not identify the latest release version."),
     }
 }
 
-/// Returns a list of [`ReleaseInfoBuilder`]s describing all Github releases, if the API request and
+/// Returns a list of [`ReleaseInfoBuilder`]s describing all Github releases if the API request and
 /// deserialization was successful. Otherwise, returns `None`.
 async fn get_release_info() -> Option<Vec<ReleaseInfoBuilder>> {
     let client = reqwest::Client::builder().user_agent("bee").build().unwrap();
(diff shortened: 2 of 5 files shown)


pvdrz (dismissed):

bee-node/bee-node/src/fullnode/builder.rs
```diff
@@ -155,7 +155,7 @@ impl<S: NodeStorageBackend> NodeBuilder<FullNode<S>> for FullNodeBuilder<S> {
         let builder = initialize_tangle(builder);
 
         // Start the version checker.
-        let builder = builder.with_worker::<VersionCheckerPlugin>();
+        let builder = builder.with_worker_cfg::<VersionCheckerPlugin>(std::env!("CARGO_PKG_VERSION").to_string());
```
thibault-martinez: We have a `BEE_VERSION` constant for that

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
```diff
@@ -5,43 +5,108 @@
 
 #![warn(missing_docs)]
 
+mod release_info;
+
+use crate::release_info::{ReleaseInfo, ReleaseInfoBuilder};
```
thibault-martinez:
```suggestion
use release_info::{ReleaseInfo, ReleaseInfoBuilder};
```

pvdrz (approved):

samuel-rufi (dismissed):

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
```diff
@@ -5,43 +5,108 @@
...
 impl<N: Node> Worker<N> for VersionCheckerPlugin {
-    type Config = ();
+    type Config = VersionCheckerConfig;
     type Error = Infallible;
 
-    async fn start(node: &mut N, _config: Self::Config) -> Result<Self, Self::Error> {
+    async fn start(node: &mut N, current_version: Self::Config) -> Result<Self, Self::Error> {
         node.spawn::<Self, _, _>(|shutdown| async move {
-            info!("Running.");
+            log::info!("Version checker running.");
```
thibault-martinez: Can you just use `Running` and `Stopped` for consistency ? The name is kind of already in the log metadata.

bee-node/bee-plugin/bee-plugin-version-checker/src/release_info.rs
```diff
@@ -0,0 +1,55 @@
...
+    ///  - `None` if there is an error parsing a version from the release tag.
+    ///  - `None` if this is a pre-release.
+    ///  - `Some` otherwise.
+    pub(crate) fn build(self) -> Option<ReleaseInfo> {
+        let mut version = self.tag_name.clone();
+        version.retain(|c| c != 'v');
+
+        match Version::parse(&version) {
+            Err(e) => {
+                log::warn!("error parsing version from tag {}: {}", self.tag_name, e);
```
thibault-martinez:
```suggestion
                log::warn!("Error parsing version from tag {}: {}", self.tag_name, e);
```

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
```diff
@@ -5,43 +5,108 @@
...
         Ok(Self::default())
     }
 }
+
+/// Alerts the user (through a `WARN` level log event) if there is a newer release version of Bee available,
+/// and provides a URL to the release's summary webpage.
+async fn show_version_status(current_version: &str) {
+    if let Some(release_info) = get_release_info().await {
+        match (semver::Version::parse(current_version), latest(release_info)) {
+            (Err(e), _) => println!("Error parsing current Bee version ({}): {}.", current_version, e),
```
thibault-martinez: Should be an error log ?

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
```diff
@@ -5,43 +5,108 @@
...
+        .get("https://api.github.com/repos/iotaledger/bee/releases")
+        .send()
+        .await
+        .and_then(|resp| resp.error_for_status())
+    {
+        Ok(resp) => match resp.json::<Vec<ReleaseInfoBuilder>>().await {
+            Ok(releases) => Some(releases),
+            Err(e) => {
+                // Error deserializing response.
+                log::error!("{}", e);
```
pvdrz: would it be useful to add a tiny description here in case it happens? Same below

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
```diff
@@ -5,43 +5,108 @@
...
+
+/// Returns the latest full release (i.e. not a pre-release version) given a vector of [`ReleaseInfoBuilder`]s.
+fn latest(release_info: Vec<ReleaseInfoBuilder>) -> Option<ReleaseInfo> {
+    let mut releases = release_info
+        .into_iter()
+        .filter_map(|info| info.build())
+        .collect::<Vec<_>>();
+
+    releases.sort();
+    releases.pop()
```
pvdrz: can't we do `.max()` instead?

pvdrz (changes requested):

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
Adam-Gleave: Yes, it definitely should be!

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
Adam-Gleave: Sure :+1:

samuel-rufi (dismissed):

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
```diff
@@ -5,43 +5,102 @@
...
+                } else {
+                    log::info!("On the latest release version ({}).", current_version);
+                }
+            }
+            _ => log::warn!("Could not identify the latest release version."),
+        }
+    }
+}
+
+/// Returns a list of [`ReleaseInfoBuilder`]s describing all Github releases, if the API request and
```
Alex6323: Remove `,` before `if` ?? :trollface:

Alex6323 (dismissed):

bee-node/bee-plugin/bee-plugin-version-checker/src/release_info.rs
```diff
@@ -0,0 +1,59 @@
...
+impl ReleaseInfoBuilder {
+    /// Attempts to build a [`ReleaseInfo`].
+    ///
+    /// Returns:
+    ///  - `None` if there is an error parsing a version from the release tag.
+    ///  - `None` if this is a pre-release, and the current `bee-node` is *not* a pre-release.
+    ///  - `Some` otherwise.
+    pub(crate) fn build(self, pre_release: bool) -> Option<ReleaseInfo> {
+        let mut version = self.tag_name.clone();
+        version.retain(|c| c != 'v');
```
thibault-martinez: I know this should be fine in pretty much all cases but it still seems a bit weird in case for some reason there is another `v`. Maybe we can be a bit smarter and only remove v if it's the first char of the tag ?

bee-node/bee-plugin/bee-plugin-version-checker/src/release_info.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        match Version::parse(&version) {
+            Err(e) => {
+                log::warn!("Error parsing version from tag {}: {}", self.tag_name, e);
+                None
+            }
+            Ok(version) => {
+                let selected = if pre_release { true } else { version.pre.is_empty() };
```
thibault-martinez: Maybe rework into `pre_release || version.pre.is_empty()`

pvdrz (approved):

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
```diff
@@ -5,43 +5,167 @@
...
+            Ok(releases) => Some(releases),
+            Err(e) => {
+                // Error deserializing response.
+                log::error!("Error deserializing Github API reponse: {}", e);
+                None
+            }
+        },
+        Err(e) => {
+            // Error occured during API request.
+            log::error!("Error performing Github API request: {}", e);
```
grtlr, line 104: Should we maybe just `warn` the user if a request fails? It could happen that GitHub is down, which wouldn't be too bad because the plugin will check again in another hour.

The `error` for  deserialization above is fine though.

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
```diff
@@ -5,43 +5,167 @@
...
+                html_url: "".to_string(),
+                tag_name: "v0.2.0".to_string(),
+            },
+            ReleaseInfoBuilder {
+                html_url: "".to_string(),
+                tag_name: "v0.3.0-rc4".to_string(),
+            },
+            ReleaseInfoBuilder {
+                html_url: "".to_string(),
+                tag_name: "v0.3.0-rc5".to_string(),
```
grtlr, line 162: We have been doing stuff like `v0.3.0-rc5-trigger-docker`, which doesn't really follow SemVer. Should we stop doing that and switch to something that's compatible, or would it make sense to catch stuff like that in the `version-checker-plugin`?

grtlr (changes requested):

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
thibault-martinez: It is really non-compliant with semver though ?

thibault-martinez: Also these tags are just tests until we stabilize docker, we won't do that anymore at some point.

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
Adam-Gleave: I'm not sure, this is still an error in the sense that the plugin has failed to perform the task it's intended to perform. I think an error message is appropriate, it's just not a fatal error. Happy to change though if others agree :+1:

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
Adam-Gleave: I don't think there's anything in the semver spec that indicates this is incorrect usage, so it should be fine.

What might be worth considering though is that, with our current pre-release naming `<prefix><number>`, we might run into problems with version precedence later, since dot-separated terms are only compared numerically if they only contain  [numerical values](https://docs.rs/semver/1.0.6/semver/struct.Prerelease.html#total-ordering). From the spec:
> Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:
>
>    Identifiers consisting of only digits are compared numerically.
>
>    Identifiers with letters or hyphens are compared lexically in ASCII sort order.

\
So with our current naming system, if there was ever an `rc11` version, the precedence would be wrong: 
```
0.3.0-rc1 < 0.3.0-rc11 < 0.3.0-rc2
```

Maybe it's worth changing to use `<prefix>.<number>` instead? That would keep the order correct:
```
0.3.0-rc.1 < 0.3.0-rc.2 < 0.3.0-rc.11
```

Adam-Gleave: (And in the case of the above, I guess we would use `rc.5.trigger-docker`?)

thibault-martinez: Yeah I actually think `rc.*` is the proper way of doing it

grtlr: Yeah, the precedence was where I was getting at. I also like the "dot" notation. Thank you for clearing that up!

bee-node/bee-plugin/bee-plugin-version-checker/src/lib.rs
grtlr: Yeah, that makes sense too!

grtlr (approved):

thibault-martinez (approved):

PR #1211, opened 2022-03-04
Make `Tangle::update_metadata` atomic by using entries

pvdrz:
This should fix our data race or at least make it less likely.

I'd appreciate if you could check that the `update_metadata` logic is still the same as before. The main difference is that we acquire a write guard over the `Vertices` partition we need only once, which should forbid other threads from introducing inconsistencies while `update_metadata` is running.

The only possible issue I was able to see is that `update_metadata` is not completely atomic because parents need to be updated after the write guard is dropped. This is because parents might potentially be in other partitions.  Even then, this shouldn't affect the metadata updates done over the vertex itself.

### DIFF
--- a/bee-tangle/src/tangle.rs
+++ b/bee-tangle/src/tangle.rs
@@ -405,7 +405,7 @@ impl<B: StorageBackend> Tangle<B> {
         metadata: MessageMetadata,
         prevent_eviction: bool,
     ) -> Option<MessageRef> {
-        let mut vertex = self.vertices.get_mut_or_empty(message_id).await;
+        let mut vertex = self.vertices.entry(message_id).await.or_empty();
 
         if prevent_eviction {
             vertex.prevent_eviction();
@@ -423,7 +423,7 @@ impl<B: StorageBackend> Tangle<B> {
 
             // Insert children for parents
             for &parent in parents.iter() {
-                self.vertices.get_mut_or_empty(parent).await.add_child(message_id);
+                self.vertices.entry(parent).await.or_empty().add_child(message_id);
             }
 
             msg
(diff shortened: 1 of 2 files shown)


bee-tangle/src/tangle.rs
```diff
@@ -489,27 +489,49 @@ impl<B: StorageBackend> Tangle<B> {
...
+        let entry = self.vertices.entry(*message_id).await;
+        let was_vacant = matches!(entry, Entry::Vacant(_));
+        let mut vertex = entry.or_empty();
+
+        let mut perform_eviction = false;
+        let mut parents = None;
+
+        if was_vacant || vertex.message().is_none() {
+            if let Ok(Some((msg, metadata))) = self.storage_get(message_id) {
+                vertex.prevent_eviction();
```
pvdrz: @thibault-martinez do you think this actually matters? there is no await point between `prevent_eviction` and `allow_eviction` and we are holding the guard to the partition via `vertex`.

bee-tangle/src/vertices.rs
```diff
@@ -21,6 +24,59 @@ fn equivalent_id(message_id: &MessageId) -> impl Fn(&(MessageId, Vertex)) -> boo
...
+            let entry = table.insert_entry(self.hash, (self.message_id, Vertex::empty()), move |(message_id, _)| {
+                make_hash(&self.hash_builder, message_id)
+            });
+            self.len.fetch_add(1, Ordering::Relaxed);
+            &mut entry.1
+        })
+    }
+}
+
+pub(crate) enum Entry<'a> {
```
grtlr, line 60:
```suggestion
#[must_use]
pub(crate) enum Entry<'a> {
```

grtlr: In general, I think we should slap `#[must_use]` onto many function in `Tangle` to find places where we might misuse the API at the moment.

pvdrz:
Closing this until we are sure we can fix the issue

PR #1239, opened 2022-03-13
Replace warp in bee-rest-api with axum

samuel-rufi:


bee-api/bee-rest-api/src/endpoints/error.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::types::body::DefaultErrorResponse;
+
+#[derive(Debug, Clone)]
+pub(crate) enum ApiError {
+    Forbidden,
+    BadRequest(String),
+    NotFound(String),
+    ServiceUnavailable(String),
+    InternalError,
+    StorageBackend,
```
pvdrz: Do we actually use this variant in a different way than `InternalError`? Otherwise I'd just use `InternalError` for `StorageBackend` errors.

bee-api/bee-rest-api/src/endpoints/auth.rs
```diff
@@ -0,0 +1,148 @@
...
+        let Extension(args) = Extension::<ApiArgsFullNode<S>>::from_request(req)
+            .await
+            .map_err(|_| ApiError::Forbidden)?;
+
+        // Check if the requested endpoint is open for public use.
+        if args.rest_api_config.public_routes().is_match(&uri.to_string()) {
+            return Ok(Auth { phantom: PhantomData });
+        }
+
+        // Extract the token from the authorization header
```
Alex6323:
```suggestion
        // Extract the token from the authorization header.
```
Can't help myself :trollface:

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -90,7 +66,7 @@ impl RestApiConfigBuilder {
     }
 
     /// Sets the binding address for the REST API.
-    pub fn bind_address(mut self, addr: &str) -> Self {
+    pub fn with_bind_address(mut self, addr: &str) -> Self {
```
Alex6323, line 68: I would simplify the method body. Since you panic in the error case I would just use `expect`. The method body becomes shorter this way:
```rust
self.bind_address.replace(addr.parse().expect("error parsing IP address"));
self
```

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -100,26 +76,32 @@ impl RestApiConfigBuilder {
...
     /// Set if the feature proof-of-work should be enabled or not.
-    pub fn feature_proof_of_work(mut self, value: bool) -> Self {
+    pub fn with_feature_proof_of_work(mut self, value: bool) -> Self {
         self.feature_proof_of_work.replace(value);
         self
     }
 
     /// Sets the while flag solidification timeout.
-    pub fn white_flag_solidification_timeout(mut self, timeout: u64) -> Self {
+    pub fn with_white_flag_solidification_timeout(mut self, timeout: u64) -> Self {
```
Alex6323, line 99: Should we document that we expect seconds here?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -152,24 +134,37 @@ impl RestApiConfigBuilder {
...
+            RegexSet::new(routes.iter().map(|r| route_to_regex(r)).collect::<Vec<_>>())
+                .expect("invalid protected route provided")
+        };
+
         let feature_proof_of_work = self.feature_proof_of_work.unwrap_or(DEFAULT_FEATURE_PROOF_OF_WORK);
+
         let white_flag_solidification_timeout = self
             .white_flag_solidification_timeout
-            .unwrap_or(DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT);
+            .map_or_else(|| DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT, Duration::from_secs);
```
Alex6323:
```suggestion
            .map_or(DEFAULT_WHITE_FLAG_SOLIDIFICATION_TIMEOUT, Duration::from_secs);
```
because it's just a constant we don't need to be lazy here.

bee-api/bee-rest-api/src/endpoints/routes/api/plugins/debug/white_flag.rs
```diff
@@ -1,115 +1,69 @@
...
+        ));
     } else {
-        let array = parents_json.as_array().ok_or_else(|| {
-            reject::custom(CustomRejection::BadRequest(
-                "Invalid parents: expected an array of MessageId".to_string(),
-            ))
-        })?;
+        let array = parents_json
+            .as_array()
+            .ok_or_else(|| ApiError::BadRequest("Invalid parents: expected an array of MessageId".to_string()))?;
```
Alex6323: There's a lot of repetition of this `BadRequest` message. You might want to store it in a constant. I don't know if the Rust compiler is smart enough to deduplicate those strings. @pvdrz ?

bee-api/bee-rest-api/src/endpoints/routes/api/plugins/debug/white_flag.rs
```diff
@@ -159,36 +120,28 @@ pub(crate) async fn white_flag<B: StorageBackend>(
         }
     }
 
-    // TODO
     let mut metadata = WhiteFlagMetadata::new(index, 0);
 
     // Wait for either all parents to get solid or the timeout to expire.
```
Alex6323:
```suggestion
    // Wait for all parents to get solid or the timeout to expire.
```
The `either` confused me :trollface: . At first I thought it would be associated with `parents` as in: `either parent to get solid`, but then realized it's `either ... or`. Removing it should be fine and less confusing to readers :slightly_smiling_face:

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
```diff
@@ -1,87 +1,55 @@
...
     let alias_v = &value["alias"];
 
     let mut multi_address = multi_address_v
         .as_str()
-        .ok_or_else(|| {
-            reject::custom(CustomRejection::BadRequest(
-                "invalid multi address: expected a string".to_string(),
-            ))
-        })?
+        .ok_or_else(|| ApiError::BadRequest("invalid multiaddress".to_string()))?
```
Alex6323: I feel like the `ApiError` variants should contain `'static str`s instead of `String`s. This way you can prevent those closures and make the code overall a bit nicer.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
```diff
@@ -1,87 +1,55 @@
...
-pub(crate) fn add_peer(
-    value: JsonValue,
-    peer_manager: ResourceHandle<PeerManager>,
-    network_controller: ResourceHandle<NetworkCommandSender>,
-) -> Result<impl Reply, Rejection> {
+pub(crate) async fn add_peer<B: StorageBackend>(
+    Json(value): Json<Value>,
+    Extension(args): Extension<ApiArgsFullNode<B>>,
+) -> Result<impl IntoResponse, ApiError> {
     let multi_address_v = &value["multiAddress"];
```
Alex6323: I personally think `multi_addr_val` or `multi_addr_value` are more readable abbreviations, but I would even consider not abbreviating at all: `multi_address_value`. Your choice though :smile: 
(same for alias below)

bee-api/bee-rest-api/src/endpoints/routes/api/v2/info.rs
```diff
@@ -1,106 +1,63 @@
...
+                v_byte_factor_data: args.protocol_config.byte_cost().v_byte_factor_data,
             },
         },
         metrics: MetricsResponse {
-            messages_per_second: 0.0,            // TODO
-            referenced_messages_per_second: 0.0, // TODO
-            referenced_rate: 0.0,                // TODO
+            messages_per_second: 0f64,            // TODO
+            referenced_messages_per_second: 0f64, // TODO
+            referenced_rate: 0f64,                // TODO
```
Alex6323: Wil this be part of another PR? One TODO above `metrics` should be enough though, and despite being obvious that those values are temporarily set to `0`, I would document it with e.g `TODO: use actual metrics values` or something.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/info.rs
```diff
@@ -1,106 +1,63 @@
...
         },
         features: {
             let mut features = Vec::new();
-            if rest_api_config.feature_proof_of_work() {
+            if args.rest_api_config.feature_proof_of_work() {
                 features.push("PoW".to_string())
             }
             features
         },
         plugins: Vec::new(), // TODO
```
Alex6323: Properly document the TODO if it won't make it into this PR (or fix the TODO :trollface: )

bee-api/bee-rest-api/src/endpoints/routes/api/v2/message_children.rs
```diff
@@ -1,53 +1,34 @@
...
-    children.truncate(max_results);
-    Ok(warp::reply::json(&MessageChildrenResponse {
+    children.truncate(MAX_RESPONSE_RESULTS);
+
+    Ok(Json(MessageChildrenResponse {
         message_id: message_id.to_string(),
-        max_results,
+        max_results: MAX_RESPONSE_RESULTS,
         count,
         children_message_ids: children.iter().map(MessageId::to_string).collect(),
```
Alex6323: You don't need to call `truncate`, just use `children.iter().take(MAX_RESPONSE_RESULTS).map(...).collect()`. You can then also get rid of the extra binding `count`.

bee-api/bee-rest-api/src/endpoints/routes/api/v2/message_metadata.rs
```diff
@@ -1,56 +1,35 @@
...
+    Extension(args): Extension<ApiArgsFullNode<B>>,
+) -> Result<impl IntoResponse, ApiError> {
+    if !args.tangle.is_confirmed_threshold(CONFIRMED_THRESHOLD) {
+        return Err(ApiError::ServiceUnavailable("the node is not synchronized".to_string()));
     }
 
-    match tangle.get(&message_id).await.map(|m| (*m).clone()) {
+    match args.tangle.get(&message_id).await.map(|m| (*m).clone()) {
         Some(message) => {
             // existing message <=> existing metadata, therefore unwrap() is safe
```
Alex6323:
```suggestion
            // Panic: existing message <=> existing metadata, therefore unwrap() is safe.
```
Sorry...so nit-picky of me :trollface: You wanna go even more mathy?
(message exists <=> metadata exists) => unwrap save. :smile:

bee-api/bee-rest-api/src/endpoints/routes/api/v2/output.rs
```diff
@@ -1,87 +1,63 @@
...
-                Ok(warp::reply::json(&OutputResponse {
+                Ok(Json(OutputResponse {
                     message_id: output.message_id().to_string(),
                     transaction_id: output_id.transaction_id().to_string(),
                     output_index: output_id.index(),
-                    is_spent: is_spent.is_some(),
+                    is_spent: consumed_output.is_some(),
                     // TODO
                     milestone_index_spent: None,
                     // TODO
```
Alex6323: I know it's not introduced in this PR, but can we document or remove or fix those TODOs?

bee-api/bee-rest-api/src/endpoints/routes/api/v2/message_metadata.rs
samuel-rufi: haha good one 😃

bee-api/bee-rest-api/src/endpoints/config.rs
pvdrz: would it be bad to pass a `Duration` instead?

bee-api/bee-rest-api/src/endpoints/config.rs
pvdrz: true

bee-api/bee-rest-api/src/endpoints/routes/api/plugins/debug/white_flag.rs
pvdrz: I think it isn't, static string slices are tricky tho. Maybe `ApiError` should have an `invalid_parents()` constructor method

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
pvdrz: I think static string slices are optimized fancily but I don't have any hard evidence of it

bee-api/bee-rest-api/src/endpoints/routes/api/v2/message_metadata.rs
pvdrz: I think you can get both with `get_message_and_metadata` or something.

pvdrz (changes requested):

bee-api/bee-rest-api/src/endpoints/config.rs
Alex6323: I also thought about that, but as a builder it needs to be deserializable from the config file, so internally it has to be a `u64`. If the argument becomes a `Duration` then we'ld end up converting from Duration (argument) -> u64 (builder field) -> Duration (config field). I don't know what's better. I think documenting that we expect seconds should be fine I guess.

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: good point! done

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: Yes, that was also the outcome of the last discussion about this. So I guess I will just document as Alex said that it's in seconds.

bee-api/bee-rest-api/src/endpoints/error.rs
samuel-rufi: remove it

bee-api/bee-rest-api/src/endpoints/routes/api/v2/info.rs
samuel-rufi: done 👍

bee-api/bee-rest-api/src/endpoints/routes/api/v2/info.rs
samuel-rufi: done 👍

bee-api/bee-rest-api/src/endpoints/routes/api/v2/message_children.rs
samuel-rufi: good point, done!

bee-api/bee-rest-api/src/endpoints/routes/api/v2/output.rs
samuel-rufi: done

bee-api/bee-rest-api/src/endpoints/routes/api/v2/message_metadata.rs
samuel-rufi: seems like that code is only in https://github.com/iotaledger/bee/blob/mainnet-develop-0.4/bee-tangle/src/tangle.rs#L357-L363 branch, not shimmer-develop yet

so i guess i have to change the comment as Alex suggested

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
samuel-rufi: @Alex6323 should we maybe even add the suffix `_json` instead? would be even more clear I guess

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
samuel-rufi: But how to solve these kinds of errors?

`serde_json::from_slice(&bytes.to_vec()).map_err(|e| ApiError::BadRequest(e.to_string()))?;`

If adding `&` infront then the borrow doesn't last for `'static`
Any solution I'm missing?

bee-api/bee-rest-api/src/endpoints/routes/api/plugins/debug/white_flag.rs
samuel-rufi: Not sure if storing in a constant is really so nice or if it's better to live with the two duplicate strings.
I actually like the idea from @pvdrz however for consistency we would need to do this for all other errors too.
I decided against it when I started the implementation because we would end up with a quite a big error enum with lots of variants probably that couldn't be really reused as these errors are quite special.
But maybe it makes sense to change this in an future PR?

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
Alex6323: You could introduce different error variants, maybe: `ApiError::BadJsonRequest(JsonError)` that simply wraps the error (no extra allocation), and `ApiError::BadRequest(&'static str)` (also no allocation), but if you feel this is too much of a burden right now, then just leave it bee. I don't think it's a biggie, but something that can be optimized. Your choice! :slightly_smiling_face:

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
Alex6323: @samuel-rufi I don't have a strong opinion on that. I sometimes would do it, sometimes not. Probably depends on my mood. Your choice!!

bee-api/bee-rest-api/src/endpoints/routes/api/plugins/debug/white_flag.rs
Alex6323: Weren't it even 3? Remember that those strings end up being stored in the binary several times. If that's the only instance here, I would say it's not an issue yet, and you can ignore me.

bee-api/bee-rest-api/src/endpoints/config.rs
Alex6323: Alright, then let's do this!

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
Alex6323: or would something like that work for you?
```rust
#[derive(thiserror::Error, Debug)]
enum ApiError {
   ...
   [#error("{0}")]
   JsonBadRequest(JsonError),
   [#error("Invalid parents: ...")]
   InvalidParentsBadRequest,
   ...
}
```

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
samuel-rufi: okay to me it seems JSON is more clear than value. Will do it consistently in the API!

pvdrz (dismissed):

bee-api/bee-rest-api/src/endpoints/routes/api/v2/add_peer.rs
samuel-rufi: Fixed this

Alex6323 (dismissed):

pvdrz (dismissed):

pvdrz (dismissed):

bee-node/bee-plugin/bee-plugin-dashboard/Cargo.toml
```diff
@@ -11,6 +11,7 @@ keywords = [ "iota", "tangle", "bee", "plugin", "dashboard" ]
 homepage = "https://www.iota.org"
 
 [dependencies]
+bee-gossip = { version = "0.6.0", path = "../../../bee-network/bee-gossip", default-features = false}
```
thibault-martinez:
```suggestion
bee-gossip = { version = "0.6.0", path = "../../../bee-network/bee-gossip", default-features = false }
```

bee-node/bee-node/src/lib.rs
```diff
@@ -7,6 +7,8 @@
 
 #![cfg_attr(doc_cfg, feature(doc_cfg))]
 
+extern crate bee_rest_api;
```
thibault-martinez: No!!!

bee-api/bee-rest-api/Cargo.toml
```diff
@@ -27,24 +27,29 @@ bee-storage = { version = "0.12.0", path = "../../bee-storage/bee-storage", defa
 bee-tangle = { version = "0.3.0", path = "../../bee-tangle", default-features = false, optional = true }
 
 async-trait = { version = "0.1.51", default-features = false, optional = true }
+auth-helper = { version = "0.2.0", default-features = false }
+axum = { version = "0.4.8", default-features = false, features = [ "json", "headers" ], optional = true }
+base64 = { version = "0.13.0", default-features = false}
```
thibault-martinez:
```suggestion
base64 = { version = "0.13.0", default-features = false }
```

bee-api/bee-rest-api/src/endpoints/path_params.rs
```diff
@@ -1,64 +0,0 @@
-// Copyright 2020-2021 IOTA Stiftung
```
thibault-martinez, line 1: Delete this file ?

bee-api/bee-rest-api/src/endpoints/routes/api/v2/mod.rs
```diff
@@ -1,6 +1,10 @@
-// Copyright 2020-2021 IOTA Stiftung
+// Copyright 2020-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use axum::Router;
+
+use crate::endpoints::storage::StorageBackend;
```
thibault-martinez: Imports after mods

bee-node/bee-node/src/lib.rs
thibault-martinez: Reopening as it's still here.

bee-api/bee-rest-api/Cargo.toml
```diff
@@ -27,24 +27,29 @@ bee-storage = { version = "0.12.0", path = "../../bee-storage/bee-storage", defa
...
 futures = { version = "0.3.17", default-features = false, optional = true }
 iota-crypto = { version = "0.10.0", default-features = false, features = [ "blake2b" ], optional = true }
+lazy_static = {version = "1.4.0", default-features = false }
 log = { version = "0.4.14", default-features = false, optional = true }
 multiaddr = { version = "0.13.0", default-features = false }
 num_cpus = { version = "1.13.0", default-features = false, optional = true }
 packable = { version = "0.3.2", default-features = false }
 prefix-hex = { version = "0.2.0", default-features = false }
 primitive-types = { version = "0.10.1", default-features = false, features = [ "serde" ] }
+regex = { version = "1.5.4", default-features = false }
```
thibault-martinez: Aren't all these new deps only needed for the endpoints feature ?

bee-api/bee-rest-api/src/endpoints/routes/api/v2/mod.rs
```diff
@@ -22,141 +22,34 @@ pub mod treasury;
...
+            .merge(output::filter::<B>())
+            .merge(peer::filter::<B>())
+            .merge(peers::filter::<B>())
+            .merge(receipts::filter::<B>())
+            .merge(receipts_at::filter::<B>())
+            .merge(remove_peer::filter::<B>())
+            .merge(submit_message::filter::<B>())
+            .merge(tips::filter::<B>())
+            .merge(transaction_included_message::filter::<B>())
+            .merge(treasury::filter::<B>()),
```
thibault-martinez: Can you order them alphabetically please, easier to follow with the mod order

bee-api/bee-rest-api/src/endpoints/routes/api/v2/mod.rs
```diff
@@ -22,141 +22,34 @@ pub mod treasury;
...
-        public_routes.clone(),
-        allowed_ips.clone(),
-        peer_manager.clone(),
-        network_command_sender.clone(),
+pub(crate) fn filter<B: StorageBackend>() -> Router {
+    Router::new().nest(
+        "/v2",
+        add_peer::filter::<B>()
+            .merge(info::filter::<B>())
+            .merge(message::filter::<B>())
```
thibault-martinez: Can you put the module to plural like in the URL please? It's a bit inconsistent atm, you have peers but have message and output...

bee-api/bee-rest-api/src/endpoints/routes/health.rs
```diff
@@ -1,51 +1,28 @@
...
+pub(crate) fn filter<B: StorageBackend>() -> Router {
+    Router::new().route("/health", get(health::<B>))
 }
 
-pub(crate) fn health<B: StorageBackend>(
-    tangle: ResourceHandle<Tangle<B>>,
-    peer_manager: ResourceHandle<PeerManager>,
-) -> Result<impl Reply, Infallible> {
-    if is_healthy(&tangle, &peer_manager) {
+pub(crate) async fn health<B: StorageBackend>(
```
thibault-martinez: For this an all similar function in all other module, make them private if they don't actually need to be pub crate

bee-api/bee-rest-api/Cargo.toml
```diff
@@ -27,24 +27,29 @@ bee-storage = { version = "0.12.0", path = "../../bee-storage/bee-storage", defa
 bee-tangle = { version = "0.3.0", path = "../../bee-tangle", default-features = false, optional = true }
 
 async-trait = { version = "0.1.51", default-features = false, optional = true }
+auth-helper = { version = "0.2.0", default-features = false }
+axum = { version = "0.4.8", default-features = false, features = [ "json", "headers" ], optional = true }
```
thibault-martinez: Can you update to latest please ? Better do the changes now if needed

samuel-rufi: updated

bee-api/bee-rest-api/src/endpoints/routes/api/v2/mod.rs
samuel-rufi: fixed

bee-api/bee-rest-api/src/endpoints/routes/api/v2/mod.rs
samuel-rufi: do you think we should change `peer` too? because currently `peers` returns all and `peer` only a single one
we could change it to something like: 
`peers_one`
`peers_all`

what do you think?

bee-api/bee-rest-api/src/endpoints/routes/health.rs
samuel-rufi: done

bee-api/bee-rest-api/Cargo.toml
samuel-rufi: fixed 👍

pvdrz (dismissed):

bee-api/bee-rest-api/src/endpoints/routes/api/v2/messages_submit.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    let nonce = if nonce_json.is_null() {
+        None
+    } else {
+        let parsed_nonce = nonce_json
+            .as_str()
+            .ok_or(ApiError::BadRequest("invalid nonce: expected an u64-string"))?
+            .parse::<u64>()
+            .map_err(|_| ApiError::BadRequest("invalid nonce: expected an u64-string"))?;
+        if parsed_nonce == 0 { None } else { Some(parsed_nonce) }
```
samuel-rufi: We can remove line 125 now because milestone nonces can be zero now with the latest stardust changes, right?

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -220,7 +215,18 @@ impl RestApiConfig {
...
     /// Returns the white flag solidification timeout.
-    pub fn white_flag_solidification_timeout(&self) -> u64 {
+    pub fn white_flag_solidification_timeout(&self) -> Duration {
         self.white_flag_solidification_timeout
     }
 }
+
+pub(crate) fn route_to_regex(route: &str) -> String {
+    // Escape the string to make sure a regex can be built from it.
+    // Existing wildcards `*` get escaped to `\\*`.
```
DaughterOfMars, line 225: `*` actually is not a meta character in regex, so it is not escaped. See [`regex_syntax`](https://docs.rs/regex-syntax/latest/src/regex_syntax/lib.rs.html#178-182) which is called by [`regex`](https://docs.rs/regex/latest/src/regex/re_unicode.rs.html#21-23).

Also, `regex::escape` does not consider unicode characters (\w, \d, \s), but you are not using the feature for those so it will error if someone did try to use them. 🤷

bee-api/bee-rest-api/src/endpoints/error.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use axum::{
+    http::StatusCode,
+    response::{IntoResponse, Response},
+    Json,
+};
+
+use crate::types::body::{DefaultErrorResponse, ErrorBody};
+
+pub enum ApiError {
```
DaughterOfMars, line 15: Use `thiserror` to simplify this?

bee-api/bee-rest-api/src/endpoints/routes/api/plugins/debug/white_flag.rs
```diff
@@ -1,115 +1,64 @@
...
 pub(crate) async fn white_flag<B: StorageBackend>(
-    body: JsonValue,
-    storage: ResourceHandle<B>,
-    tangle: ResourceHandle<Tangle<B>>,
-    bus: ResourceHandle<Bus<'static>>,
-    block_requester: BlockRequesterWorker,
-    requested_blocks: ResourceHandle<RequestedBlocks>,
-    rest_api_config: RestApiConfig,
-) -> Result<impl Reply, Rejection> {
+    Json(body): Json<Value>,
```
Alexandcoats, line 33: `CustomJson`?

DaughterOfMars: `CustomJson`?

samuel-rufi: Fixed

bee-api/bee-rest-api/src/endpoints/config.rs
```diff
@@ -159,24 +129,37 @@ impl RestApiConfigBuilder {
...
-            .into_boxed_slice();
-        let allowed_ips: Box<[IpAddr]> = self
-            .allowed_ips
-            .unwrap_or_else(|| DEFAULT_ALLOWED_IPS.to_vec())
-            .into_boxed_slice();
+            .expect("unsupported protocol");
+
+        let jwt_salt = self.jwt_salt.unwrap_or_else(|| DEFAULT_JWT_SALT.to_string());
+
+        let public_routes = {
```
pvdrz, line 136: maybe you can extract this code to a free function so you don't write it twice

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -68,51 +124,43 @@ where
...
+                .fallback(fallback.into_service());
 
-            server.await;
+            axum::Server::bind(&args.rest_api_config.bind_socket_addr())
+                .serve(app.into_make_service())
+                .with_graceful_shutdown(async {
+                    shutdown.await.ok();
+                })
+                .await
+                .unwrap(); // TODO: handle unwrap
```
pvdrz, line 162: todo!

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -121,68 +169,43 @@ where
...
+                .fallback(fallback.into_service());
 
-            server.await;
+            axum::Server::bind(&config.rest_api_config.bind_socket_addr())
+                .serve(app.into_make_service())
+                .with_graceful_shutdown(async {
+                    shutdown.await.ok();
+                })
+                .await
+                .unwrap(); // TODO: handle unwrap
```
pvdrz, line 199: todo!

bee-api/bee-rest-api/src/endpoints/error.rs
samuel-rufi: Fixed

bee-api/bee-rest-api/src/endpoints/routes/api/v2/blocks.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+async fn blocks<B: StorageBackend>(
+    headers: HeaderMap,
+    CustomPath(block_id): CustomPath<BlockId>,
+    Extension(args): Extension<ApiArgsFullNode<B>>,
+) -> Result<Response, ApiError> {
+    if let Some(value) = headers.get(axum::http::header::ACCEPT) {
+        if value.eq(&*BYTE_CONTENT_HEADER) {
+            return blocks_raw::<B>(block_id, args.clone()).await.map(|r| r.into_response());
```
pvdrz: if you `return` here, you can deduplicate the `else`.

pvdrz (changes requested):

bee-api/bee-rest-api/src/endpoints/mod.rs
samuel-rufi: Hmm not sure how to handle this, do you have an idea?

spawn of node returns `()` so returning an error doens't really work if I understand correctly
```
`    fn spawn<W, G, F>(&mut self, g: G)
    where
        W: Worker<Self>,
        G: FnOnce(oneshot::Receiver<()>) -> F,
        F: Future<Output = ()> + Send + 'static;
```

bee-api/bee-rest-api/src/endpoints/config.rs
samuel-rufi: Hmm, they are in some way similar, but require completely different values. not sure if extracting this code to a function makes it less confusing? :carlos:

DaughterOfMars (dismissed):

pvdrz (dismissed):

Alex6323 (dismissed):

bee-node/bee-plugin/bee-plugin-dashboard/Cargo.toml
```diff
@@ -11,6 +11,7 @@ keywords = [ "iota", "tangle", "bee", "plugin", "dashboard" ]
 homepage = "https://www.iota.org"
 
 [dependencies]
+bee-gossip = { version = "0.6.0", path = "../../../bee-network/bee-gossip", default-features = false }
```
thibault-martinez: Order

bee-node/bee-node/config.alphanet.toml
```diff
@@ -76,34 +76,23 @@ v_byte_factor_data = 1
...
+      "/api/v2/milestones*",
+      "/api/v2/outputs*",
+      "/api/v2/treasury",
+      "/api/v2/receipts*",
 ]
-allowed_ips = [
-    "127.0.0.1",
-    "::1"
+protected_routes       = [
+      "/api/v2/*",
```
thibault-martinez, line 92: What does that mean ? All routes are protected (JWT ?) by default ?

samuel-rufi: Yes exactly, Hornet is doing similar <https://github.com/iotaledger/hornet/blob/develop/config.json#L161-L164>
Btw one line below it says `"/api/plugins/*"` ; since we don't have plugins, should I remove this from the config?

bee-node/bee-plugin/bee-plugin-dashboard/Cargo.toml
samuel-rufi: fixed

bee-node/bee-node/config.alphanet.toml
thibault-martinez: Isn't whiteflag under plugin ? Used to be at least

bee-api/bee-rest-api/src/endpoints/routes/api/v2/milestones_by_id.rs
```diff
@@ -0,0 +1,27 @@
...
+pub(crate) fn filter<B: StorageBackend>() -> Router {
+    Router::new().route("/milestones/:milestone_id", get(milestones_by_id::<B>))
+}
+
+async fn milestones_by_id<B: StorageBackend>(
+    headers: HeaderMap,
+    CustomPath(milestone_id): CustomPath<MilestoneId>,
+    Extension(args): Extension<ApiArgsFullNode<B>>,
+) -> Result<impl IntoResponse, ApiError> {
+    let milestone_index = match args.tangle.get_milestone(milestone_id) {
```
thibault-martinez: Why don't we just return the payload since we have it ?

bee-node/bee-node/config.alphanet.toml
samuel-rufi: you're right, whiteflag is hosted under "/api/plugins/*"

bee-api/bee-rest-api/src/endpoints/auth.rs
```diff
@@ -0,0 +1,148 @@
...
+#[cfg(feature = "dashboard")]
+const DASHBOARD_JWT_HINT: &str = "\"aud\":\"dashboard\"";
+
+lazy_static! {
+    static ref DASHBOARD_ROUTES: RegexSet = {
+        let routes = vec![
+            "/api/v2/info",
+            "/api/v2/messages/*",
+            "/api/v2/outputs/*",
+            "/api/v2/addresses/*",
```
Thoralf-M: Outdated routes

bee-api/bee-rest-api/src/endpoints/routes/api/v2/milestones_by_id.rs
samuel-rufi: Good catch, fixed

bee-api/bee-rest-api/src/endpoints/error.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl IntoResponse for ApiError {
+    fn into_response(self) -> Response {
+        let status_code = match self {
+            ApiError::BadRequest(_) => StatusCode::BAD_REQUEST,
+            ApiError::NotFound => StatusCode::NOT_FOUND,
+            ApiError::ServiceUnavailable(_) => StatusCode::SERVICE_UNAVAILABLE,
+            ApiError::InternalServerError => StatusCode::INTERNAL_SERVER_ERROR,
+            ApiError::Forbidden => StatusCode::FORBIDDEN,
+            ApiError::DependencyError(_) => StatusCode::BAD_REQUEST,
```
Thoralf-M, line 57: Shouldn't this rather be `INTERNAL_SERVER_ERROR`?

Thoralf-M: For most of them probably not

bee-api/bee-rest-api/src/endpoints/auth.rs
samuel-rufi: Good catch, fixed

bee-api/bee-rest-api/src/endpoints/error.rs
samuel-rufi: Nope `StatusCode::BAD_REQUEST` is fine for these errors imo.
`DependencyError` here means JSON deserialization errors, URL path errors, invalid block submitted errors, DTO deserializataion erros, ... which all are "bad request" errors; so should be fine

Or why do you think so?

bee-node/bee-node/src/tools/jwt_api.rs
```diff
@@ -0,0 +1,36 @@
...
+) -> Result<(), JwtApiError> {
+    let claims = ClaimsBuilder::new(
+        local.peer_id().to_string(),
+        node_config.rest_api.jwt_salt().to_owned(),
+        API_AUDIENCE_CLAIM.to_owned(),
+    )
+    .build()?;
+    let jwt = JsonWebToken::new(claims, local.keypair().secret().as_ref())?;
+
+    println!("{}", jwt);
```
Thoralf-M:
```suggestion
    println!("{jwt}");
```

bee-api/bee-rest-api/src/endpoints/routes/api/v2/transactions_included_block.rs
```diff
@@ -0,0 +1,57 @@
...
+    args: ApiArgsFullNode<B>,
+    transaction_id: TransactionId,
+) -> Result<BlockId, ApiError> {
+    // Safe to unwrap since 0 is a valid index;
+    let output_id = OutputId::new(transaction_id, 0).unwrap();
+
+    let fetched = Fetch::<OutputId, CreatedOutput>::fetch(&*args.storage, &output_id).map_err(|e| {
+        error!("cannot fetch from storage: {}", e);
+        ApiError::InternalServerError
+    })?;
```
Thoralf-M, line 52: What if the transaction has multiple outputs and the first one got spent and pruned? Shouldn't it still be able to return the block id if there are other unspent outputs of this transaction?

samuel-rufi: Good question. But what happens if the first and also the second output is removed? try for the third?
Also how to know how many outputs exist?

Thoralf-M: Don't know, try all possible 128 :carlos: Or maybe just leave like it is, after all the block will also be pruned if an output that got created "in it" is pruned?

samuel-rufi: > Don't know, try all possible 128 :carlos:
:carlos:

@thibault-martinez what do you suggest here? should we keep it like it is? Also IMO it's not something that has to be part of this PR!

bee-api/bee-rest-api/src/endpoints/routes/api/v2/milestones_by_index.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+use axum::{
```
thibault-martinez:
```suggestion

use axum::{
```

bee-api/bee-rest-api/src/endpoints/error.rs
```diff
@@ -0,0 +1,67 @@
...
+use thiserror::Error;
+
+use crate::types::body::{DefaultErrorResponse, ErrorBody};
+
+// Errors that are returned to the user.
+#[derive(Error, Debug)]
+pub enum ApiError {
+    #[error("bad request: {0}")]
+    BadRequest(&'static str),
+    #[error("bad request: {0}")]
```
thibault-martinez, line 18: Wrong comment

bee-node/bee-node/src/entrynode/builder.rs
```diff
@@ -14,6 +14,7 @@ use bee_autopeering::{
 };
 use bee_gossip::Keypair;
 use bee_plugin_version_checker::VersionCheckerPlugin;
+use bee_rest_api::endpoints::InitConfigEntryNode;
```
thibault-martinez: Rename to put the `Config` part last for consistency with other config objects

bee-api/bee-rest-api/src/endpoints/error.rs
samuel-rufi: why is it wrong?
dependency errors are "bad request errors"

bee-node/bee-node/src/entrynode/builder.rs
samuel-rufi: Fixed

bee-api/bee-rest-api/src/endpoints/error.rs
thibault-martinez: Idk it feels copy-pasted from the previous one, why not a more precise message ?

samuel-rufi: The idea was that the part before `{0}` in the comments hints about the HTTP status code class. 
So to reflect:
- bad request
- not found
- service unavailable
- internal server error
- forbidden

And since `DependencyError` is a `bad request` i added that prefix. 
But what would could we add as a more precise message? 
Keep in mind the full error details will get included into the "{0}"

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -1,51 +1,107 @@
...
-mod filters;
-
 pub mod config;
-pub mod path_params;
-pub mod permission;
-pub mod rejection;
+pub mod error;
+pub mod extractors;
 pub mod routes;
 pub mod storage;
```
thibault-martinez, line 9: Remove the new line

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -1,51 +1,107 @@
...
-    node::{Node, NodeBuilder},
+    event::Bus,
+    node::{Node, NodeBuilder, NodeInfo},
+    resource::ResourceHandle,
     worker::{Error as WorkerError, Worker},
 };
 use bee_tangle::{Tangle, TangleWorker};
-use log::{error, info};
-use warp::{http::StatusCode, Filter, Rejection, Reply};
+use config::RestApiConfig;
```
thibault-martinez: Prefix self otherwise fmt thinks it's a 3rd paty

bee-api/bee-rest-api/src/endpoints/mod.rs
```diff
@@ -1,51 +1,107 @@
...
+use log::info;
+use storage::StorageBackend;
```
thibault-martinez: Same

bee-api/bee-rest-api/src/endpoints/auth.rs
```diff
@@ -0,0 +1,148 @@
...
+    http::Uri,
+};
+use lazy_static::lazy_static;
+use regex::RegexSet;
+use serde::{Deserialize, Serialize};
+
+use crate::endpoints::{config::route_to_regex, error::ApiError, storage::StorageBackend, ApiArgsFullNode};
+
+pub const API_AUDIENCE_CLAIM: &str = "api";
+pub const DASHBOARD_AUDIENCE_CLAIM: &str = "dashboard";
```
thibault-martinez, line 24: Should this be under the dashboard feature ?

bee-api/bee-rest-api/src/endpoints/auth.rs
```diff
@@ -0,0 +1,148 @@
...
+const API_JWT_HINT: &str = "\"aud\":\"api\"";
+#[cfg(feature = "dashboard")]
+const DASHBOARD_JWT_HINT: &str = "\"aud\":\"dashboard\"";
+
+lazy_static! {
```
thibault-martinez, line 30: dashboard feature ?

bee-api/bee-rest-api/src/endpoints/auth.rs
samuel-rufi: There was a reason to not add it (unfortunately I can't remember the reason fully, but if I remember correctly it was some dependency which made use of that constant). I tested Bee now with this feature flag added, and everything runs fine.
So I added it as you suggested!

bee-api/bee-rest-api/src/endpoints/routes/api/v2/milestones_by_id.rs
```diff
@@ -24,7 +24,7 @@ async fn milestones_by_id<B: StorageBackend>(
 ) -> Result<impl IntoResponse, ApiError> {
     let milestone_payload = match args.tangle.get_milestone(milestone_id) {
```
pvdrz: I think you can do `let milestone_payload =  args.tangle.get_milestone(milestone_id).ok_or_else(|| ApiError::NotFound)?;` instead.

pvdrz (changes requested):

thibault-martinez (approved):

Alex6323 (approved):

PR #1304, opened 2022-03-30
Protocol version checks

thibault-martinez:
## Change checklist


pvdrz (approved):

bee-message/src/constant.rs
```diff
@@ -3,3 +3,6 @@
 
 /// The total number of IOTA tokens in circulation.
 pub const IOTA_SUPPLY: u64 = 2_779_530_283_277_761;
+
+/// Protocol version currently used by the network.
+pub const PROTOCOL_VERSION: u8 = 2;
```
samuel-rufi, line 8: Just wondering, why can we change this now? Now the constant describes exactly the message layout that is in the crate. So "protocol version" actually means "message layout version"?

Would it make sense to call it "message layout version" instead? The term "protocol version" might be misleading?

bingyanglin (approved):

PR #1324, opened 2022-04-04
Easier output creation with byte cost as amount

### LINKED ISSUE #1268: Easier output creation with minimum storage deposit
## Description

Make it possible to create outputs with the minimum required storage deposit without calculating it with the same output before
Some `.use_minimum_storage_deposit(byte_cost_config)` method (name can be different) in the output builder would probably be the easiest way

## Motivation

Easier to use, now I use https://github.com/iotaledger/wallet.rs/blob/shimmer-develop/src/account/operations/transfer/high_level/minimum_storage_deposit.rs only for that

## Requirements

1. Make it easier than now, so https://github.com/iotaledger/wallet.rs/blob/shimmer-develop/src/account/operations/transfer/high_level/minimum_storage_deposit.rs isn't needed anymore

## Open questions (optional)

Instead of `OutputBuilder::new(amount).use_minimum_storage_deposit()`would `Output::new_with_minimum_storage_deposit(byte_cost_config)` the cleaner way, because one doesn't has to provide some random `amount` first?

No.

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-message/src/output/alias.rs
+++ b/bee-message/src/output/alias.rs
@@ -17,8 +17,8 @@ use crate::{
     output::{
         feature_block::{verify_allowed_feature_blocks, FeatureBlock, FeatureBlockFlags, FeatureBlocks},
         unlock_condition::{verify_allowed_unlock_conditions, UnlockCondition, UnlockConditionFlags, UnlockConditions},
-        AliasId, ByteCost, ByteCostConfig, ChainId, NativeToken, NativeTokens, Output, OutputAmount, OutputId,
-        StateTransitionError, StateTransitionVerifier,
+        AliasId, ByteCost, ByteCostConfig, ChainId, NativeToken, NativeTokens, Output, OutputAmount,
+        OutputBuilderAmount, OutputId, StateTransitionError, StateTransitionVerifier,
     },
     semantic::{ConflictReason, ValidationContext},
     unlock_block::UnlockBlock,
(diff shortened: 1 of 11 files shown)


bee-message/src/output/alias.rs
```diff
@@ -170,7 +192,18 @@ impl AliasOutputBuilder {
...
-        })
+        };
+
+        output.amount = match (self.amount, self.byte_cost_config) {
+            (Some(amount), None) => amount,
+            (None, Some(byte_cost_config)) => Output::Alias(output.clone())
+                .byte_cost(&byte_cost_config)
+                .try_into()
+                .map_err(Error::InvalidOutputAmount)?,
+            _ => unreachable!(),
```
pvdrz: This makes me thing that `amount` and `byte_cost_config` should be something like `Result<OutputAmount, ByteCostConfig>`

pvdrz (changes requested):

bee-message/src/output/alias.rs
thibault-martinez: I proposed something very similar to @Thoralf-M but it looked like it was better in terms of API to have 2 different methods.

thibault-martinez: I mean, we could actually do a mix of both. Have the two methods AND the enum.

pvdrz: Yeah I'd keep the methods but use an enum internally so these illegal states cannot be represented at all.

pvdrz (approved):

samuel-rufi (approved):

Thoralf-M (approved):

PR #1330, opened 2022-04-05
Actually sum native tokens

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-ledger/src/types/unspent.rs
+++ b/bee-ledger/src/types/unspent.rs
@@ -37,7 +37,7 @@ impl Unspent {
 
 impl core::fmt::Display for Unspent {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
-        write!(f, "{}", *self)
+        write!(f, "{}", self.0)
     }
 }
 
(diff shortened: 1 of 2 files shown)


bingyanglin (dismissed):

bingyanglin (approved):

bee-ledger/src/types/unspent.rs
```diff
@@ -37,7 +37,7 @@ impl Unspent {
 
 impl core::fmt::Display for Unspent {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
-        write!(f, "{}", *self)
+        write!(f, "{}", self.0)
```
Alex6323, line 40:
```suggestion
        self.0.fmt(f)
```
I prefer doing this if I just want to "forward" the formatting. It's a bit less noisy and doesn't require macro expansion.

thibault-martinez: I would rather not do this, or at least not now as the whole codebase uses the macro and it would be inconsistent. I personally prefer the macro as it's easier to maintain if we need to change the formatting.

grtlr (approved):

Alex6323 (approved):

pvdrz (approved):

PR #1338, opened 2022-04-07
Reorganise foundry arithmetic operations to avoid overflows/underflows

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-message/src/output/foundry.rs
+++ b/bee-message/src/output/foundry.rs
@@ -417,10 +417,16 @@ impl StateTransitionVerifier for FoundryOutput {
             Ordering::Less => {
                 // Mint
 
-                if &(current_token_scheme.minted_tokens() + (output_tokens - input_tokens))
-                    != next_token_scheme.minted_tokens()
-                    || current_token_scheme.melted_tokens() != next_token_scheme.melted_tokens()
-                {
+                // This can't underflow as it is known that current_minted_tokens <= next_minted_tokens.
+                let minted_diff = next_token_scheme.minted_tokens() - current_token_scheme.minted_tokens();
+                // This can't underflow as it is known that input_tokens < output_tokens (Ordering::Less).
+                let token_diff = output_tokens - input_tokens;
+
+                if minted_diff != token_diff {
+                    return Err(StateTransitionError::InconsistentNativeTokensMint);
+                }
+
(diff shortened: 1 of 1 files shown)


grtlr (approved):

bingyanglin (approved):

bee-message/src/output/foundry.rs
```diff
@@ -442,9 +448,12 @@ impl StateTransitionVerifier for FoundryOutput {
                     return Err(StateTransitionError::InconsistentNativeTokensMeltBurn);
```
Alex6323, line 448: I can only comment on this line because the previous 3 are not part of this PR, but I just want to make sure the above if-condition is correct: Currently only if both inqualities are true, we return this state transition error, but what if only one of those is true? Isn't that inconsistent already?

thibault-martinez: Yes this is correct. It means that you are not allowed to mint while melting but you are allowed to mint while burning. It implements the following rule from https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#token-scheme-semantic-validation-rules `When Current State(Melted Tokens) != Next State(Melted Tokens), it must be true that Current State(Minted Tokens) = Next State(Minted Tokens)`.

Alex6323: Oh no, I looked at the wrong line: Ordering::Equal, but there it's correct.

thibault-martinez: All good then ?

Alex6323: Honestly, I still have a hard time why the melting is not mirroring the minting transition :thinking: I feel like I am looking at a bug here, but I am probably having a misunderstanding.

Alex6323: ah, okay, didn't see your message. All good then :+1:

samuel-rufi (approved):

bee-message/src/output/foundry.rs
thibault-martinez: Because `Ordering::Greater` is about melting + burning, not only melting. Going for lunch now, can answer questions later.

Alex6323: me too, need some lunch

Alex6323 (approved):

PR #1411, opened 2022-05-18
Propagate `ProcessMetrics::update` error

pvdrz:
cc @AlexandCoats

### DIFF
--- a/bee-metrics/src/metrics/process.rs
+++ b/bee-metrics/src/metrics/process.rs
@@ -32,12 +32,9 @@ impl ProcessMetrics {
         (MemoryUsage(self.mem.clone()), CpuUsage(self.cpu.clone()))
     }
 
-    /// Updates the metrics values.
-    ///
-    /// The value is not updated if the command used to retrieve the new value:
-    ///  - cannot be spawned,
-    ///  - returns an unsuccessful exit code,
-    ///  - its output cannot be parsed.
+    /// Updates the metrics values. Returns an error if it was not possible to fetch the values
+    /// from the respective OS command:
+    /// - `ps` for unix-like platforms.
     pub async fn update(&self) -> Result<(), Box<dyn std::error::Error>> {
         #[derive(Debug)]
         enum UpdateError {
--- a/bee-metrics/src/metrics/process.rs
+++ b/bee-metrics/src/metrics/process.rs
@@ -38,32 +38,58 @@ impl ProcessMetrics {
     ///  - cannot be spawned,
     ///  - returns an unsuccessful exit code,
     ///  - its output cannot be parsed.
-    pub async fn update(&self) {
+    pub async fn update(&self) -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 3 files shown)


bee-node/bee-node/src/workers.rs
```diff
@@ -61,7 +61,9 @@ impl<N: Node> Worker<N> for MetricsRegistryWorker {
                 ShutdownStream::new(shutdown, IntervalStream::new(interval(PROCESS_METRICS_UPDATE_INTERVAL)));
 
             while ticker.next().await.is_some() {
-                process_metrics.update().await;
+                if let Err(e) = process_metrics.update().await {
+                    debug!("Cannot update process metrics: {e}.");
```
thibault-martinez: Just checking, is this important that we are aware of this error ? Because debug is so spammed that we could barely see it

bee-metrics/src/metrics/process.rs
```diff
@@ -32,38 +32,61 @@ impl ProcessMetrics {
...
-    ///
-    /// The value is not updated if the command used to retrieve the new value:
-    ///  - cannot be spawned,
-    ///  - returns an unsuccessful exit code,
-    ///  - its output cannot be parsed.
-    pub async fn update(&self) {
+    /// Updates the metrics values. Returns an error if it was not possible to fetch the values
+    /// from the respective OS command:
+    /// - `ps` for unix-like platforms.
+    pub async fn update(&self) -> Result<(), Box<dyn std::error::Error>> {
```
thibault-martinez, line 38: Why dyn and not `UpdateError` ?

bee-node/bee-node/src/workers.rs
pvdrz: nah. If it happens the metric won't update. We just need to catch it once in case we want to know what went wrong,

bee-metrics/src/metrics/process.rs
pvdrz: I'm not sure if exposing the error actually matters as it will be most likely platform specific due to output formatting and things like that.

At the same time, if this fails, you'd want to just print the error to see what went wrong and not actually handle it.

bee-node/bee-node/src/workers.rs
```diff
@@ -51,6 +51,11 @@ impl<N: Node> Worker<N> for MetricsRegistryWorker {
         let registry = node.resource::<Registry>();
 
         let process_metrics = ProcessMetrics::new(config.pid);
+
+        if let Err(e) = process_metrics.update().await {
+            info!("Cannot update process metrics: {e}.");
```
thibault-martinez: Can we do warn instead ?

bee-node/bee-node/src/workers.rs
pvdrz: I decided to emit the log if it fails the first time just to check that the command can actually run and ignore the rest.

thibault-martinez (approved):

Alex6323 (approved):

PR #1456, opened 2022-07-11
Fix TaggedDataPayloadDto

Thoralf-M:
# Description of change

Fix TaggedDataPayloadDto

## Type of change

iota.rs example

## Change checklist

### DIFF
--- a/bee-block/src/payload/tagged_data/mod.rs
+++ b/bee-block/src/payload/tagged_data/mod.rs
@@ -73,6 +73,7 @@ pub mod dto {
     pub struct TaggedDataPayloadDto {
         #[serde(rename = "type")]
         pub kind: u32,
+        #[serde(skip_serializing_if = "String::is_empty", default)]
         pub tag: String,
         #[serde(skip_serializing_if = "String::is_empty", default)]
         pub data: String,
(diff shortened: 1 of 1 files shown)


bee-block/src/payload/tagged_data/mod.rs
```diff
@@ -74,6 +74,7 @@ pub mod dto {
         #[serde(rename = "type")]
         pub kind: u32,
         pub tag: String,
+        #[serde(skip_serializing_if = "String::is_empty", default)]
```
thibault-martinez, line 78: Should probably do the same with the tag ?

Thoralf-M: Doesn't sound so
```
        tag:
          type: string
          description: The indexation key to allow external tools to find/look up this block. It has a size between 1 and 64 bytes and must be encoded as a hex-string with 0x prefix.
```

Thoralf-M: Btw will comment on the tip because it says indexation key :trollface:

thibault-martinez (approved):

samuel-rufi (approved):

PR #1457, opened 2022-07-14
Add OutputDtos new methods

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-block/src/output/mod.rs
+++ b/bee-block/src/output/mod.rs
@@ -300,7 +300,7 @@ pub mod dto {
     };
     use crate::error::dto::DtoError;
 
-    #[derive(Clone)]
+    #[derive(Clone, Debug, Deserialize)]
     pub enum OutputBuilderAmountDto {
         Amount(String),
         MinimumStorageDeposit(ByteCostConfig),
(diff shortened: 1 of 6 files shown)


samuel-rufi (dismissed):

bee-block/src/output/mod.rs
```diff
@@ -300,6 +300,12 @@ pub mod dto {
     };
     use crate::error::dto::DtoError;
 
+    #[derive(Clone)]
```
Thoralf-M: What about Debug and Deserialize?

thibault-martinez: Yeah can add, doubt Deserialize would be interesting as this is never stored, just temporarily used

Thoralf-M: Yeah don't know how you want to use it, if should be provided from bindings we need Deserialize

pvdrz (approved):

bee-block/src/output/mod.rs
thibault-martinez: Added them, but shouldn't be an issue anyway

Thoralf-M (approved):

PR #1475, opened 2022-07-21
Add `bee-inx` crate

grtlr:
## Change checklist


bee-inx/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+edition = "2021"
+description = "Rust bindings for IOTA node extensions (INX)"
+readme = "README.md"
+repository = "https://github.com/iotaledger/bee"
+license = "Apache-2.0"
+keywords = [ "iota", "plugin", "node", "inx", "bee" ]
+homepage = "https://www.iota.org"
+
+[dependencies]
+bee-block = { path = "../bee-block", version = "1.0.0-beta.3", default-features = false, features = [ "inx", "std" ] }
```
Alex6323: Why are we giving the full beta release version here for a path dependency again? This implies a lot of unnecessary changes when the dependency bumps the index, but I don't see the advantage of doing it. Simply locking it to `version = "1"` would actually achieve the same and prevent lots of line changes, right?

bee-inx/README.md
```diff
@@ -0,0 +1 @@
+# bee-inx
```
Alex6323, line 1:
```suggestion
# bee-inx

Rust bindings for IOTA node extensions (INX).
```

bee-inx/examples/milestone.rs
```diff
@@ -0,0 +1,37 @@
...
+
+        while let Some(Ok(block_metadata)) = cone_stream.next().await {
+            println!(
+                "Block {}: {:#?}",
+                block_metadata.metadata.block_id, block_metadata.block
+            );
+            count += 1;
+        }
+
+        println!("Milestone {:?} contained {count} blocks", milestone_index);
```
Alex6323: I know it's just an example, but I think this should be spawned in a task.

bee-inx/src/ledger.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    fn try_from(value: proto::LedgerOutput) -> Result<Self, Self::Error> {
+        Ok(Self {
+            output_id: value
+                .output_id
+                .ok_or(Self::Error::MissingField("output_id"))?
+                .try_into()?,
+            block_id: value
+                .block_id
+                .ok_or(Self::Error::MissingField("message_id"))?
```
Thoralf-M:
```suggestion
                .ok_or(Self::Error::MissingField("block_id"))?
```

bee-inx/examples/milestone.rs
```diff
@@ -0,0 +1,37 @@
...
+async fn main() -> Result<(), Error> {
+    let mut inx = client::Inx::connect(INX_ADDRESS.into()).await?;
+    let mut milestone_stream = inx.listen_to_confirmed_milestones((..).into()).await?;
+
+    // Listen to the milestones from the node.
+    while let Some(milestone) = milestone_stream.next().await {
+        let milestone_index = milestone?.milestone_info.milestone_index;
+        println!("Fetch cone of milestone {milestone_index}");
+
+        // Listen to messages in the past cone of a milestone.
```
Thoralf-M:
```suggestion
        // Listen to blocks in the past cone of a milestone.
```

bee-inx/examples/milestone.rs
Alex6323: Also...it's probably enough to just print the `BlockId` for each block in the the cone. Otherwise the console will just be spammed with data that doesn't really add much to the example.

bee-inx/src/client.rs
```diff
@@ -0,0 +1,86 @@
...
+    ) -> Result<impl Stream<Item = Result<crate::UnspentOutput, Error>>, Error> {
+        Ok(self
+            .inx
+            .read_unspent_outputs(proto::NoParams {})
+            .await?
+            .into_inner()
+            .map(unpack_proto_msg))
+    }
+
+    pub async fn read_milestone_cone(
```
Thoralf-M, line 77: Maybe this should be `read_milestone_past_cone`, since it could also be `read_milestone_future_cone` in theory?

Thoralf-M: Or if it's only returning the blocks that got referenced by this milestone and not from other milestones before, then not sure if `cone` is the right word at all

bee-inx/src/block.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use bee_block as bee;
+use inx::proto;
+
+/// The [`Block`] type.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Block {
```
Alex6323, line 11: Below we have `BlockWithMetadata`. Consistently this could be named `BlockWithId` which would more clearly separate it from the `bee_block::Block` type.

bee-inx/src/block.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// The [`BlockId`](bee::BlockId) of the block.
+    pub block_id: bee::BlockId,
```
Alex6323, line 13: Maybe just rename this field to `id` because there's no other `xxx_id` in this struct to distinct it from.

bee-inx/src/block.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    fn try_from(value: proto::BlockWithMetadata) -> Result<Self, Self::Error> {
+        let raw = value.block.ok_or(bee::InxError::MissingField("block"))?;
+        Ok(BlockWithMetadata {
+            metadata: value
+                .metadata
+                .ok_or(Self::Error::MissingField("metadata"))?
+                .try_into()?,
+            block: raw.clone().try_into()?,
+            raw: raw.data,
```
Alex6323: It's probably(!) more performant to do a mem-copy of `raw.data` beforehand into a `data` Vec, and then just call `try_into` on the `raw` struct itself. This clones only the data that actually needs to be cloned. Same for all other instances.

bee-inx/Cargo.toml
grtlr: The other Bee crates do it the same way. If we were to change it here, we should change it everywhere. I've flipped `version` and `path` to be more consistent though.

bee-inx/examples/milestone.rs
grtlr: Why do you think this should be done in a task? Also, I've changed the output.

bee-inx/src/block.rs
grtlr: I'm following the naming conventions from the [protobuf definitions](https://github.com/iotaledger/inx/blob/develop/proto/inx.proto) here.

bee-inx/src/block.rs
grtlr: I should have actually just used the the `Raw<bee::Block>` struct that I created for these cases. Now the user can choose when the copy is done and the performance hit occurs.

bee-inx/examples/milestone.rs
Alex6323: Just to make it faster on multi-core machines I guess, and essentially make the `milestone` processing more independent from how long the cone takes to be processed.

bee-inx/src/ledger.rs
grtlr: Created a convenience macro `crate::field!` that prevents these type of mistakes.

bee-inx/src/block.rs
Alex6323: Right...From the proto perspective it might make sense to just name it `Block` (as it doesn't have to care much about the higher layers), but I wonder if that (the naming) cannot just be abstracted away here at Bee's layer. :shrug: From simply looking at the type definitions in this module (and how similar they are modelled it seems weird, that one is just called `Block` and the other `BlockWithMetadata`. Of course your argument also makes sense, and we could just follow the naming conventions from upstream. So this is just a suggestion.

bee-inx/src/client.rs
grtlr: I'm following the naming conventions from the [protobuf definitions](https://github.com/iotaledger/inx/blob/develop/proto/inx.proto) here.

bee-inx/examples/milestone.rs
grtlr: But then the output would not be in order anymore, because multiple tasks could work on this concurrently? I don't know if it's worth the additional complexity.

bee-inx/src/block.rs
grtlr: I think for now, especially since documentation around INX is still sparse, it makes sense to stick with the same naming conventions for now. Once INX is more stable, it might make sense to deviate more from them to make things more Rust-y.

bee-inx/src/block.rs
grtlr: I'm following the naming conventions from the [protobuf definitions](https://github.com/iotaledger/inx/blob/develop/proto/inx.proto) here.

bee-inx/src/client.rs
Alex6323: What would be a better term then in your opinion? :trollface:

Thoralf-M: Don't know :trollface:

bee-inx/Cargo.toml
Alex6323: I am closing this, but I'll open an issue to get clear on that. At least it doesn't seem to be THAT obvious :slightly_smiling_face:

bee-inx/examples/milestone.rs
Alex6323: Yea, the output would potentially not be in order anymore. I didn't know that this is a requirement for this example. I thought it should be more an example of how you should implement/use this API.

bee-inx/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+pub mod proto {
+    pub use inx::proto::*;
+}
+
+#[macro_export]
+macro_rules! field {
+    ($object:ident.$field:ident) => {
+        $object.$field.ok_or(Self::Error::MissingField(stringify!($field)))?
+    };
+}
```
Alex6323: kinda neat, but the naming makes the code harder to read at first glance. How'about `maybe_missing_field` or just `maybe_missing` or something similar already signaling what issue the macro is going to address?

bee-inx/src/node/config.rs
```diff
@@ -0,0 +1,30 @@
...
+
+/// The [`NodeConfiguration`] type.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct NodeConfiguration {
+    pub protocol_parameters: bee::protocol::ProtocolParameters,
+    pub milestone_public_key_count: u32,
+    // TODO: `milestone_key_ranges`
+    // TODO: `base_token`
+    pub supported_protocol_versions: Box<[u8]>,
+    // TODO: `pending_protocol_parameters`
```
Alex6323: Are those TODOs to be done in this PR?

bee-inx/examples/milestone.rs
grtlr: If you want, you can always provide a more elaborate example! :trollface:

bee-inx/src/request.rs
```diff
@@ -0,0 +1,104 @@
...
+                end_milestone_index: 42
+            })
+        );
+    }
+
+    #[test]
+    fn inclusive() {
+        let from = 17u32;
+        let to = 42;
+        let range = MilestoneRangeRequest::from(from..=to);
```
Alex6323:
```suggestion
        let range = MilestoneRangeRequest::from(17..=42);
```

bee-inx/src/node/config.rs
grtlr: Some of these, like `milestone_key_ranges` are still kind of blocked by Bee types, so I thought I'd do those later.

grtlr: I've implemented everything except for `milestone_key_ranges`.

Alex6323 (approved):

PR #1487, opened 2022-08-04
Update `inx` version to `v1.0.0-beta.4`

grtlr:
## Change checklist

### DIFF
--- a/bee-inx/src/ledger.rs
+++ b/bee-inx/src/ledger.rs
@@ -38,9 +38,9 @@ pub struct UnspentOutput {
 #[allow(missing_docs)]
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Marker {
-    milestone_index: MilestoneIndex,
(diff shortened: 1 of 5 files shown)


bee-inx/src/ledger.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use bee::payload::milestone::MilestoneIndex;
```
Thoralf-M: Should be aligned with the other imports, why don't import them all at the top?

grtlr: Generally I agree with you, but here it makes it easier to see where the types come from, in my opinion.

Thoralf-M: But why import the MilestoneIndex then not also in the code where needed?

Thoralf-M (approved):

PR #1499, opened 2022-08-24
Add Alias state transition constraint to unlock an Alias address

thibault-martinez:
## Change checklist

### DIFF
--- a/bee-block/src/output/alias.rs
+++ b/bee-block/src/output/alias.rs
@@ -521,7 +521,6 @@ impl StateTransitionVerifier for AliasOutput {
             // Governance transition.
             if current_state.amount != next_state.amount
                 || current_state.native_tokens != next_state.native_tokens
-                || current_state.state_index != next_state.state_index
                 || current_state.state_metadata != next_state.state_metadata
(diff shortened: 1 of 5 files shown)


DaughterOfMars (dismissed):

bee-block/src/output/alias.rs
```diff
@@ -437,27 +437,21 @@ impl AliasOutput {
...
-                    } else {
-                        self.state_controller_address()
-                    }
+        match next_state {
+            Some(Output::Alias(next_state)) => {
+                if self.state_index() == next_state.state_index() {
+                    self.governor_address().unlock(unlock, inputs, context)?;
+                } else {
+                    self.state_controller_address().unlock(unlock, inputs, context)?;
+                    context
```
Thoralf-M, line 448: Maybe good to add a comment here that the alias is only unlocked when it's a state transition, otherwise it could look like we forgot it for the state controller when not reading the tip

thibault-martinez: Sure thing

Thoralf-M (dismissed):

